PR #47, opened 2025-06-13
fix(uart): fix the UART driver in SF32LB52

HalfSweet:


src/fw/drivers/sf32lb52/uart.c
```diff
@@ -23,10 +23,13 @@
 #include "bf0_hal_dma.h"
 #include "bf0_hal_uart.h"
 
+#define container_of(ptr, type, member) ((type *)((char *)(ptr) - (size_t)&(((type *)0)->member)))
```
gmarull: nit: create a separate commit adding this to a new header e.g. in `src/libutil/includes/util/misc.h`.

src/libbtutil/include/util/misc.h
```diff
@@ -0,0 +1,5 @@
+#pragma once
```
gmarull: missing license header here, please add this change in a first commit, then uart fixes.

HalfSweet: Do I need to create a separate PR to create `misc.h/misc.h`, or can it be included in this PR?

gmarull: same PR is fine, just another commit. It's easier to track changes this way when browsing git history.

gmarull (approved):
tested on eval kit, uart seems to work fine after these changes. just pending split of container_of bit.

PR #48, opened 2025-06-14
Delay vibration for one second after a notification is received

ericmigi:
This is an improvement I've been dreaming of for years. Right now, when you get a notification, the vibration happens immediately, but the UI takes a while to show the animation and finally display the text on screen. This PR fixes it by adding a delay. Now, when you look down at your watch after feeling a vibration, you'll see the full notification rather than waiting a few milliseconds to see the notification. 

- Add delayed vibration timer to notification window
- Use 1s delay for first notification, 0.2s for subsequent notifications

Keep in mind that I can't actually code, this was vibe coded with [cline](https://github.com/cline/cline) + Sonnet 3.7. You can be ruthless with the code review - you can't hurt my feelings. 

_(video: turn sound on)_

https://github.com/user-attachments/assets/ad648211-d819-4fdd-a7a7-6ddda45010e0

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_notifications.c
+++ b/src/fw/apps/system_apps/settings/settings_notifications.c
@@ -57,6 +57,7 @@ enum NotificationsItem {
 #endif
   NotificationsItemTextSize,
   NotificationsItemWindowTimeout,
+  NotificationsItemDelayedVibration,
   NotificationsItem_Count,
 };
 
@@ -185,7 +186,7 @@ static void prv_text_size_menu_push(SettingsNotificationsData *data) {
       true /* icons_enabled */, s_text_size_names, data);
 }
 
-// Text Size
+// Window Timeout
 ////////////////////////
 
 // NOTE: Keep the following two arrays in sync and with the same size.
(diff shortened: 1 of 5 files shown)


aveao:
Potential edge case that comes to mind (have not tested it): During the initial notification draw (the 1 second long one) if you get another notification, it will show 1/2 etc on the top right and will still play the 1s anim til the end. I think the current structure of the code may mean that you cancel the first 1s timer and instead switch to a 0.2s one as soon as the second one arrives.

WuerfelDev:
I don't like the idea of delaying notifications. I'm not against it, but if this is going to be merged, can it be with a setting or a configuration of some kind?

Since there could be a lot of small changes and tweaks like this, i suggest having "advanced settings": They are not shown on the watch to keep it simple, but they can be set through the app.

ericmigi:
Good idea to make this an optional setting, easy to do by adding it to the notification settings menu

ericmigi:
Done. Added it as a setting. Later if/when we add the ability to change settings from the phone app, it would be natural to allow that to be changed there.

 I choose not to make the vibration happen precisely after the animation because I kinda liked seeing the tail end of it. If anyone has any suggestions on a better algorithm to use, feel free to suggest. One idea is to actually speed up the animation itself, but that would probably require building the animation file from source.

src/fw/apps/system_apps/settings/settings_notifications.c
```diff
@@ -185,7 +186,7 @@ static void prv_text_size_menu_push(SettingsNotificationsData *data) {
       true /* icons_enabled */, s_text_size_names, data);
 }
 
-// Text Size
+// Window Timeout
```
gmarull, line 189: ?

gmarull (commented):
I don't see anything particularly wrong, just a nitpick (note, I'm not very familiar with this area). Commits need to adhere to gitlint rules to pass compliance checks.

PR #56, opened 2025-06-19
fw/apps/alarm_editor: make timeline edit action select relevant alarm

BTPankow:
When using the "Edit" action for a given alarm timeline entry, the alarms app just selects the first alarm in the list rather than the relevant timeline alarm.

This adds a timeline action argument to select the correct alarm when editing from the timeline.

This video shows the previous behavior of editing alarms via the timeline. Selecting "Edit" for an alarm in the timeline always opens up the alarms editor with just the first entry selected.

https://github.com/user-attachments/assets/641b040c-2c1a-45d6-9a87-7090108bb486

And this video shows the new behavior with this change. Now the relevant alarm is highlighted once "Edit" is selected for a given alarm in the timeline.

https://github.com/user-attachments/assets/fbae60c8-4271-4c95-b4b4-b8a5a27a4f36


src/fw/apps/system_apps/alarms/alarms.c
```diff
@@ -439,9 +439,18 @@ static void prv_handle_init(void) {
...
+      if (list_idx == 0) {
+        list_idx = 1; // Default to first alarm if idx not found
+      }
+    }
+
     app_window_stack_push(&data->window, true);
-    menu_layer_set_selected_index(&data->menu_layer, MenuIndex(0, 1),
-                                  PBL_IF_RECT_ELSE(MenuRowAlignNone, MenuRowAlignCenter), false);
+    menu_layer_set_selected_index(&data->menu_layer, MenuIndex(0, list_idx),
+                                  MenuRowAlignCenter, false);
```
gmarull, line 453: is `PBL_IF_RECT_ELSE` removal intentional?

src/fw/services/normal/alarms/alarm_pin.c
```diff
@@ -41,22 +41,19 @@ static void prv_set_pin_attributes(AttributeList *list, AlarmType type, AlarmKin
...
   const unsigned num_actions = 1; // We are just supporting "edit" for now
   TimelineItemActionGroup action_group = {
     .num_actions = num_actions,
     .actions = task_zalloc_check(sizeof(TimelineItemAction) * num_actions),
   };
+
   AttributeList edit_attr_list = {0};
-  prv_set_edit_action_attributes(&edit_attr_list);
+  attribute_list_add_cstring(&edit_attr_list, AttributeIdTitle, i18n_get("Edit", &edit_attr_list));
+  attribute_list_add_uint32(&edit_attr_list, AttributeIdLaunchCode, (uint32_t) id);
```
gmarull: maybe keep `prv_set_edit_action_attributes`, passing id?

BTPankow: Sure, I didn't see a huge reason to have the indirection, but I can see where it's consistent with `prv_set_pin_attributes` to have a separate function. Updated.

src/fw/apps/system_apps/alarms/alarms.c
BTPankow: Yes, I changed this to ensure the selected row shows up on screen. Otherwise, it would not be centered on the screen or could even be off-screen for rectangular screens.

In this case, [MenuAlignCenter](https://developer.rebble.io/developer.pebble.com/docs/c/User_Interface/Layers/MenuLayer/index.html#MenuRowAlign) makes sense to center the selected alarm on launch.

I'm honestly not sure why the distinction exists in the first place. There is no difference in behavior with `MenuRowAlignCenter` and `MenuRowAlignNone` when just launching the alarms editor.

This video shows the behavior without this change:
https://github.com/user-attachments/assets/87a3edd0-dcd3-44b6-bf18-ac72a25802ab

PR #59, opened 2025-06-19
Initial BLE support for SF32LB52

gmarull:
Based on https://github.com/coredevices/pebbleos/pull/7


third_party/nimble/transport/hci_sf32lb52.c
```diff
@@ -0,0 +1,316 @@
...
+static ipc_queue_handle_t s_ipc_port;
+
+#ifdef NIMBLE_HCI_SF32LB52_TRACE
+static uint8_t s_hci_trace_buf[1024]
+static uint16_t s_hci_trace_seq;
+#endif
+
+static uint8_t s_read_buf[64];
+static uint8_t s_hci_acl[256];
+static uint8_t s_hci_cmd[256];
```
gmarull: @gangheivt do all these really need to be static buffers?

third_party/nimble/transport/hci_sf32lb52.c
```diff
@@ -0,0 +1,316 @@
...
+            host2ctrl ? "TX" : "RX", len);
+  PBL_HEXDUMP_D(LOG_DOMAIN_BT_STACK, LOG_LEVEL_DEBUG, data, len);
+  portENABLE_INTERRUPTS();
+#else
+  len += 16;
+  portDISABLE_INTERRUPTS();
+  for (uint16_t i = 0U; i < len; i++) {
+    uart_write_byte(HCI_TRACE_UART, s_hci_trace_buf[i]);
+  }
+  portENABLE_INTERRUPTS();
```
gmarull: @gangheivt is it really needed to disable IRQs here?

third_party/nimble/transport/hci_sf32lb52.c
```diff
@@ -0,0 +1,316 @@
...
+}
+#else
+#define prv_hci_trace(type, data, len, host2ctrl)
+#endif  // NIMBLE_HCI_SF32LB52_DEBUG
+
+static int32_t prv_ipc_rx_ind(ipc_queue_handle_t handle, size_t size) {
+  static BaseType_t xHigherPriorityTaskWoken = pdFALSE;
+
+  xSemaphoreGiveFromISR(s_ipc_data_ready, &xHigherPriorityTaskWoken);
+  // TODO: yield?
```
gmarull: @gangheivt ?

third_party/nimble/transport/hci_sf32lb52.c
```diff
@@ -0,0 +1,316 @@
...
+  struct os_mbuf *om;
+
+  switch (pkt_type) {
+  case HCI_H4_EVT:
+      ev = data;
+      cmd_complete = (void *)ev->data;
+
+      if (ev->opcode == BLE_HCI_EVCODE_COMMAND_COMPLETE) {
+        PBL_LOG_D(LOG_DOMAIN_BT_STACK, LOG_LEVEL_DEBUG, "\tCMD complete %x", cmd_complete->opcode);
+        if (cmd_complete->opcode == 0xFC11) {
```
gmarull, line 193: @gangheivt magic number please clarify

third_party/nimble/transport/hci_sf32lb52.c
```diff
@@ -0,0 +1,316 @@
...
+  struct ble_hci_cmd *cmd = buf;
+  int written;
+
+  s_hci_cmd[0] = 1;
+  memcpy(&(s_hci_cmd[1]), buf, 3 + cmd->length);
+
+  prv_hci_trace(HCI_H4_CMD, buf, 3 + cmd->length, 1);
+
+  written = ipc_queue_write(s_ipc_port, s_hci_cmd, 3 + cmd->length + 1, 10);
+  ble_transport_free(buf);
```
gmarull: @gangheivt quite a few magic numbers here (1, 3, 10...) please explain

gangheivt: first byte is HCI H4 type.   3 is H4 command packet header length,  10 is timeout value in ticks

third_party/nimble/transport/hci_sf32lb52.c
gangheivt: This is first vendor event for LCPU up.  Nimble will have issue if got this. I have to filter it out.

third_party/nimble/transport/hci_sf32lb52.c
gangheivt: Do not have to, but it is better to use as static

third_party/nimble/transport/hci_sf32lb52.c
gangheivt: It might be better to disable IRQs, other threads might use the UART.

third_party/nimble/transport/hci_sf32lb52.c
gangheivt: It could be removed now.

third_party/nimble/transport/hci_sf32lb52.c
gmarull: Does this event have a name? Some explanation is needed so things are clear

third_party/nimble/transport/hci_sf32lb52.c
gmarull: The question is: I think a yield is missing if flag returns true

third_party/nimble/transport/hci_sf32lb52.c
gmarull: HCI Uart is not shared, logging can handle it without special needs, was weird behavior observed without?

third_party/nimble/transport/hci_sf32lb52.c
gmarull: adjusted code so that no magic is present

third_party/nimble/transport/hci_sf32lb52.c
gmarull: added a comment for now

third_party/nimble/transport/hci_sf32lb52.c
gmarull: added yield

third_party/nimble/transport/hci_sf32lb52.c
gmarull: removed disabling, unclear why it was needed

third_party/nimble/transport/hci_sf32lb52.c
gmarull: simplified things, feels like some of these sizes should actually be adjusted according to NimBLE settings.

PR #75, opened 2025-06-25
fw/drivers/sf32lb52: sf32lb52 add i2c driver#25

w200024212:
The sf32lb52 i2c_hal driver provides but init and data transmission.
Signed-off-by: Jerrywang w200024212@163.com>


src/fw/drivers/sf32lb52/i2c_hal.c
```diff
@@ -0,0 +1,1071 @@
...
+    #endif
+    
+    #ifdef I2C5
+    I2C5_CFG_DEFAULT,
+    #endif
+        
+    #ifdef I2C6
+    I2C6_CFG_DEFAULT,
+    #endif
+};
```
gmarull: -1 to this method, this all needs to go at board level. See how UART was implemented.

gmarull (changes requested):
Check UART implementation on how to avoid all the ifdeffery/hardcoding of instances inside the driver.

src/fw/drivers/sf32lb52/i2c_hal.c
w200024212: ok

PR #83, opened 2025-06-27
fw/drivers/sf32lb52: sf32lb52 add rng driver

w200024212:
Signed-off-by: Jerrywang w200024212@163.com>

### DIFF
--- a/src/fw/apps/prf_apps/recovery_first_use_app/recovery_first_use_app.c
+++ b/src/fw/apps/prf_apps/recovery_first_use_app/recovery_first_use_app.c
@@ -105,7 +105,7 @@ typedef struct RecoveryFUAppData {
 } RecoveryFUAppData;
 
 #if PLATFORM_ASTERIX
-static const char *s_qr_url_fmt = "https://app.repebble.com/qr?sn=%s&model=%s";
+static const char *s_qr_url_fmt = "https://qr.core.store/?sn=%s&model=%s";
 #endif
 
 // Unfortunately, the event_service_client_subscribe doesn't take a void *context...


src/fw/drivers/sf32lb52/rng.c
```diff
@@ -0,0 +1,52 @@
...
+
+  /*##-2- Generate random ###################################*/
+  seed = 0;
+  status = HAL_RNG_Generate(&RngHandle, &seed, 1);
+  PBL_LOG(LOG_LEVEL_INFO, "Generate Random value=%x, status=%d", (unsigned int)seed, (int)status);
+  PBL_ASSERTN(seed != 0);
+  *rand_out = seed;
+
+  return true;
+}
```
gmarull:
```suggestion
#include "drivers/rng.h"
#include "system/passert.h"

static bool s_init;
static RNG_HandleTypeDef s_rng_hdl = {
  .Instance = hwp_trng,
};

bool rng_rand(uint32_t *rand_out) {
  HAL_StatusTypeDef status;

  if (!s_inited) {
    // Does it consume power to leave this ON?
    status = HAL_RNG_Init(&RngHandle);
    if (status != HAL_OK) {
      PBL_LOG(LOG_LEVEL_ERROR, "HAL_RNG_Init failed (%d)", status);
      return false;
    }
    s_init = true;
  }

  // Why did we have a variable called 'seed' initialized to zero?
  status = HAL_RNG_Generate(&RngHandle, rand_out, 1);
  if (status != HAL_OK) ...

  return true;
}
```

gmarull (changes requested):

src/fw/drivers/sf32lb52/rng.c
```diff
@@ -0,0 +1,47 @@
...
+      /* Initialization Error */
+      PBL_LOG(LOG_LEVEL_ERROR, "rng_rand init fail!");
+      return false;
+    }
+    s_init = true;
+  }
+
+  status = HAL_RNG_Generate(&s_rng_hdl, rand_out, 1);
+  PBL_LOG(LOG_LEVEL_INFO, "Generate Random value=%x, status=%d", (unsigned int)*rand_out,
+          (int)status);
```
gmarull: please do not add polluting logs, just log if there's an error. Also, you're ignoring status return value...

gmarull (changes requested):

src/fw/drivers/sf32lb52/rng.c
gmarull: ping

src/fw/drivers/sf32lb52/rng.c
```diff
@@ -0,0 +1,47 @@
...
+
+static bool s_init = false;
+static RNG_HandleTypeDef s_rng_hdl = {
+    .Instance = hwp_trng,
+};
+bool rng_rand(uint32_t *rand_out) {
+  PBL_ASSERTN(rand_out);
+  HAL_StatusTypeDef status;
+
+  if (!s_init) {
```
cameled: Hi @w200024212 , does there have implicit background for static rng initialization? Since our device using battery, no need to enable it if not in use.

rabbitsaviola: TRNG peripheral current leakage is about 200uA if HPSYS is not in sleep state. If HPSYS goes to sleep, this peripheral's clock would be gated also, i.e. no additional power consumption introduced.
If 200uA does affect battery life, it can be optimized by disabling its clock after the use.

PR #97, opened 2025-07-08
asterix: explicitly shut down sleeping peripherals

jwise:
This doesn't seem to have much of a difference on Asterix, but it's good to rule these kinds of things out, I suppose, and it's not bad practice to have them anyway.


gmarull:
merged resolving conflicts

PR #98, opened 2025-07-08
fw/drivers/display/sharp_ls013b7dh01_nrf5: use RTC instead of PWM

jwise:
The PWM implementation on nRF52 is power-hungry, so instead, we use GPIOTE's "toggle" mode, and our last remaining RTC, to pulse the Vcom pin on the LCD.

There should be power savings associated with this, but they appear to be hiding behind something else so far...


gmarull (changes requested):
RTC2 is also used by task watchdog, how does this change affect it?

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
```diff
@@ -154,10 +158,46 @@ void display_init(void) {
...
   prv_display_start();
 
-  // Generate PWM signal for EXTCOMIN (120Hz, ~100us pulse width)
-  pwm_init(&BOARD_CONFIG_DISPLAY.extcomin, 125000 / 120, 125000);
-  pwm_set_duty_cycle(&BOARD_CONFIG_DISPLAY.extcomin, (100U * 125000UL) / 1000000UL);
-  pwm_enable(&BOARD_CONFIG_DISPLAY.extcomin, true);
+  // Set up the RTC to tick at 60Hz, like Zephyr does.  Some boards will
+  // reuse the WDT RTC for this.
+  if (BOARD_CONFIG_DISPLAY.extcomin_rtc == BOARD_WATCHDOG_RTC_INST) {
+    PBL_ASSERTN(DISPLAY_TOGGLE_HZ == BOARD_WATCHDOG_RTC_FREQUENCY);
```
gmarull, line 164: nit: can likely be a static assert?

jwise: Apparently `BOARD_CONFIG_DISPLAY.extcomin_rtc != BOARD_WATCHDOG_RTC_INST` is not constant enough to be a static assert.  It would happen to work in this case if I wrote:
```c
  _Static_assert(BOARD_CONFIG_DISPLAY.extcomin_rtc != BOARD_WATCHDOG_RTC_INST ||
                 BOARD_WATCHDOG_RTC_FREQUENCY == DISPLAY_TOGGLE_HZ);
```
because the latter condition is 1 (and therefore constant), and if I change `DISPLAY_TOGGLE_HZ` in this case, it does fail, but it fails with:
```
../src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c: In function 'display_init':
../src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c:163:79: error: expression in static assertion is not constant
  163 |   _Static_assert(BOARD_CONFIG_DISPLAY.extcomin_rtc != BOARD_WATCHDOG_RTC_INST ||
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
  164 |                  BOARD_WATCHDOG_RTC_FREQUENCY == DISPLAY_TOGGLE_HZ);
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~             
```
which indicates that it would not work even if we chose a different `extcomin_rtc`.

gmarull:
Reverted in #106, see details there.

PR #100, opened 2025-07-08
DNM: QSPI: use controller's DPM interface to enter DPM

jwise:
Here's a first cut at enabling DPD, and in general, saving some power around QSPI.  Annoyingly, there are stability issues remaining.  Entering DPD reliably in stop mode could save us 80 uA or so.  And deactivating the QSPI interface entirely could save as much as 100 uA of peripheral current on top of that (measured with a "wfi forever" loop)!  This is a start at it, mostly to prove the idea out -- I went back to chasing other things to try to find where all the power was actually going, and did not complete this yet, but it is a place to start.


gmarull:
See #206

PR #102, opened 2025-07-08
drivers/pmic/npm1300: workaround nPM1300 erratum 27

jwise:
In the Asterix bootloader, we carefully set voltages to make sure that any ADC errors in nPM1300 do not mess up a running system.  Well, nPM1300 erratum 27 says that if we set the voltage to a voltage that is already set, then we get an extra 1 mA of quiescent current!  And indeed we do.  We work around this by turning the voltage up by 100 mV, and then back down again.  This is equivalent to if we had not ever tripped the erratum to begin with (I tested this by modifying the bootloader not to trip the erratum and measuring power).

Saves ~800 uA in OS sleep (on top of all other changes), saves ~780 uA in `wfi forever` floor state (floor current now 778 uA with this and all other changes).


gmarull (approved):
would be good to fix bootloader as well

ericmigi:
> would be good to fix bootloader as well

Should we do that before next production batch?

PR #115, opened 2025-07-15
asterix: reenable RTC-driven FreeRTOS timer, but keep sleep mode off [FIRM-264]

jwise:
This is the more conservative of two PRs.  It disables sleep mode entirely.  It should have no battery life improvement, but also, it should have no regressions in terms of Bluetooth stability.  It would be good to have quite a few watches shaking this down.


gmarull (approved):

gmarull:
Tested quickly, I cannot see the "reconnect regression" after flashing this branch.

@jwise for next commits, the `FIRM-XXX` can (needs to) go in the commit boddy as 'Fixes FIRM-XXX'

PR #116, opened 2025-07-15
 asterix: reenable RTC-driven FreeRTOS timer and allow sleep mode, but not STOP mode [FIRM-264]

jwise:
This is the more aggressive of two PRs.  It disables stop mode, but not sleep mode.  This will cause a meaningful and substantial improvement in battery life, but there is also a chance that there will be Bluetooth regressions (if there are, at least it'll make sense to me; if not, I will be extremely puzzled about a data point from earlier today).  Bluetooth regressions are still recoverable from PRF.  It would be nice for internal watches to shake this down, at least.


ericmigi:
@gmarull worth merging and cutting a release for internal

jwise:
incorporated in #120

gmarull:
can this be closed then?

PR #117, opened 2025-07-15
fw/services/voice: Implement Opus Codec

jplexer:
My work on an Opus implementation.

It compiles but it is not confirmed working yet (as I do not have a device to test on)


jplexer:
currently this doesnt compile on legacy (snowy, spalding, silk)
I can't for the life of it figure out of to get it working, would be nice if someone better at embedded than me could look at it :)

jplexer:
This really isn't working..

jplexer:
@ericmigi yeah that would really help! Sadly no tracking number 😔

PR #121, opened 2025-07-17
fw/drivers/sf32lb52: f32lb52 Solve the boot abnormality caused by the rng driver

w200024212:
rng driver  parameter is_seed need set to 1;


gmarull:
@w200024212 please add comment and we’ll then merge

w200024212:
> @w200024212 please add comment and we’ll then merge

ok!

w200024212:
Before generating random numbers, it is necessary to ensure that the seed is valid.So, I made some adjustments to the code.

gmarull:
merged here https://github.com/coredevices/PebbleOS/commit/44572c6d4f98f61dabbb5ad89f3372af997d145d

also refactored code as I was having difficulties in following the code.

PR #122, opened 2025-07-17
docs: add openocd dependency and change cd pebbles

ericmigi:
this was a blocker when I was helping them get it installed

### DIFF
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -37,7 +37,7 @@ sudo apt update
 2. Install required dependencies
 
 ```shell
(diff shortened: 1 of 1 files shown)


jwise:
I'll beef this up tomorrow when I'm back on Pebble stuff to describe the possibility to use a pebble-patched openocd for if you plan to actually debug with it, and to give it a signed-off-by (which is what gitlint is complaining about)

ericmigi:
I think if it works with the brew/apt `openocd` we should just stick with that for now, we can always change to a different one later right? just to keep things simple

gmarull:
commits need to be signed (-s)

ericmigi:
signed now?

PR #126, opened 2025-07-19
fw: STM mic support

jplexer:
I am currently debating again which codec to implement (in my tests opus takes up too much memory)

These commits are needed as a base for general voice support though!


jplexer:
@gmarull these are the STM32 commits, will open the rest in a minute!

src/fw/drivers/wscript_build
```diff
@@ -1164,11 +1164,32 @@ elif mcu_family == 'SF32LB52':
...
-if bld.capability('HAS_MICROPHONE'):
+if bld.is_snowy_compatible():
+    bld.objects(
+        name='driver_mic',
+        source=[
+            'mic/mic_command.c',
+            'mic/stm32/pdm.c',
+        ],
+        use=[
+            'PDMFilter_CM4_GCC',
```
gmarull: is this needed?

gmarull (approved):
just one nitpick, pls check, otherwise lgtm (I cannot test this though)

src/fw/drivers/wscript_build
jplexer: relic from the old pdm filter (i just imported the commit from the google repo), can remove that

PR #138, opened 2025-07-28
Implement support for Asterix's LSM6DSO accelerometer

matthewwardrop:
This patchset brings up initial working support for the IMU in Asterix (~Core~ Pebble 2 Duo).

I have not done too much low-level programming on as complex a device as Pebble (think Arduino with a module or two), but with a lot of help from AI (which I nudged along, sanity checked, pared back, correlated with Pebble HWV, and then restructured a little), I finally got something working (at least so far as that the accelerometer app I downloaded from the pebble store works correctly). I'm not sure how all this ladders up into step detection or movement activated backlights yet.

At this point, I plan to spend more time poking at and learning from this code, as well as cleaning up more of the AI slop... but likely won't prioritise this work unless I get some sort of signal from Core devs that this will ever be merged (to be clear, I have no real expectation of that occurring, since I bet you have things well in hand and will implement this far better than I).

But... I succeeded in this little challenge I set myself!

### DIFF
--- a/python_libs/pebble-loghash/pebble/loghashing/newlogging.py
+++ b/python_libs/pebble-loghash/pebble/loghashing/newlogging.py
@@ -201,7 +201,7 @@ def parse_message(msg, log_dict):
         return None
 
     # Python's 'printf' doesn't support %p. Sigh. Convert to %x and hope for the best
-    safe_output_msg = POINTER_FORMAT_TAG_PATTERN.sub('\g<format>x', output_dict['msg'])
+    safe_output_msg = POINTER_FORMAT_TAG_PATTERN.sub(r'\g<format>x', output_dict['msg'])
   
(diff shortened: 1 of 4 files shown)


gmarull (changes requested):
please split changes into smaller commits so this can be reviewed easily. Also, implementing stuff in `imu_asterix.c` is wrong, this IMU can be used in other platforms.

matthewwardrop:
Hi @gmarull ! Thanks for your feedback! I'll take that as an expression of interest in this PR getting cleaned up for an eventual merge (originally, as indicated above, this was just a challenge I set myself for my own edification to flesh out the `imu_asterix.c` file to a working state). I'll take some time to process the code and properly prepare it for merging, including splitting up the code and cleaning up some of the AI generated slop.

PR #140, opened 2025-07-29
Driver implementation for Asterix's mmc5603nj magnetometer

Bob-Wei1:
This adds a mmc5603nj driver for the asterix board. The driver uses timer based polling as this magnetometer lacks a data ready pin that allows for an interrupt based driver

Currently there is a problem with the calibration sequence on the compass app where the initial heading is incorrect but everything else about the driver is functional.

### DIFF
--- a/src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
+++ b/src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
@@ -63,12 +63,11 @@ static bool s_polling_active = false;
 
 
 static bool mmc5603nj_read(uint8_t reg_addr, uint8_t data_len, uint8_t *data) {
-  //PBL_LOG(LOG_LEVEL_DEBUG, "I2C read: device=MMC5603NJ, reg=0x%02x, len=%d", reg_addr, data_len);
+  
   i2c_use(I2C_MMC5603NJ);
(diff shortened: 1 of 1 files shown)


src/fw/board/boards/board_asterix.h
```diff
@@ -94,15 +94,15 @@ static const BoardConfigPower BOARD_CONFIG_POWER = {
...
+    .axes_offsets[AXIS_X] = 0,  // X (North) now maps to sensor X axis
+    .axes_offsets[AXIS_Y] = 1,  // Y (East) now maps to sensor Y axis
     .axes_offsets[AXIS_Z] = 2,
     .axes_inverts[AXIS_X] = false,
     .axes_inverts[AXIS_Y] = true,
     .axes_inverts[AXIS_Z] = true,
 #else
-    .axes_offsets[AXIS_X] = 1,
-    .axes_offsets[AXIS_Y] = 0,
+    .axes_offsets[AXIS_X] = 0,  // X (North) now maps to sensor X axis
```
matthewwardrop, line 104: These axis configurations were not correct in my testing. I think we want:

X=1, Y=0, Z=2 (as per the state pre this patch); with no inversion of the x and y axes. The z-axis is doing something funky, and doesn't appear to be signed in the way that I expect. But with this configuration, I have things working for reasonably nicely on my devkit.

PR #145, opened 2025-07-31
ble: Add LXT_DISABLE micro definition required by SiFli SDK

gangheivt:


gmarull (changes requested):
`git grep --recurse-submodules LXT_DISABLE` shows this used in files that are never compiled/used, so why is this exactly needed?

PR #147, opened 2025-08-02
Implement driver for Asterix's LSM6DSO IMU

matthewwardrop:
This patchset will implement full support for the LSM6DSO IMU found in Pebble 2 Duo (aka Asterix). IIUC, this is also the IMU found in Pebble Time 2 (aka Obelix), and so it should be straightforward to enable this driver for Obelix at some later point. Since I don't have Obelix hardware to play with, that is out of scope for this PR.

This PR is current in draft state, and will rebased as necessary to keep things clean (and for my own benefit, updated against latest main code).

**Current state**:

- [x] Accelerometer sampling.
- [x] Interrupt driven accelerometer updates.
- [x] Check and fix alignment and scaling of accelerometer axes (as defined [here](https://developer.rebble.io/developer.pebble.com/guides/events-and-services/accelerometer/index.html)).
- [ ] Use FIFO buffers on LSM6DSO when `n_samples` > 1.
- [x] Double tap detection.
- [x] (partial) Shake detection.
- [ ] Ignore tap interrupts when vibing.
- [ ] Sign off all commits.
- [ ] ...?

(This PR replaces my earlier exploratory work in #138, which I will now close).

### DIFF
--- a/src/fw/board/boards/board_asterix.h
+++ b/src/fw/board/boards/board_asterix.h
@@ -59,6 +59,29 @@ static const BoardConfigActuator BOARD_CONFIG_VIBE = {
   .vsys_scale = 3300,
 };
 
+static const BoardConfigAccel BOARD_CONFIG_ACCEL = {
+  .accel_config = {
+    .axes_offsets[AXIS_X] = 1,
+    .axes_offsets[AXIS_Y] = 0,
+    .axes_offsets[AXIS_Z] = 2,
+    .axes_inverts[AXIS_X] = false,
+    .axes_inverts[AXIS_Y] = false,
+    .axes_inverts[AXIS_Z] = false,
+    // This will need calibration.
+    .shake_thresholds[AccelThresholdHigh] = 64,
+    .shake_thresholds[AccelThresholdLow] = 0xf,
+    .double_tap_threshold = 12500,
+  },
+  // Ideally we would configure both interrupt pins, but we have run out of GPIOTE channels.
+  // We will use INT1 (connected to pin 13) for accelerometer interrupts, and leave INT2 (pin 11) unused.
+  .accel_int_gpios = {
+    [0] = { .gpio = NRF5_GPIO_RESOURCE_EXISTS, .gpio_pin = NRF_GPIO_PIN_MAP(1, 13) },
+  },
+  .accel_ints = {
+    [0] = { .peripheral = NRFX_GPIOTE_INSTANCE(0), .channel = 7, .gpio_pin = NRF_GPIO_PIN_MAP(1, 13) },
+  },
+};
+
 extern UARTDevice * const DBG_UART;
 
 extern PwmState BACKLIGHT_PWM_STATE;
--- a/src/fw/drivers/imu/imu_asterix.c
+++ b/src/fw/drivers/imu/imu_asterix.c
@@ -1,10 +1,11 @@
 #include "board/board.h"
-
 #include "drivers/accel.h"
+#include "drivers/i2c.h"
 #include "drivers/imu.h"
+#include "drivers/imu/lsm6dso/lsm6dso.h"
 #include "drivers/spi.h"
 #include "kernel/util/delay.h"
-#include "drivers/i2c.h"
+#include "system/logging.h"
 
 #define MMC5603_PRODUCT_ID 0x39
 #define MMC5603_PRODUCT_ID_VALUE 0x10
@@ -13,17 +14,6 @@
 #define BMP390_CHIP_ID 0x00
 #define BMP390_CHIP_ID_VALUE 0x60
 #define BMP390_PWR_CTRL 0x1B
-
-#define LSM6D_FUNC_CFG_ACCESS 0x01
-
-#define LSM6D_WHO_AM_I 0x0F
-#define LSM6D_WHO_AM_I_VALUE 0x6C
-
-#define LSM6D_CTRL1_XL 0x10
-#define LSM6D_CTRL2_G 0x11
-#define LSM6D_CTRL4_C 0x13
-#define LSM6D_CTRL4_C_SLEEP_G 0x40
-
 
 
 static bool prv_read_register(I2CSlavePort *i2c, uint8_t register_address, uint8_t *result) {
@@ -63,23 +53,13 @@ void imu_init(void) {
     (void) prv_write_register(I2C_BMP390, BMP390_PWR_CTRL, 0);
   }
   
-  rv = prv_read_register(I2C_LSM6D, LSM6D_WHO_AM_I, &result);
-  if (!rv || result != LSM6D_WHO_AM_I_VALUE) {
-    PBL_LOG(LOG_LEVEL_DEBUG, "LSM6DSO probe failed; rv %d, result 0x%02x", rv, result);
-  } else {
-    PBL_LOG(LOG_LEVEL_DEBUG, "found the LSM6DSO, setting to lower power");
-    (void) prv_write_register(I2C_LSM6D, LSM6D_FUNC_CFG_ACCESS, 0);
-    (void) prv_write_register(I2C_LSM6D, LSM6D_CTRL1_XL, 0);
-    (void) prv_write_register(I2C_LSM6D, LSM6D_CTRL2_G, 0);
-    (void) prv_write_register(I2C_LSM6D, LSM6D_CTRL4_C, LSM6D_CTRL4_C_SLEEP_G);
-  }
+  lsm6dso_init();
 }
 
 void imu_power_up(void) {
-  // NYI
+  lsm6dso_power_up();
 }
 
 void imu_power_down(void) {
-  // NYI
+  lsm6dso_power_down();
 }
-
--- a/src/fw/drivers/imu/lsm6dso/lsm6dso.c
+++ b/src/fw/drivers/imu/lsm6dso/lsm6dso.c
@@ -31,6 +31,10 @@ static int32_t prv_lsm6dso_write(void *handle, uint8_t reg_addr, const uint8_t *
                                  uint16_t write_size);
 static void prv_lsm6dso_mdelay(uint32_t ms);
 static void prv_lsm6dso_init(void);
+static void prv_lsm6dso_chase_target_state(void);
+static void prv_lsm6dso_configure_interrupts(void);
+static void prv_lsm6dso_interrupt_handler(bool *should_context_switch);
+static void prv_lsm6dso_process_interrupts(void);
 
 // HAL context for LSM6DSO
 stmdev_ctx_t lsm6dso_ctx = {
@@ -42,6 +46,18 @@ stmdev_ctx_t lsm6dso_ctx = {
 // Toplevel module state
 
 static bool s_lsm6dso_initialized = false;
+static bool s_lsm6dso_enabled = true;
+static bool s_lsm6dso_running = false;
+typedef struct {
+  uint32_t sampling_interval_us;
+  uint32_t num_samples;
+  bool shake_detection_enabled;
+  bool shake_sensitivity_high;
+  bool double_tap_detection_enabled;
+} lsm6dso_state_t;
+lsm6dso_state_t s_lsm6dso_state = {0};
+lsm6dso_state_t s_lsm6dso_state_target = {0};
+static bool s_interrupts_pending = false;
 
 // LSM6DSO configuration entrypoints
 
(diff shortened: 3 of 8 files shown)


wscript
```diff
@@ -520,6 +520,7 @@ def configure(conf):
         conf.env.append_value('DEFINES', ['BT_CONTROLLER_CC2564X'])
     elif conf.is_asterix():
         conf.env.bt_controller = 'nrf52'
+        conf.env.imu_sensor = 'lsm6dso'
```
jwise: my recommendation: I think that the `env.bt_controller` thing is kind of an anti-pattern that should (and will, now that we are all NimBLE!) die.  so I think things like this probably should move to `platform_capabilities` for now -- for instance, `asterix` and `obelix` probably should have `CAPABILITY_HAS_LSM6DSO` or something

src/fw/board/boards/board_asterix.h
```diff
@@ -141,17 +139,16 @@ static const BoardConfigAccel BOARD_CONFIG_ACCEL = {
     .shake_thresholds[AccelThresholdLow] = 0xf,
     .double_tap_threshold = 12500,
   },
-  .accel_int_gpios = {
-    [0] = { GPIOA, GPIO_Pin_6 },
-    [1] = { GPIOA, GPIO_Pin_3 },
+  // TODO: Verify that there is only one INT pin connected to the LSM6DSO
```
matthewwardrop: @gmarull Is this correct?

wscript
matthewwardrop: Happy to follow any configuration pattern. If you make this change in main, I'll rebase my branch to follow the same pattern.

gmarull: The build system is in general a big mess; we haven't spent time cleaning it up, and as of now it's growing organically from where it was left off ~10 years ago. I think though, that the platform capabilities way is wrong, as from what I get, it signals availability of interfaces, not any particular HW.

src/fw/board/boards/board_asterix.h
gmarull: image

matthewwardrop: Ah.... Thank you! We have pin 11 for Int2.

gmarull:
> @gmarull: Locally I now have double tap and shake detection working (technically I used the finite state machine to detect wrist tilt, which isn't actually the right implementation given the API... but the experience is nice for screen wakeup; I'll change it to properly detect the shake prior to pushing up code).
> 
> However, just a few quick clarification. Although pin 11 is connected to INT2, it appears that we are out of GPIOTE channels to actually leverage this pin. Am I missing something?
> 
> This matters right now because there seems to be some interference running all of the interrupts via INT1 (in particular the double tap detection causes further interrupts not to be sent when used simultaneously with regular data collection... there's some kind of internal delay inside of the LSM6DSO that causes the interrupt not to be reset after being read; like perhaps the ODR is not aligned with the double tap reset clock). I've solved this for now with a separate external trigger which clears the state on a regular interval, but it is really not ideal. I'll continue fiddling... but a separate interrupt would likely solve this.

Using a single line should likely be enough. I guess the 2 were routed just in case.

matthewwardrop:
Solved the issue: the drdy interrupts need to be separately set to pulsed interrupt mode (I had thought that setting "ALL" events to pulsed would do this already).

matthewwardrop:
@gmarull This is now to a point where things are working pretty well on my hardware. It is definitely still not perfect or well calibrated, but I'd love to have some formal review before I go too much further.

third_party/wscript
```diff
@@ -35,4 +35,7 @@ def build(bld):
     if bld.env.bt_controller == 'cc2564x':
         bld.recurse('ti_bt_sp')
 
+    if bld.env.imu_sensor == 'lsm6dso':
+        bld.recurse('lsm6dso')
```
gmarull: for now: just use `if bld.is_asterix()`, we'll eventually make build system better using e.g. Kconfig+CMake

src/fw/board/boards/board_asterix.h
```diff
@@ -59,6 +59,36 @@ static const BoardConfigActuator BOARD_CONFIG_VIBE = {
...
+static const BoardConfigAccel BOARD_CONFIG_ACCEL = {
+  .accel_config = {
+    .axes_offsets[AXIS_X] = 0,
+    .axes_offsets[AXIS_Y] = 1,
+    .axes_offsets[AXIS_Z] = 2,
+#if IS_BIGBOARD
+    .axes_inverts[AXIS_X] = false,
+    .axes_inverts[AXIS_Y] = false,
+    .axes_inverts[AXIS_Z] = false,
+#else
```
gmarull: there's no bigboard

src/fw/board/boards/board_asterix.h
```diff
@@ -59,6 +59,36 @@ static const BoardConfigActuator BOARD_CONFIG_VIBE = {
...
+    .double_tap_threshold = 12500,
+  },
+  // Ideally we would configure both interrupt pins, but we have run out of GPIOTE channels.
+  .accel_int_gpios = {
+    [0] = { .gpio = NRF5_GPIO_RESOURCE_EXISTS, .gpio_pin = NRF_GPIO_PIN_MAP(1, 13) },
+    // [1] = { .gpio = NRF5_GPIO_RESOURCE_EXISTS, .gpio_pin = NRF_GPIO_PIN_MAP(1, 11) },
+  },
+  .accel_ints = {
+    [0] = { .peripheral = NRFX_GPIOTE_INSTANCE(0), .channel = 7, .gpio_pin = NRF_GPIO_PIN_MAP(1, 13) },
+    // [1] = { .peripheral = NRFX_GPIOTE_INSTANCE(0), .channel = ??, .gpio_pin = NRF_GPIO_PIN_MAP(1, 11) },
```
gmarull: remove commented code

src/fw/drivers/wscript_build
```diff
@@ -465,14 +465,15 @@ if bld.is_silk() and not bld.env.QEMU:
     )
 
 if bld.is_asterix() or bld.is_obelix():
-    # FIXME(ASTERIX,OBELIX): provide working implementation
```
gmarull, line 468: note: since this branch currently applies to obelix, you'll need to modify its board as well for things to compile. I'd suggest to split into asterix, leave obelix with stubs for now.

src/fw/drivers/wscript_build
```diff
@@ -465,14 +465,15 @@ if bld.is_silk() and not bld.env.QEMU:
...
     bld.objects(
         name='driver_lsm6dso',
         source=[
-            'stubs/accel.c',
+            'imu/lsm6dso/lsm6dso.c',
         ],
         use=[
             'fw_includes',
+            'lsm6dso',
+            'lsm6dso_includes',
```
gmarull: name library `hal_lsm6dso`, no need to use `_includes` if original target exports includes.

third_party/lsm6dso/wscript
```diff
@@ -0,0 +1,17 @@
+def build(bld):
+    lsm6dso_includes = [
+        'lsm6dso-pid'
+    ]
+
+    lsm6dso_sources = [
+        'lsm6dso-pid/lsm6dso_reg.c',
+    ]
+
+    bld(export_includes=lsm6dso_includes, name='lsm6dso_includes')
```
gmarull: is this needed?

third_party/lsm6dso/wscript
```diff
@@ -0,0 +1,17 @@
...
+
+    bld.stlib(source=lsm6dso_sources,
+              includes=lsm6dso_includes,
+              export_includes=lsm6dso_includes,
+              target='lsm6dso',
```
gmarull: `hal_lsm6dso`, also name folder accordingly.

src/fw/drivers/imu/imu_asterix.c
```diff
@@ -1,10 +1,12 @@
 #include "board/board.h"
-
+#include "board/boards/board_asterix.h"
```
gmarull: this should not be needed

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -16,56 +16,370 @@
...
+
+static int32_t prv_lsm6dso_write(void *handle, uint8_t reg_addr, const uint8_t *buffer,
+                                 uint16_t write_size) {
+  i2c_use(I2C_LSM6D);
+  uint8_t d[write_size + 1];
+  d[0] = reg_addr;
+  memcpy(&d[1], buffer, write_size);
+  bool result = i2c_write_block(I2C_LSM6D, write_size + 1, d);
+  i2c_release(I2C_LSM6D);
+  return !result;
```
gmarull: return type is not boolean, but `int32_t`

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -16,56 +16,370 @@
...
+  if (s_lsm6dso_state.num_samples > 0) {
+    accel_cb_new_sample(data);
+  }
+
+  return 0;
+}
+
+// Other utility functions
+
+static void prv_delay_ms(uint32_t ms) { delay_us(ms * 1000); }
```
gmarull: `psleep`?

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -132,7 +138,7 @@ void lsm6dso_init(void) {
   }
 
   // Set default full scale
-  if (lsm6dso_xl_full_scale_set(&lsm6dso_ctx, LSM6DSO_2g)) {
+  if (lsm6dso_xl_full_scale_set(&lsm6dso_ctx, LSM6DSO_4g)) {
```
gmarull, line 242: not related to commit changes?

src/fw/board/boards/board_asterix.h
```diff
@@ -61,18 +61,12 @@ static const BoardConfigActuator BOARD_CONFIG_VIBE = {
...
     .axes_offsets[AXIS_Z] = 2,
-#if IS_BIGBOARD
     .axes_inverts[AXIS_X] = false,
     .axes_inverts[AXIS_Y] = false,
     .axes_inverts[AXIS_Z] = false,
-#else
-    .axes_inverts[AXIS_X] = true,
-    .axes_inverts[AXIS_Y] = true,
-    .axes_inverts[AXIS_Z] = true,
-#endif
```
gmarull: fixup original commit

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -25,22 +25,25 @@
 #include "lsm6dso_reg.h"
 #include "lsm6dso.h"
 
-// Forward declaration of private functions defined below public functions
+#define MIN(a, b) ((a) < (b) ? (a) : (b))
```
gmarull: already in `src/libutil/includes/util/math.h`

gmarull: also please avoid combo commits like 07d4f866c66cbefb29c512c5f9612d45aa4a82a7 where many things are changed, making review difficult.

src/fw/board/boards/board_asterix.h
matthewwardrop: This got dropped in a later patch when IMU axes were aligned (and was a stayover here from the previous state of things that was brought in through a rebase). Will iron it out of this patch.

src/fw/board/boards/board_asterix.h
matthewwardrop: It was left in as documentation around pin 11. Should I add that as a comment, or just remove knowledge of pin 11 from the code here altogether?

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: I think I had some trouble with psleep for some reason. Will try again.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: Will definitely clean this up! Was just worried that the approach I was taking was in some other way objectionable and I was wasting my time.

third_party/lsm6dso/wscript
matthewwardrop: Likely not. I was copying some other modules which likely did need it. Will remove.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: Ah... I had to shim it anyway to comply with hal_lsm6dso, so just used the function with a name. I've changed it to `psleep`.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: I was cheating here and using the automatic cast of false to zero; but I'll make this more explicit.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: Will resplit out changes into logical blocks.

src/fw/drivers/imu/lsm6dso/lsm6dso.h
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2025 Joshua Jun
```
jplexer: nit: you should probably change this to your own name
I didn't author this and shouldn't take the credit for it :)

matthewwardrop: Whoops... Not sure how that snuck in there!

matthewwardrop:
Will fix the prf builds.

PR #152, opened 2025-08-05
Make the SDK work

griffinli:
This updates the SDK code to be compatible with Python 3 and Waf 2.1.4, so that the SDK generated by a build of PebbleOS can be used to compile watch apps.

Known issues:
- Watch apps that use Pebble Packages will compile but error out at runtime.


gmarull:
It would probably be good to have a CI job that builds SDK, so it does not bit rot?

gmarull (approved):

PR #176, opened 2025-08-13
fw: implement lsm6dso IMU

jplexer:
Finished implementing the IMU driver based on #147

@gmarull please review :)


src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -432,6 +437,63 @@ static void prv_lsm6dso_process_interrupts(void) {
...
+      const AccelConfig *cfg = &BOARD_CONFIG_ACCEL.accel_config;
+      if (wake_src.x_wu) {
+        axis = AXIS_X;
+      } else if (wake_src.y_wu) {
+        axis = AXIS_Y;
+      } else if (wake_src.z_wu) {
+        axis = AXIS_Z;
+      }
+      // Read current sample to infer direction
+      int16_t accel_raw[3];
```
matthewwardrop, line 537: Is this sufficient? The direction will almost always be biased in the direction of gravity, which is not necessarily the shake direction. If you take this approach, you'll likely need a two-sample delta at least?

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -78,6 +78,62 @@ lsm6dso_state_t s_lsm6dso_state = {0};
 lsm6dso_state_t s_lsm6dso_state_target = {0};
 static bool s_interrupts_pending = false;
 static uint32_t s_tap_threshold = BOARD_CONFIG_ACCEL.accel_config.double_tap_threshold / 1250;
+static bool s_fifo_in_use = false; // true when we have enabled FIFO batching
+
+// Maximum FIFO watermark supported by hardware (diff_fifo is 10 bits -> 0..1023)
+#define LSM6DSO_FIFO_MAX_WATERMARK 1023
+
+// Map output data rate (interval) to FIFO batching rate enum
+static lsm6dso_bdr_xl_t prv_get_fifo_batch_rate(uint32_t interval_us) {
```
matthewwardrop: nit: I was pretty careful in my PR to group functions based on their associated interface implementation or functionality; with private methods below all interface methods. This doesn't conform with that convention; fwiw.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -534,9 +602,58 @@ static int32_t prv_lsm6dso_set_sampling_interval(uint32_t interval_us) {
 // Accelerometer sample reading (and reporting)
 
 static void prv_lsm6dso_read_samples(void) {
-  // TODO: Properly implement FIFO buffer.
-  AccelDriverSample sample;
-  prv_lsm6dso_read_sample(&sample);
+  if (s_lsm6dso_state.num_samples <= 1 || !s_fifo_in_use) {
+    // Legacy single sample path
```
matthewwardrop: nit: Probably could remove the "legacy".

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -696,3 +696,139 @@ static uint64_t prv_get_timestamp_ms(void) {
...
+  s_lsm6dso_state_target.sampling_interval_us = 19231; // ~52Hz per mapping
+  s_lsm6dso_state_target.num_samples = 0; // disable callbacks during test
+  s_lsm6dso_state_target.shake_detection_enabled = false;
+  s_lsm6dso_state_target.double_tap_detection_enabled = false;
+  prv_lsm6dso_chase_target_state();
+
+  // Force FS=4g (required for mg conversion helper used elsewhere)
+  lsm6dso_xl_full_scale_set(&lsm6dso_ctx, LSM6DSO_4g);
+
+
```
matthewwardrop: nit: whitespace

matthewwardrop (commented):
Hi @jplexer !

I was having fun implementing this, albeit slowly, and figuring things out as I went. Holding down a full time job with four kids doesn't leave a lot of time to play. But thanks for helping to get this across the line.

These changes all look pretty sensible to me. I left some comments where things were perhaps not ideal. We may also want implement enabling the sensor briefly during an `accel_peek` when the device was otherwise enabled but not running; before chasing the original target state. Should only be a few lines of extra code.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
jplexer: in my testing I didn't really get false positive so I guess it is alright? feel free to correct me though!

gmarull:
> Hi @jplexer !
> 
> I was having fun implementing this, albeit slowly, and figuring things out as I went. Holding down a full time job with four kids doesn't leave a lot of time to play. But thanks for helping to get this across the line.
> 
quitarse el sombrero diego fournier


Thank you so much for the effort! Impressed this was done so fast with 4 kids :smile: I will review/test tomorrow.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -0,0 +1,837 @@
...
+// Forward declaration of private functions defined below public functions
+static int32_t prv_lsm6dso_read(void *handle, uint8_t reg_addr, uint8_t *buffer,
+                                uint16_t read_size);
+static int32_t prv_lsm6dso_write(void *handle, uint8_t reg_addr, const uint8_t *buffer,
+                                 uint16_t write_size);
+static void prv_lsm6dso_mdelay(uint32_t ms);
+static void prv_lsm6dso_init(void);
+static void prv_lsm6dso_chase_target_state(void);
+static void prv_lsm6dso_configure_interrupts(void);
+static void prv_lsm6dso_configure_double_tap(bool enable);
```
matthewwardrop, line 40: nit: suggest moving the fifo method declarations to just after `*_configure_interrupts`; both here and in the implementation; to keep everything grouped logically.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
matthewwardrop: I expect you would potentially be off if in your direction if your threshold were less than about ~1000mg, a threshold that would be unreasonably high. I'll do some testing later and see.

src/fw/board/boards/board_asterix.h
```diff
@@ -67,10 +67,10 @@ static const BoardConfigAccel BOARD_CONFIG_ACCEL = {
     .axes_inverts[AXIS_X] = false,
     .axes_inverts[AXIS_Y] = false,
     .axes_inverts[AXIS_Z] = false,
+    .double_tap_threshold = 12500, //This already is a good value and does not need calibration.
     // This will need calibration.
     .shake_thresholds[AccelThresholdHigh] = 64,
     .shake_thresholds[AccelThresholdLow] = 0xf,
-    .double_tap_threshold = 12500,
```
gmarull: unnecessary move

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -358,6 +365,33 @@ static void prv_lsm6dso_configure_interrupts(void) {
...
+
+    // Enable tap detection on all axes
+    lsm6dso_tap_detection_on_x_set(&lsm6dso_ctx, PROPERTY_ENABLE);
+    lsm6dso_tap_detection_on_y_set(&lsm6dso_ctx, PROPERTY_ENABLE);
+    lsm6dso_tap_detection_on_z_set(&lsm6dso_ctx, PROPERTY_ENABLE);
+
+    // Configure tap timing
+    lsm6dso_tap_shock_set(&lsm6dso_ctx, 0x03);  // Shock duration
+    lsm6dso_tap_quiet_set(&lsm6dso_ctx, 0x02);  // Quiet period
+    lsm6dso_tap_dur_set(&lsm6dso_ctx, 0x08);    // Double tap window
```
gmarull: magic numbers, should also be board dependent

gmarull: add to accel config, add comments explaining their meaning

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -398,6 +456,11 @@ static int32_t prv_lsm6dso_set_sampling_interval(uint32_t interval_us) {
     return -1;
   }
 
+  if (s_lsm6dso_state.double_tap_detection_enabled) {
+    interval_us =
+        MIN(interval_us, 2398);  // Max interval for shake/double tap detection to work reliably
+  }
```
gmarull: magic number, board dependent?

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -534,9 +604,58 @@ static int32_t prv_lsm6dso_set_sampling_interval(uint32_t interval_us) {
...
+      break;
+    }
+
+    uint8_t raw_bytes[6];
+    if (lsm6dso_read_reg(&lsm6dso_ctx, LSM6DSO_FIFO_DATA_OUT_X_L, raw_bytes, sizeof(raw_bytes)) != 0) {
+      PBL_LOG(LOG_LEVEL_ERROR, "LSM6DSO: Failed to read FIFO sample (%u/%u)", i, fifo_level);
+      break;
+    }
+
+    if (tag != LSM6DSO_XL_NC_TAG && tag != LSM6DSO_XL_NC_T_1_TAG && tag != LSM6DSO_XL_NC_T_2_TAG && tag != LSM6DSO_XL_2XC_TAG && tag != LSM6DSO_XL_3XC_TAG) {
```
gmarull: loooong line, maybe last commit a clang format (to avoid entering a rebase nightmare)

src/fw/drivers/imu/lsm6dso/lsm6dso.c
jplexer: Could be moved to board config if you expect per-board variation (different sensor or power tuning), but it’s fundamentally a sensor algorithm constraint (needs ≥417 Hz ODR) rather than a board-specific parameter.

src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -74,6 +79,14 @@ typedef struct {
...
 lsm6dso_state_t s_lsm6dso_state_target = {0};
 static bool s_interrupts_pending = false;
+static uint32_t s_tap_threshold = BOARD_CONFIG_ACCEL.accel_config.double_tap_threshold / 1250;
+static bool s_fifo_in_use = false;  // true when we have enabled FIFO batching
+
+// Maximum FIFO watermark supported by hardware (diff_fifo is 10 bits -> 0..1023)
+#define LSM6DSO_FIFO_MAX_WATERMARK 1023
+
+// Maximum allowed sampling interval (i.e., slowest rate, in microseconds)
+#define LSM6DSO_EVENT_MAX_INTERVAL_US 2398
```
matthewwardrop, line 89: nit: Should this be named more closely to its function: something like `LSM6DSO_MAX_INTERVAL_US_FOR_TAP_DETECTION` ? Obviously not a huge deal.

gmarull: let's do extra tweaking in a follow-up PR

gmarull (approved):

PR #182, opened 2025-08-14
reenable STOP mode on Asterix [FIRM-307]

jwise:
Originally, we used `new_timer` for NPL callouts in NimBLE.  `new_timer` was not safe to call from ISRs, though, and NimBLE set up callout timers from ISRs on nRF5!  So we added a quick hack to use FreeRTOS Timers, which *were* safe to call from ISRs ... but the only problem is that the tick count in FreeRTOS itself was not always accurate with regard to the RTC, especially around stop mode intervals on Asterix.

We solve this by fixing the `new_timer` version to enqueue an event for later if we happen to be running from inside an ISR.  Additionally, we avoid the RTC timer getting way ahead of us by ensuring that we can actually go to sleep for multiple ticks in a STOP, and we have NimBLE disable STOP mode when needed to keep wake latency low.

This takes us from 1.29 mA connected standby to 1.20mA connected standby, since we can now put flash in DPD in STOP mode.


third_party/nimble/port/include/nimble/nimble_npl_os.h
```diff
@@ -176,7 +189,7 @@ static inline void ble_npl_callout_set_arg(struct ble_npl_callout *co, void *arg
   co->ev.arg = arg;
 }
 
-static inline uint32_t ble_npl_time_get(void) { return xTaskGetTickCountFromISR(); }
+static inline uint32_t ble_npl_time_get(void) { return rtc_get_ticks(); }
```
gmarull: note: ticks are 64bit, we're casting down here, is this safe?

third_party/nimble/port/include/nimble/nimble_npl_os.h
```diff
@@ -62,6 +62,19 @@ struct ble_npl_callout {
   TimerHandle_t handle;
 #else
   TimerID handle;
+
+  /// whether we have made changes to this timer from ISR state and need
+  /// someone to go sync them back up on the main thread for us
```
gmarull: in pebbleos `//!` is normally used

third_party/nimble/port/src/npl_os_pebble.c
```diff
@@ -305,9 +306,39 @@ npl_pebble_callout_get_ticks(struct ble_npl_callout *co) {
...
+static void os_callout_timer_cb(void *timer);
+
+static void npl_pebble_callout_do_update(struct ble_npl_callout *co) {
+  if (!co->update_pending) {
+    return;
+  }
+
+  int rem_ticks = co->update_target_time_ticks - rtc_get_ticks();
+  if (rem_ticks < 1) {
+    // Oh, well...
```
gmarull, line 264: explain clamping better here, can this really happen? maybe assert?

third_party/nimble/port/src/npl_os_pebble.c
```diff
@@ -323,16 +354,28 @@ void npl_pebble_callout_init(struct ble_npl_callout *co, struct ble_npl_eventq *
...
+
+  co->update_pending = 1;
+  co->update_target_time_ticks = rtc_get_ticks() + ticks;
+
+  // we may be in an ISR; if we are, we make the event queue actually
+  // schedule the timer for us later.  otherwise, we can do it ourselves.
+  if (mcu_state_is_isr()) {
+    co->update_ev.fn = npl_pebble_callout_do_update_from_eventq;
+    co->update_ev.arg = (void *)co;
+    ble_npl_eventq_put(nimble_port_get_dflt_eventq(), &co->update_ev);
```
gmarull, line 324: AFAIK this will put the event into the host queue, which gets consumed by the host task. I guess FreeRTOS timers push it to the timer task queue, which runs at a higher priority than the host task. So if I read correctly, here the timer will be scheduled when the host task has a chance to run, whereas before, the timer task would handle it directly.

third_party/nimble/port/include/nimble/nimble_npl_os.h
jwise: good point, I need to check how this wraps.  if it is a problem here, though, it would have always been a problem

third_party/nimble/port/src/npl_os_pebble.c
jwise: I will beef up this comment to explain what's going on here, but here is the sequence in which this can happen:

* `npl_pebble_callout_reset` gets called in an ISR with a very short number of `ticks`
* the `update_ev` is put on the event queue to actually go update the callout
* something higher priority is happening, and the `update_ev` takes more than `ticks` to fire
* the `update_ev` is eventually triggered, finally completing the `npl_pebble_callout_do_update`
* but now the remaining time is negative!  obviously we cannot pass a negative number of milliseconds to `new_timer`, for it takes a `uint32_t timeout_ms`
* but we do want to defer it so it's not happening in the event queue thread, but in the same thread as it "always" runs in, just to avoid having exceptionally rare codepaths where most of the time callouts run in the `new_timer` thread but rarely they run in the NPL eventq thread.  so we defer it for the minimum amount of time possible, so that it runs as a `NewTimerCallback`

third_party/nimble/port/src/npl_os_pebble.c
jwise: well, ideally timers should be a decent amount of time into the future.  so indeed the timer will be enqueued to be scheduled whenever the host task has a chance to run (and then it will eventually be put into the new_timer queue), but if the system was heavily loaded enough that the timer would be late in this case, it probably would be late for whatever host work needs to be done later anyway.  I looked at where callouts are used and they do not seem to be all that critical in terms of timing, anyway.

third_party/nimble/port/src/npl_os_pebble.c
jwise: FWIW, I expect that this should nearly-never happen.  But it is good to be able to handle it.

third_party/nimble/port/include/nimble/nimble_npl_os.h
jwise: I spent a while sifting through this, and I'm not entirely convinced it's correct.  *But*, if it's incorrect, there's nothing we can do about it, so I wrote a comment and shrugged my shoulders.  Pushed this comment.

third_party/nimble/port/include/nimble/nimble_npl_os.h
jwise: Updated.

third_party/nimble/port/src/npl_os_pebble.c
jwise: Wrote approximately this into a comment.

third_party/nimble/port/src/npl_os_pebble.c
jwise: Clarified this in a comment.

jwise:
But this does give me an idea of where to look.  The "no intent for connection" thing seems to happen when the watch's bonding state is somehow out of sync with the phone's.  I wonder if somehow we are failing to actually bring the flash out of DPD sometimes, and we get the wrong results when reading a pairing from the gapdb.  The repro sequenc eI can find seems to be:

* delete bonding from phone BLE settings to repair the state that it's stuck in
* allow the app to re-bond
* repeatedly `waf flash` until it once again gets stuck `Negotiating`

I looked back to see how we worked around nRF52840 erratum 244, and apparently we reverted our fix for that for some reason?  I brought it back, and that did not seem to solve it for me (but it did make the system extremely slow as we turned HFXO on and off all the time).  I made a change to make HFXO turn on and off with flash DPD.

I nopped out flash_power_down_for_stop_mode and flash_power_up_for_stop_mode in stop.c, and we went back to watchdog timers firing regularly, which makes no sense to me.

I reverted that change, and found watchdog timers firing regularly again.  I started to suspect that something else was even more broken than I knew.

I reconfigured with `--nostop` and ran `./waf flash` about 5 or 6 times; I no longer had the WDT firing in any of the cases and successfully connected each time.  I came to the sinking conclusion that whatever I was chasing had nothing to do with the changes I was making, and just happened to happen infrequently enough that it only felt like it did.

I reconfigured without `--nostop`.  I at least tripped over the 'no intent for connection' error.  I sighed deeply, rubbed my eyes, and declared this a problem for tomorrow.

jwise:
I have bad news about the new_timer code, yes.

The `new_timer` is actually just a very specific `task_timer` inside.  (Actually, it is the only `task_timer`!)  But doing this, I think, would require substantially reworking `task_timer` either to run mostly with interrupts off, or to teach most of its accessors to use lock-free accessors for linked lists.  And I don't see much in the way of test infrastructure around `task_timer` -- or, for that matter, around even `new_timer` -- in `tests/`.  So I feel a little nervous about modifying that code without something that I can run to bang on it overnight to look for any race conditions I might have added.

src/fw/freertos_application.c
```diff
@@ -108,42 +111,60 @@ extern void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime ) {
...
       power_tracking_start(PowerSystemMcuCoreSleep);
       __DSB();  // Drain any pending memory writes before entering sleep.
       do_wfi();  // Wait for Interrupt (enter sleep mode). Work around F2/F4 errata.
       __ISB();  // Let the pipeline catch up (force the WFI to activate before moving on).
       power_tracking_stop(PowerSystemMcuCoreSleep);
 
+#if defined(MICRO_FAMILY_NRF5)
+      uint32_t rtc_end = NRF_RTC0->COUNTER;
+      
+      if (rtc_end < rtc_start) /* NimBLE uses RTC0 (24 bits), 32 kiHz */
```
gmarull, line 155: question: why are we using RTC0 here?

src/fw/drivers/nrf5/hfxo.c
```diff
@@ -21,9 +21,11 @@
 #include <FreeRTOS.h>
 #include <hal/nrf_clock.h>
 
+#include <kernel/util/stop.h>
```
gmarull: not needed

gmarull:
Oh, remember to tweak

https://github.com/coredevices/PebbleOS/blob/d65fe49926dc8c09e1d56bcc9a8df3fd169ba445/src/fw/board/boards/board_asterix.h#L51-L54

if needed

src/fw/drivers/nrf5/hfxo.c
jwise: Good point.  Done.

src/fw/freertos_application.c
jwise: I beefed up the comment a few lines above to answer this: https://github.com/coredevices/PebbleOS/pull/182/commits/b52c0316600e47374f406af6277d1ab598352e57#diff-c59803c4b996804dcbd865a80b231ed3b2d9acd0cda164c3440cbddb8151b701R114-R135

jwise:
> Oh, remember to tweak
> 
> https://github.com/coredevices/PebbleOS/blob/d65fe49926dc8c09e1d56bcc9a8df3fd169ba445/src/fw/board/boards/board_asterix.h#L51-L54
> 
> if needed

This is probably within the margin of error still, plus or minus accelerometer changes.  But once we have QSPI power savings, we will definitely want to fix that.

src/fw/freertos_application.c
gmarull: ok, missed that part, got it. We probably need to rethink how RTC tick rate is configured and align with, e.g. Zephyr, see https://github.com/nrfconnect/sdk-zephyr/blob/6997770e93b3fc57c5325b430fb806f13d345f40/dts/arm/nordic/nrf52840.dtsi#L312 and https://github.com/nrfconnect/sdk-zephyr/blob/6997770e93b3fc57c5325b430fb806f13d345f40/soc/nordic/nrf52/Kconfig.defconfig#L15, but PebbleOS may explode if we change the 1024 Hz premise.

gmarull (approved):
lgtm, not ideal, but I guess the best we can do given the constraints.

PR #184, opened 2025-08-14
Enable LSM6DSO accelerometer on Obelix

gmarull:
Just tested MFG...


jplexer (approved):
lgtm! as mentioned I cannot hardware test it, but I'm gonna trust you on this one

PR #193, opened 2025-08-18
Implement MMC5603NJ driver

### LINKED ISSUE #34: Add driver for MMC5603NJ magnetometer

### Proposed feature request

Add driver for MMC5603NJ magnetometer

### Alternatives

_No response_

### Anything else?

https://lcsc.com/datasheet/lcsc_datasheet_2410121608_MEMSIC-MMC5603NJ_C404328.pdf

matthewwardrop:
This is a rebased version of #34 with a few small patches that is working for me, @gmarull . If I were writing it from scratch I'd probably do things a little differently (mainly around naming of private methods, function ordering, and formatting), but things look sensible here otherwise. It is also missing license headers.

From here I can just turn it over to you, or I could go through and clean things up. Also not sure if @Bob-Wei1 is keen to continue driving this. I'll take my lead from you.

Fixes #34

### DIFF
--- a/src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
+++ b/src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
@@ -16,26 +16,341 @@
  *
  */
 
+#include "drivers/i2c.h"
 #include "drivers/mag.h"
+#include "kernel/events.h"
+#include "kernel/util/sleep.h"
+#include "os/mutex.h"
+#include "system/logging.h"
+#include "system/passert.h"
+#include "services/common/new_timer/new_timer.h"
+#include "util/math.h"
 
 #include "mmc5603nj.h"
+#include "registers.h"
+
+// Forward declarations of private methods
+static bool prv_mmc5603nj_read(uint8_t reg_addr, uint8_t data_len, uint8_t *data);
+static bool prv_mmc5603nj_write(uint8_t reg_addr, uint8_t data);
+static bool prv_mmc5603nj_init(void);
+static bool prv_mmc5603nj_check_whoami(void);
+static bool prv_mmc5603nj_reset(void);
+static bool prv_mmc5603nj_set_sample_rate_hz(uint8_t rate_hz);
+static bool prv_configure_polling(void);
+static void prv_mmc5603nj_polling_callback(void *data);
+static bool prv_mmc5603nj_is_data_ready(void);
+static MagReadStatus prv_mmc5603nj_get_sample(MagData *sample);
+typedef enum {
+  X_AXIS = 0,
+  Y_AXIS = 1,
+  Z_AXIS = 2,
+} axis_t;
+static int16_t prv_get_axis_projection(axis_t axis, int16_t *raw_vector);
+
+// Runtime state
+static bool s_initialized = false;
+static int s_use_refcount = 0;
+static PebbleMutex *s_mag_mutex;
+static uint8_t s_sample_rate_hz = 0;
+static TimerID s_polling_timer = TIMER_INVALID_ID;
+static uint16_t s_polling_interval_ms = 0;
+static bool s_measurement_ready = false;
 
 // MMC5603NJ entrypoints
 
-void mmc5603nj_init(void) {}
+void mmc5603nj_init(void) {
+  s_mag_mutex = mutex_create();
+  if (prv_mmc5603nj_init()) {
+    PBL_LOG(LOG_LEVEL_DEBUG, "MMC5603NJ: Initialization complete");
+  } else {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Initialization failed");
+  }
+}
 
 // mag.h implementation
 
-void mag_use(void) {}
-
-void mag_start_sampling(void) {}
-
-void mag_release(void) {}
-
-MagReadStatus mag_read_data(MagData *data) {
-    return MagReadCommunicationFail;
+void mag_use(void) {
+  PBL_ASSERTN(s_initialized);
+  mutex_lock(s_mag_mutex);
+  ++s_use_refcount;
+  mutex_unlock(s_mag_mutex);
 }
-  
+
+void mag_start_sampling(void) {
+  mag_use();
+  mag_change_sample_rate(MagSampleRate5Hz);
+}
+
+void mag_release(void) {
+  PBL_ASSERTN(s_initialized && s_use_refcount != 0);
+  mutex_lock(s_mag_mutex);
+  --s_use_refcount;
+  if (s_use_refcount == 0) {
+    if (!prv_mmc5603nj_set_sample_rate_hz(0)) {
+      PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Failed to disable sensor on release");
+    }
+  }
+  mutex_unlock(s_mag_mutex);
+}
+
+// callers responsibility to know if there is valid data to be read
+MagReadStatus mag_read_data(MagData *data) {
+  mutex_lock(s_mag_mutex);
+  MagReadStatus rv = prv_mmc5603nj_get_sample(data);
+  mutex_unlock(s_mag_mutex);
+  return rv;
+}
+
 bool mag_change_sample_rate(MagSampleRate rate) {
+  mutex_lock(s_mag_mutex);
+
+  if (s_use_refcount == 0) {
+    mutex_unlock(s_mag_mutex);
+    return true;
+  }
+
+  uint8_t rate_hz = 0;
+  switch (rate) {
+    case MagSampleRate20Hz:
+      rate_hz = 20;
+      break;
+    case MagSampleRate5Hz:
+      rate_hz = 5;
+      break;
+    default:
+      mutex_unlock(s_mag_mutex);
+      return false;
+  }
+
+  bool rv = prv_mmc5603nj_set_sample_rate_hz(rate_hz);
+  mutex_unlock(s_mag_mutex);
+  return rv;
+}
+
+// I2C read/write helpers
+
+static bool prv_mmc5603nj_read(uint8_t reg_addr, uint8_t data_len, uint8_t *data) {
+  i2c_use(I2C_MMC5603NJ);
+  bool rv = i2c_write_block(I2C_MMC5603NJ, 1, &reg_addr);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: I2C write failed for register 0x%02x", reg_addr);
+  }
+  rv = i2c_read_block(I2C_MMC5603NJ, data_len, data);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: I2C data read failed for register 0x%02x", reg_addr);
+  }
+  i2c_release(I2C_MMC5603NJ);
+  return rv;
+}
+
+static bool prv_mmc5603nj_write(uint8_t reg_addr, uint8_t data) {
+  i2c_use(I2C_MMC5603NJ);
+  uint8_t d[2] = {reg_addr, data};
+  bool rv = i2c_write_block(I2C_MMC5603NJ, 2, d);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: I2C write failed for register 0x%02x", reg_addr);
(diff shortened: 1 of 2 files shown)


matthewwardrop:
Roger that, and thanks all! I'll prepare a cleaned up PR based on this work in the next couple of days

jplexer:
@matthewwardrop not that I know of, sorry. I also regularly do the "death grip" ;)

matthewwardrop:
@jplexer I tried to do some debugging on this again tonight, but there's still something borked about my kit. The LSM6DSO unit no longer seems to connect (the WHOAMI check fails). I have tried rebasing this on the latest main, to no avail; and so backed away from that to remain based on core35, which is the last release that actually worked on my watch. Using recovery to reset back to unmodified core35 results in the watch still failing to run the accelerometer, so... either there is some memorised state that is somehow not being reset, or my unit is borked.

I'll keep trying, but I may be unable to keep pushing this forward (since the apps I used to test compass also use the accelerometer). According to the spec sheet, this implementation should be working... but I just cannot test it.

If I manage to make things work again, I'll pick this up again if it is not already done... but I'm not sure I will be able to. Something seems fully broken after I tried the Core36+ builds.

matthewwardrop:
@jplexer Okay... sorry for the spam. I finally managed to get things working again by flashing the pebble hwv tool, interacting with the IMU there, and then flashing the pebble firmware again. Of course, I also lost most of my dev time tonight... tomorrow night I'll try again. (The compass is definitely not performing as it should yet).

matthewwardrop:
@jplexer With my device working, couldn't help but debug things. This is now fully working on my device.

src/fw/board/boards/board_asterix.h
```diff
@@ -92,6 +92,19 @@ static const BoardConfigAccel BOARD_CONFIG_ACCEL = {
...
+  .mag_config = {
+    .axes_offsets[AXIS_X] = 1,
+    .axes_offsets[AXIS_Y] = 0,
+    .axes_offsets[AXIS_Z] = 2,
+    .axes_inverts[AXIS_X] = false,
+    .axes_inverts[AXIS_Y] = true,
+    .axes_inverts[AXIS_Z] = false,
+  },
+  .mag_int_gpio = { NRF5_GPIO_RESOURCE_EXISTS, NRF_GPIO_PIN_MAP(0, 3) },
+  .mag_int = { .peripheral = NRFX_GPIOTE_INSTANCE(0), .channel = 6, .gpio_pin = NRF_GPIO_PIN_MAP(0, 3), },
```
gmarull: magnetometer does not have any int pin, also all gpiote channels are used now

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,367 @@
...
+  return rv;
+}
+
+// I2C read/write helpers
+
+static bool prv_mmc5603nj_read(uint8_t reg_addr, uint8_t data_len, uint8_t *data) {
+  i2c_use(I2C_MMC5603NJ);
+  bool rv = i2c_write_block(I2C_MMC5603NJ, 1, &reg_addr);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_DEBUG, "MMC5603NJ: I2C write failed for register 0x%02x", reg_addr);
```
gmarull: use ERROR level for errors

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,367 @@
...
+  s_sample_rate_hz = rate_hz;
+  prv_configure_polling();
+  return true;
+}
+
+// Configure interrupts and/or polling
+
+static void prv_configure_polling(void) {
+  // We are communicating with the MMC5603NJ chip via the I2C protocol, but
+  // interrupts are only provided for via the I3C protocol. We can simulate
```
gmarull: this sensor has no I3C

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,367 @@
...
+  // interrupts by scheduling timers that poll the device state at the rate
+  // expected by the device ODR.
+
+  uint16_t polling_interval_ms = s_sample_rate_hz == 0 ? 0 : CEILING(1000.0 / s_sample_rate_hz);
```
gmarull: `DIVIDE_CEIL`?

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,367 @@
...
+                                               // keeping measurement error low
+        .prd_set_en = true,
+        .cmm_en = true,
+    };
+    if (!prv_mmc5603nj_write_ctrl2(&ctrl2)) {
+      PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Failed to write INTERNAL_CONTROL_2");
+      mutex_unlock(s_mag_mutex);
+      return false;
+    }
+  }
```
gmarull: I find these intermediate structs/helpers adding too much pseudo-abstraction, why not just perform simple register writes oring fields?

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
matthewwardrop: Oh... I was parroting the spec sheet here. I can remove that comment, though.

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
matthewwardrop: I just didn't want to have to do the casts every time, though there are fewer than I originally thought there would be. 

Is your proposal to get rid of all of the bit fields in registers.h? I personally found them to be much clearer than magic numbers... And allows logic to change in the future. Or are you proposing that I get rid of the intermediate write methods and just inline any type casts?

gmarull: something like

```
reg_write(REG_NAME, FIELD1 | FIELD2);
```

matthewwardrop: Ah. Or-ing not o-ring. That makes more sense.

The main benefit I see is that it is easier with the structs to always cluster the right bits together into the control register, which otherwise needs to be prefixed.

If it is preferable, I am happy to make it look like:
```
prv_mmc5603nj_write(MMC5603NJ_REG_CTRL2,  MMC5603NJ_CTRL2_AUTOSET_PRD_25 | MMC5603NJ_CTRL2_PRD_SET_EN | MMC5603NJ_CTRL2_CMD_EN)
```
with lots of hex codes instead of bit fields in the registers.h file. I think I mildly prefer the structs just for the additional clarity... You *cannot* use the wrong constants anywhere, which gets rid of a bunch of potential bugs... But I agree that this is probably close to (if not already) over-engineering it.

I could also just remove the helper functions for the ctrl registers and do the casts from the structs to uints inline.

I'll do whatever you feel more comfortable maintaining.

gmarull: I prefer simple register writes. While structs can potentially offer type safety, they increase code complexity, and IMHO that gain is not worth it, at least in this case.

matthewwardrop:
@gmarull I have updated the driver to use direct register writes. I did leave in the status bitfield because it seemed like an easier way to consume the response than masks everywhere; but I remain disposed to align this code with your preferences over mine, if you happen to disagree.

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,334 @@
...
+                             MMC5603NJ_CTRL0_AUTO_SR_EN | MMC5603NJ_CTRL0_CMM_FREQ_EN)) {
+      PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Failed to trigger measurement calculation update.");
+    }
+
+    // Start continuous mode
+    if (!prv_mmc5603nj_write(MMC5603NJ_REG_CTRL2, MMC5603NJ_CTRL2_AUTOSET_PRD_100 |
+                                                      MMC5603NJ_CTRL2_PRD_SET_EN |
+                                                      MMC5603NJ_CTRL2_CMM_EN)) {
+      PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Failed to start continuous mode.");
+      mutex_unlock(s_mag_mutex);
```
gmarull: not locking in this function

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,334 @@
...
+  mag_change_sample_rate(MagSampleRate5Hz);
+}
 
-void mag_release(void) {}
+void mag_release(void) {
+  PBL_ASSERTN(s_initialized && s_use_refcount != 0);
+  mutex_lock(s_mag_mutex);
+  --s_use_refcount;
+  if (s_use_refcount == 0) {
+    prv_mmc5603nj_set_sample_rate_hz(0);
```
gmarull: ignoring error here

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,334 @@
...
+static bool prv_mmc5603nj_read(uint8_t reg_addr, uint8_t data_len, uint8_t *data) {
+  i2c_use(I2C_MMC5603NJ);
+  bool rv = i2c_write_block(I2C_MMC5603NJ, 1, &reg_addr);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: I2C write failed for register 0x%02x", reg_addr);
+  }
+  rv = i2c_read_block(I2C_MMC5603NJ, data_len, data);
+  if (!rv) {
+    PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: I2C data read failed for register 0x%02x", reg_addr);
+  }
```
gmarull, line 141: maybe `i2c_read_register_block(I2C_MMC5603NJ, reg_addr, data_len, data);`

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
```diff
@@ -16,26 +16,334 @@
...
+                                                      MMC5603NJ_CTRL2_CMM_EN)) {
+      PBL_LOG(LOG_LEVEL_ERROR, "MMC5603NJ: Failed to start continuous mode.");
+      mutex_unlock(s_mag_mutex);
+      return false;
+    }
+  }
+
+  s_sample_rate_hz = rate_hz;
+  prv_configure_polling();
+  return true;
```
gmarull: polling appears to be fallible, even though it returns nothing, best we'll get is a log, but we should return false.

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
matthewwardrop: Whoops. Thanks for rooting up this vestigial unlock!

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
matthewwardrop: I was relying on the error messages returned already by `prv_mmc5603nj_set_sample_rate_hz`; but I can add another error log here.

src/fw/drivers/imu/mmc5603nj/mmc5603nj.c
matthewwardrop: Good call. I inherited this from the original PR, and didn't scrutinise it much.

matthewwardrop: Actually... this didn't work. I think I also tried this when I was implementing the LSM6DSO driver... and it didn't work then either. I also note that this pattern is used a lot by other drivers... so perhaps this has come up before. Will leave as is for now.

gmarull (approved):

matthewwardrop:
@gmarull As I mentioned above, Pebble seems to declare things calibrated *very* quickly, and yes... at that stage the accuracy is pretty bad. I'm not sure why it does that.

If you keep rotating the device around for another 3-4 times, it seemed fairly accurate and responsive.

PR #196, opened 2025-08-19
[FIRM-495] fw/voice: Fix voice assertion fails and races during startup and cancellation

jplexer:
I swear this microphone driver is gonna plague me till the end of time..
Fixes critical assertion failures in the voice service that occurred during startup and cancellation scenarios.

- Fixed PDM startup race that caused assertions when voice service started
- Implemented delayed Speex cleanup strategy to prevent heap corruption during resource cleanup
- Enhanced cancellation handling with state-specific logic for early vs recording vs post-recording cancellation
- Added session start protection to prevent race conditions between cleanup and new session start
- Improved error handling throughout the voice service state machine

### DIFF
--- a/src/fw/drivers/mic/nrf5/pdm.c
+++ b/src/fw/drivers/mic/nrf5/pdm.c
@@ -55,10 +55,11 @@ static bool prv_allocate_buffers(MicDeviceState *state) {
     PBL_LOG(LOG_LEVEL_ERROR, "Failed to allocate circular buffer storage");
     return false;
   }
-  
+
   // Allocate PDM buffers
   for (int i = 0; i < PDM_BUFFER_COUNT; i++) {
-    state->pdm_buffers[i] = kernel_malloc(PDM_BUFFER_SIZE_SAMPLES * sizeof(int16_t));
+    size_t buffer_size = PDM_BUFFER_SIZE_SAMPLES * sizeof(int16_t);
+    state->pdm_buffers[i] = kernel_malloc(buffer_size);
     if (!state->pdm_buffers[i]) {
       PBL_LOG(LOG_LEVEL_ERROR, "Failed to allocate PDM buffer %d", i);
       // Free any previously allocated buffers
@@ -66,7 +67,7 @@ static bool prv_allocate_buffers(MicDeviceState *state) {
       return false;
     }
     // Clear the buffer
-    memset(state->pdm_buffers[i], 0, PDM_BUFFER_SIZE_SAMPLES * sizeof(int16_t));
+    memset(state->pdm_buffers[i], 0, buffer_size);
   }
   
   // Initialize circular buffer with allocated storage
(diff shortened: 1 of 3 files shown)


src/fw/drivers/mic/nrf5/pdm.c
```diff
@@ -364,6 +386,12 @@ void mic_stop(const MicDevice *this) {
   // Stop PDM capture
   nrfx_pdm_stop(&this->pdm_instance);
   
+  // Give the PDM hardware a moment to finish any pending operations
+  // This helps ensure no DMA operations are still accessing our buffers
+  for (volatile int i = 0; i < 1000; i++) { 
+    // Small delay to let hardware settle
+  }
```
gmarull: feels hacky, is this really needed or we're missing something? If needed, why isn't a psleep sufficient, or a busy wait?

src/fw/services/normal/voice/voice.c
```diff
@@ -482,12 +518,27 @@ void voice_cancel_dictation(VoiceSessionId session_id) {
...
       prv_stop_recording();
+      // Use delayed cleanup to avoid race condition with microphone
+      s_delayed_speex_cleanup = true;
+      new_timer_start(s_timeout, 100, prv_delayed_cleanup_timeout, NULL, 0);
+      prv_reset();
+    } else if (s_state == SessionState_WaitForSessionResult) {
+      // Recording is done but audio endpoint might still be processing
+      s_delayed_speex_cleanup = true;
+      new_timer_start(s_timeout, 100, prv_delayed_cleanup_timeout, NULL, 0);
+      prv_reset();
```
gmarull, line 536: How are these delays calculated to "avoid races"? In general, this all feels very fragile. Maybe pebbleos was already fragile?

src/fw/drivers/mic/nrf5/pdm.c
jplexer: yea you're right. psleep(1) is the proper solution here. the volatile busy wait was indeed just hacky. I honestly blanked on psleep existing for a moment there 🤦‍♂️. 1ms is more than sufficient for any pending DMA operations to complete during shutdown, and psleep is much cleaner than burning CPU cycles with a busy wait.

src/fw/services/normal/voice/voice.c
jplexer: Yep, those 100ms delays are pure "throw some time at it and hope for the best" engineering 😅. They're total band-aids over race conditions. The real issue is we have zero coordination between microphone/audio endpoint/codec teardown - just crossing fingers that 100ms is enough for everything to settle. A bit janky, but that does seems par for the course with some of the cleanup logic in pebbleos.

gmarull: Was STM32 code racy as well? It would not surprise me, but just to confirm, we're not taking too many shortcuts.

jplexer: oh definitely

gmarull:
needs rebase

gmarull (approved):

PR #197, opened 2025-08-19
fw/drivers/lsm6dso: Add error detection and recovery

jplexer:
Users have reported that the accelerometer randomly stops working, with symptoms including:
- Shake to backlight stops functioning
- Step counting ceases
- Motion-based watch face features break
- Issue persists until device reboot
- Affects multiple users intermittently

The LSM6DSO driver lacked robust error handling for I2C communication failures:
1. **No I2C error detection**: Failed reads/writes were ignored, leaving the driver unaware of communication issues
2. **Missing health monitoring**: No mechanism to detect when the sensor becomes unresponsive
3. **No recovery logic**: Once communication failed, the driver had no way to recover without full system restart
4. **Silent failures**: Errors weren't tracked or logged adequately for debugging

This PR implements comprehensive error detection and recovery mechanisms:

### Error Tracking & Detection
- Track I2C communication failures with detailed logging
- Count consecutive read failures to detect persistent issues
- Monitor time since last successful sensor communication
- Implement periodic health checks via WHO_AM_I register validation

### Automatic Recovery
- Attempt sensor recovery when consecutive failures exceed threshold
- Perform full sensor reset and reconfiguration during recovery
- Restore previous operational state after successful recovery
- Clear error counters after successful recovery

### DIFF
--- a/src/fw/drivers/imu/lsm6dso/lsm6dso.c
+++ b/src/fw/drivers/imu/lsm6dso/lsm6dso.c
@@ -44,6 +44,8 @@ static void prv_lsm6dso_configure_shake(bool enable, bool sensitivity_high);
 static void prv_lsm6dso_interrupt_handler(bool *should_context_switch);
 static void prv_lsm6dso_process_interrupts(void);
 static bool prv_is_vibing(void);
+static bool prv_lsm6dso_health_check(void);
+static void prv_lsm6dso_attempt_recovery(void);
 typedef struct {
   lsm6dso_odr_xl_t odr;
   uint32_t interval_us;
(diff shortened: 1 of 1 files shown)


jplexer:
You're absolutely right that sensor unresponsiveness shouldn't happen under normal circumstances, and I agree that needing workarounds like this indicates we're doing something wrong in our driver implementation. However the current driver provides no visibility into the failures we've been seeing - it just silently fails, making root cause analysis impossible. This patch is primarily a diagnostic tool to help us understand the failure patterns (power issues, I2C timing problems, state corruption, etc.) with the recovery mechanism as a secondary benefit to keep devices functional while we investigate. 
I'd much prefer to find and fix whatever we're doing wrong that's causing the sensor to become unresponsive, then remove this workaround entirely. Would you prefer I focus on root cause analysis first, or can we use this to gather diagnostic data while investigating?

gmarull (approved):

PR #198, opened 2025-08-19
Toggle EXTCOMIN using RTC+PPI+GPIOTE

gmarull:
I still need to do some extra verifications, and add some "safety" to the dirty RTC2 sharing.

image


jplexer (approved):
Looks good to me, although I won't merge as I would like jwise to take a look as well :)

jwise:
I took an alternative approach to repurposing the WDT in question: https://github.com/acceleratedtech/PebbleOS/commit/e18385642188557621335d3108b84d9de07dc05b

src/fw/board/boards/board_asterix.h
```diff
@@ -116,9 +115,14 @@ static const BoardConfigSharpDisplay BOARD_CONFIG_DISPLAY = {
   .on_ctrl = { NRF5_GPIO_RESOURCE_EXISTS, NRF_GPIO_PIN_MAP(0, 4), true },
 
   .extcomin = {
-    .state = &DISPLAY_EXTCOMIN_STATE,
-    .output = { NRF5_GPIO_RESOURCE_EXISTS, NRF_GPIO_PIN_MAP(1, 15), true },
-    .peripheral = NRFX_PWM_INSTANCE(1),
+    .rtc = NRF_RTC2,
+    .rtc_hz = 32768,
```
jwise: 32 kHz is a little aggressive... does this need to be > 512Hz or so?

jwise:
I'm ok with this, though it doesn't seem to actually hit our floor current yet with this plus QSPI changes.  More debugging to do tomorrow.  I like my approach better for sharing WDT RTCs though :)

gmarull:
> I'm ok with this, though it doesn't seem to actually hit our floor current yet with this plus QSPI changes. More debugging to do tomorrow. I like my approach better for sharing WDT RTCs though :)

So the problem of sharing RTC is that we would have two competing clear events. We can't rely on the TICK event here, as the duty cycle is not 50%, so we need to clear on every period. Maybe there's a smarter way; I just copied Zephyr's ideas. Also, I don't really like sharing peripheral resources without a clear barrier in between that prevents us from shooting ourselves in the foot...

src/fw/board/boards/board_asterix.h
gmarull: Actually, this should ideally be internal, so then the period/pulse values specified by the user have maximum resolution. Not critical here, though, as this is not a generic PWM driver even though it should be so that any future boards can do whatever they want.

PR #204, opened 2025-08-20
fw/services/activity: Validate RTC initialization and load step data if invalid [FIRM-497]

jplexer:
we may not be able to preserve rtc, but we can preserve steps from storage if we have reset!

### DIFF
--- a/src/fw/services/normal/activity/activity.c
+++ b/src/fw/services/normal/activity/activity.c
@@ -46,6 +46,7 @@
 #include "util/math.h"
 #include "util/size.h"
 #include "util/units.h"
+#include "util/time/time.h"
 
 #include <pebbleos/cron.h>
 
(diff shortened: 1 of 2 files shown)


jplexer:
noticed a bug where my steps were first persisted correctly but after that they were moved to yesterday?? gotta take a look at that

src/fw/services/normal/activity/activity.c
```diff
@@ -415,7 +420,17 @@ static void NOINLINE prv_process_minute_data_tail(time_t utc_sec) {
...
+    if (cur_day_index != s_activity_state.cur_day_index && !rtc_was_invalid_during_init) {
+      time_t rtc_time = rtc_get_time();
+
+      // Check if the RTC timestamp is valid (greater than 1.1.2000)
+      const time_t valid_epoch = 946684800; // 1.1.2000 in seconds since epoch
+      if (rtc_time < valid_epoch) {
+          PBL_LOG(LOG_LEVEL_WARNING, "RTC not initialized. Skipping data reset.");
+          return;
+      }
+
```
jwise, line 433: DRY: this and below (with the magic constants) maybe should be refactored into `rtc_is_initialized` or something?

src/fw/services/normal/activity/activity_metrics.c
```diff
@@ -677,6 +696,7 @@ void activity_metrics_prv_init(SettingsFile *file, time_t utc_now) {
           new_history.values[new_index] = old_history.values[i];
         }
       }
+      }
```
jwise, line 699: either update indentation here -- or invert the case of it, and just `return` if the RTC is not valid?

jwise:
Hmm, this adds a bunch of weird states that I'm not really sure I can understand, trying to recover step data into different days, hours, times, ... when the time is wrong and then suddenly becomes right again.

I am sort of thinking of a different approach -- if the RTC is wrong, maybe we should refuse to execute any `activity_` APIs at all -- even `activity_init()`?  We would not even try to log steps if the RTC is wrong, under the idea that the phone would soon come and tell us what time it is.  Then on the first time someone calls an `activity_` API when the RTC is correct, we would *then* do our `activity_init()` and continue on.

PR #206, opened 2025-08-21
drivers/nrf5/qspi: shut down QSPI controller when it is not in use [FIRM-430]

jwise:
In the STM32 world, it is sufficient to just put the flash in DPD around stop mode.  But on nRF52, it's actually not the flash that uses that much power -- it's the flash *controller*!  Having the flash controller turned on could cost as much as 800uA compared to the flash itself, which probably is around 100 uA when not in DPD.

We really want to shut it off ASAP, even if we're not in stop mode -- and we definitely don't want to power it back up when we leave stop mode, because powering it up is pretty slow (it requires turning on the HFXO in order to be reliable).  So instead of using the PebbleOS "please enter low power mode" hints, we set a new_timer of our own to power up on first access, and then power down after 50ms of nobody talking to us.

### DIFF
--- a/src/fw/drivers/nrf5/qspi.c
+++ b/src/fw/drivers/nrf5/qspi.c
@@ -16,6 +16,9 @@
 #include "system/logging.h"
 #include "system/passert.h"
 #include "util/math.h"
+#include "services/common/new_timer/new_timer.h"
+#include "os/mutex.h"
+#include "drivers/clocksource.h"
 
 #define NRF5_COMPATIBLE
 #include <mcu.h>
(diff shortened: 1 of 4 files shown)


src/fw/drivers/nrf5/qspi.c
```diff
@@ -43,6 +46,117 @@ status_t flash_impl_get_nvram_erase_status(bool *is_subsector, FlashAddress *add
...
+    // the new_timer will come and try again later
+    return;
+  }
+
+#ifdef DEBUG_FLASH_SLEEP_TIME
+  // Must be LOG_LEVEL_DEBUG ... otherwise the flash logging subsystem will
+  // wake up and write to flash!
+
+  PBL_LOG(LOG_LEVEL_DEBUG, "flash is going back to sleep");
+#endif
```
gmarull: flash log level is actually configurable to any level, we cannot log here

gmarull (commented):
question: would `nrfx_qspi_act/deactivate` be enough instead of a full init/uninit?

gmarull:
should we remove `flash_power_down_for_stop_mode` from `stop.c`?

src/fw/drivers/nrf5/qspi.c
jwise: I will update the comment -- it's `#ifdef`'ed out by default anyway

gmarull:
> > question: would `nrfx_qspi_act/deactivate` be enough instead of a full init/uninit?
> 
> Maybe. I'll try it. In my previous testing, it was not enough, but looking at `nrfx_qspi.c`, I can't find any reason why that would be true now.
> 
> > should we remove `flash_power_down_for_stop_mode` from `stop.c`?
> 
> No. Legacy devices still require it. I was considering wrapping it in a `CAPABILITY_QSPI_MANAGES_OWN_POWER` or some such, but it's ok for it to just be a no-op for now.

I refer to:
```
 58 #if MICRO_FAMILY_NRF5                                                                                                                                                 
 59 void enter_stop_mode(void) {                                                                                                                                          
 60   dbgserial_enable_rx_exti();                                                                                                                                         
 61                                                                                                                                                                       
 62   flash_power_down_for_stop_mode(); 
```

which is nRF5 specific

src/fw/drivers/nrf5/qspi.c
gmarull: you could use:
```
#if defined(DEBUG_FLASH_SLEEP_TIME) && FLASH_LOG_LEVEL < LOG_LEVEL_DEBUG
```

jwise: I added a check to avoid surprising behavior, how about that?

jwise:
> I refer to:
> 
> ```
>  58 #if MICRO_FAMILY_NRF5                                                                                                                                                 
>  59 void enter_stop_mode(void) {                                                                                                                                          
>  60   dbgserial_enable_rx_exti();                                                                                                                                         
>  61                                                                                                                                                                       
>  62   flash_power_down_for_stop_mode(); 
> ```
> 
> which is nRF5 specific

Yes, good point.  I excised this here, and also more aggressively elsewhere in the codebase.

PR #207, opened 2025-08-21
Improve MCU API, add impl for SF32LB52

gmarull:


src/fw/kernel/system_versions.c
```diff
@@ -219,10 +219,17 @@ void command_version_info(void) {
   char hw_version[MFG_HW_VERSION_SIZE + 1];
   mfg_info_get_hw_version(hw_version, sizeof(hw_version));
 
-  const uint32_t* mcu_serial = mcu_get_serial();
-  prompt_send_response_fmt(buffer, sizeof(buffer),
-                           "MCU Serial: %08"PRIx32" %08"PRIx32" %08"PRIx32,
-                           mcu_serial[0], mcu_serial[1], mcu_serial[2]);
+  uint8_t mcu_serial[32];
```
cameled, line 222: Please initialize the mcu_serial, like `uint8_t mcu_serial[32] = {0};`

gmarull: hmm, it’s always accessed if return code is 0, so it will contain valid data up to the returned size, so why would this be needed?

cameled: You are right, not needed in here.

cameled (approved):

PR #209, opened 2025-08-21
boards/board_asterix: power down PWM1 on boot [FIRM-309]

jwise:
We use GPIOTE and PPI now to toggle Vcom on asterix, which is much cheaper than PWM.  Unfortunately, the bootloader left PWM turned on, so it is sitting there demanding the PCLK, costing us power.  Turn it off on boot.

### DIFF
--- a/src/fw/board/boards/board_asterix.c
+++ b/src/fw/board/boards/board_asterix.c
@@ -242,4 +242,6 @@ void board_init(void) {
   i2c_use(I2C_DA7212);
   i2c_write_block(I2C_DA7212, 2, da7212_powerdown);
   i2c_release(I2C_DA7212);
+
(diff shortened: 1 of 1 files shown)


src/fw/board/boards/board_asterix.c
```diff
@@ -242,4 +242,6 @@ void board_init(void) {
   i2c_use(I2C_DA7212);
   i2c_write_block(I2C_DA7212, 2, da7212_powerdown);
   i2c_release(I2C_DA7212);
+
+  NRF_PWM1->ENABLE = 0; // Bootloader left this powered on.  We're done with it now, though; we use GPIOTE+PPI to toggle Vcom.
```
gmarull, line 246: we have: https://github.com/coredevices/PebbleOS/blob/main/platform/asterix/boot/src/drivers/display/sharp_ls013b7dh01.c#L329

gmarull: In fact, bootloader uses PWM0

PR #210, opened 2025-08-21
fw/drivers/lsm6dso: Apply comprehensive fixes for unresponsive accelerometer

jplexer:
The LSM6DSO accelerometer driver had a tendency to become unresponsive after I2C communication failures, leaving users without shake detection, tap detection, or accelerometer data. The original implementation tracked consecutive read failures but didn't have a comprehensive recovery strategy -- when the sensor got stuck, it would stay stuck until a reboot.

We solve this by implementing a multi-layered approach to sensor health monitoring and recovery. We add a watchdog timer that monitors for unresponsive sensors (no successful reads for 5+ seconds) and consecutive error thresholds (5+ consecutive failures). When triggered, the recovery mechanism performs a full sensor reset, restores basic configuration, and validates functionality with a health check.

Additionally, we improve FIFO robustness by detecting overflow conditions and stuck FIFO states, automatically resetting the FIFO when communication errors occur during sample reading. We also add timeout protection to sensor initialization to prevent hanging during startup, and fix a race condition in target state management that could lose configuration changes.

This takes the accelerometer from "permanently broken after first I2C glitch" to "automatically recovers from transient failures while maintaining user configuration", significantly improving the reliability of motion-based features like shake-to-light and tap gestures.

I've been trying this patch out for the past like 36 Hours and I reached 25k steps while I was at Gamescom :D

### DIFF
--- a/src/fw/drivers/imu/lsm6dso/lsm6dso.c
+++ b/src/fw/drivers/imu/lsm6dso/lsm6dso.c
@@ -45,7 +45,7 @@ static void prv_lsm6dso_interrupt_handler(bool *should_context_switch);
 static void prv_lsm6dso_process_interrupts(void);
 static bool prv_is_vibing(void);
 static bool prv_lsm6dso_health_check(void);
-static void prv_lsm6dso_attempt_recovery(void);
+static bool prv_lsm6dso_attempt_recovery(void);
 typedef struct {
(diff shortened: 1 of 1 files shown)


gmarull (approved):
approving, but we still need to find what we're doing wrong, as the sensor should always be 'okay'

PR #212, opened 2025-08-22
Fix for RTC reset causing step count to reset [FIRM-497]

### LINKED ISSUE #188: Step count is lost after reset

### Current Behavior

Steps seem to be lost after a reboot.

### Expected Behavior

Steps are preserved across reboots, or if there's a crash and system reboots.

### Steps To Reproduce

Go into healt app, "fake" some steps (swing watch a few times), then press reset button on the B2B. After reset step count will not be preserved.

### Version

v4.9.9-core33

### Host OS

N/A

### Watch

Pebble 2 Duo (Asterix)

### Anything else?

_No response_

jplexer:
Fixes #188 / FIRM-497

In order to fix the issue, we only initialize the activity service once the rtc is past 2010-01-01 (since Pebble was released in 2013 and I don't believe in time travel!)

I've added early returns to quite a bit of activity.c in order to allow pebble os to boot. (i have probably added a bit too many but better safe than sorry :P.
I've also added checks to health and workout so that they arent launchable until we have init'd.

The health service from applib provides dummy data until activity is initialized, so that stuff like kickstart (which I have added to asterix in this pr also) works :)

### DIFF
--- a/src/fw/applib/health_service.c
+++ b/src/fw/applib/health_service.c
@@ -920,6 +920,9 @@ bool health_service_private_get_metric_history(HealthMetric metric, uint32_t his
 // ----------------------------------------------------------------------------------------------
 HealthServiceAccessibilityMask health_service_metric_accessible(
   HealthMetric metric, time_t time_start, time_t time_end) {
+  if (!activity_is_initialized()) {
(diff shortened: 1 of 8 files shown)


jplexer:
thats it, I am going back to my old solution, this is too much fuckery for me

PR #217, opened 2025-08-24
Fix for RTC reset causing step count to reset [FIRM-497]

jplexer:
Fixes https://github.com/coredevices/PebbleOS/issues/188 / FIRM-497

In order to fix the issue, we only initialize the activity service once the rtc is past 2010-01-01 (since Pebble was released in 2013 and I don't believe in time travel!)

I've added early returns to quite a bit of activity.c in order to allow pebble os to boot. (i have probably added a bit too many but better safe than sorry :P).
I've also added checks to health and workout so that they arent launchable until we have init'd.

The health service from applib provides dummy data until activity is initialized, so that stuff like kickstart (which I have added to asterix in this pr also) works :)

### DIFF
--- a/src/fw/services/normal/services_normal.c
+++ b/src/fw/services/normal/services_normal.c
@@ -18,6 +18,9 @@
 
 #include <string.h>
 
+#include "applib/event_service_client.h"
+#include "drivers/rtc.h"
+#include "kernel/events.h"
 #include "process_management/app_install_manager.h" // FIXME: This should really be in services/
 #include "process_management/launcher_app_message.h" // FIXME: This should really be in services/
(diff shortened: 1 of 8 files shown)


src/fw/services/normal/services_normal.c
```diff
@@ -51,6 +54,42 @@
 
 #include "util/size.h"
 
+#ifdef PLATFORM_ASTERIX
```
gmarull: Let's force this on all watches. Time can go off anytime the battery dies, even on old watches, so the system must be prepared for that. On Obelix, this also happens on a "hard reset".

jplexer: done :)

PR #220, opened 2025-08-25
Fix crashes

gmarull:
Crashes were causing more crashes, let's fix that.


jplexer:
Awesome! Let's ship it

PR #235, opened 2025-08-28
fw/services/accel_manager: validate subscriber state in get_num_samples

jplexer:
This adds a small, defensive validation to sys_accel_manager_get_num_samples that verifies the supplied AccelManagerState pointer  is still present in the manager’s subscriber list before accessing it. If the state is not found the syscall returns 0.


gmarull (commented):
It feels like this is papering over the real issue: why did we get there?

PR #236, opened 2025-08-28
fw/services/common/accel_manager: fix use-after-free

gmarull:
It looks like the Pebble accel manager came with a race that could occur when start/stopping activity tracking (this happens e.g. when plugging the charger).

The problem could occur if prv_call_data_callback() was called for a state assigned to KernelBackground task (which defers to system task queue), then data from a previous event was consumed (so clearing the event posted flag) and then activity was stopped. Because flag was cleared despite having a new event in the queue, the unsubscribe/delete logic did not defer the de-allocation properly, causing a use-after-free when the actuall callback was processed.

The current code is still incorrect, if >1 event is queued, free will still occur (as sys_accel_manager_data_unsubscribe effectively booleanizes the new events posted counter). I haven't observed in practice this occuring, and since this involves changing a syscall, let's wait a bit more or if there are better solutions.


jplexer:
sorry, I thought this was gtg :pensive:

PR #240, opened 2025-09-01
Re-enable stop mode and QSPI suspend

gmarull:


gmarull:
Note: I see rendering being noticeably slower on my watch after re-applying these patches, but a reboot does not occur.

PR #243, opened 2025-09-03
[DNM] fw/lptim_systick: introduce lptim systick

cameled:
This introduce lptim systick for sf32lb52.

Note: There still have a FreeRTOS assert fail in task.c:1960.


cameled:
@gmarull Which low poer state for sf32lb52 stop mode, lightsleep or deepsleep. I see watchdog docuement said it will not work if entering deepsleep mode.

gmarull:
> @gmarull Which low poer state for sf32lb52 stop mode, lightsleep or deepsleep.
> 
> The WDT1 watch will not work if entering deepsleep mode. The IWDT can work in deepsleep mode.

We should enter one or another mode depending on the next scheduled wake time. We need to retain all ram, so that gives us Sleep/Deepsleep options.

cameled:
I have add a debug pin to record the precision of lptim systick. Some time lost observed at startup, each nearly 20 ms, same as flash block erase time.
 捕获222

After add some log to qspi.c to record the flash erase. There have two erase event at startup. First is the `MFG_INFO` region, at `0x13FFE000 - 0x13FFEFFF`; the second is `FILESYSTEM` region, at `0x12AA0000 - 0x13FBFFFF`. The `MFG_INFO` erase size is fixed to 4K, or 20ms erase time . the `FILESYSTEM` erase is not fixed, sometime small, sometime large.

gmarull:
> I have add a debug pin to record the precision of lptim systick. Some time lost observed at startup, each nearly 20 ms, same as flash block erase time. 捕获222
> 
> After add some log to qspi.c to record the flash erase. There have two erase event at startup. First is the `MFG_INFO` region, at `0x13FFE000 - 0x13FFEFFF`; the second is `FILESYSTEM` region, at `0x12AA0000 - 0x13FBFFFF`. The `MFG_INFO` erase size is fixed to 4K, or 20ms erase time . the `FILESYSTEM` erase is not fixed, sometime small, sometime large.

Does flash affect in any way LPTIM? If so this is bad @rabbitsaviola ?

cameled:
Nop, it's a common behavior, the systick also worked like this.

gmarull:
> Nop, it's a common behavior, the systick also worked like this.

Ok, could you post power consumption deltas?

cameled:
> > Nop, it's a common behavior, the systick also worked like this.
> 
> Ok, could you post power consumption deltas?

I get the PPK2, will post the power consumption latter.

cameled:
@gmarull  ppk power consumption result is added.

PR #259, opened 2025-09-14
feat: add display orientation setting for asterix devices

rojas-claudio:
This (my first!) pull request introduces a new 'Orientation' display setting for asterix devices, allowing the user to more comfortably wear their watch on the right wrist. As opposed to reaching around with the left thumb to press Up/Down/Select, the navigation buttons can be pressed with the left index finger as shown:

https://github.com/user-attachments/assets/30c574ee-9aa2-47cc-96b8-4995c9d576cc

### DIFF
--- a/src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
+++ b/src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
@@ -19,6 +19,7 @@
 #include "util/reverse.h"
 #include "util/units.h"
 
+#include "shell/prefs.h"
 
 #define NRF5_COMPATIBLE
 #include <mcu.h>
@@ -335,7 +336,6 @@ void display_update(NextRowCallback nrcb, UpdateCompleteCallback uccb) {
 void display_pulse_vcom(void) {
 }
 
-#if DISPLAY_ORIENTATION_ROTATED_180
 //!
 //! memcpy the src buffer to dst and reverse the bits
 //! to match the display order
@@ -347,7 +347,6 @@ static void prv_memcpy_reverse_bytes(uint8_t* dst, uint8_t* src, int bytes) {
         *dst++ = reverse_byte(*src++);
     }
 }
-#else
 //!
 //! memcpy the src buffer to dst backwards (i.e. the highest src byte
 //! is the lowest byte in dst.
@@ -358,7 +357,6 @@ static void prv_memcpy_backwards(uint32_t* dst, uint32_t* src, int length) {
     *dst-- = ntohl(*src++);
   }
 }
-#endif
 
 
 static bool prv_do_dma_update(void) {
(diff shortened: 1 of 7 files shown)


src/fw/shell/normal/prefs.c
```diff
@@ -73,6 +73,11 @@ static uint16_t s_backlight_intensity; // default pulled from BOARD_CONFIGs in s
 #define PREF_KEY_BACKLIGHT_MOTION "lightMotion"
 static bool s_backlight_motion_enabled = true;
 
+#if PLATFORM_ASTERIX
```
jplexer, line 76: nit: I think for now making this asterix only is okay, but I think this should be something for all devices :)

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
```diff
@@ -403,15 +401,15 @@ static bool prv_do_dma_update(void) {
       return was_higher_priority_task_woken != pdFALSE;
     }
 
-#if DISPLAY_ORIENTATION_ROTATED_180
```
jplexer: nit: *technically* this would be used by tintin
we don't build for tintin so it *should* be fine

jplexer (approved):
looks good to me overall (except those few nitpicks)
great job claudio!

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
rojas-claudio: Doesn't tintin use `src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01.c`? This is `src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c`

src/fw/shell/normal/prefs.c
rojas-claudio: I'd love to get it working on all devices! Soon 🙏🏼 :)

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
jplexer: Oh yeah right, sorry

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
```diff
@@ -19,6 +19,7 @@
 #include "util/reverse.h"
 #include "util/units.h"
 
+#include "shell/prefs.h"
```
gmarull, line 22: depending on shell at driver level feels wrong

gmarull (changes requested):
Since this is a runtime setting, drivers should not be modified, nor should board config. E.g. back button will always be the back button, and the display is not physically rotated.

edit: driver may have an API to indicate display has to be rotated, e.g. `display_set_rotation(180)`. This should be called somewhere in the init stage depending on configured setting, or when changing the display setting. Otherwise we have a dependency inversion here.

PR #264, opened 2025-09-16
Fix issues with voice [FIRM-593] [FIRM-606]

jplexer:
In order to fix FIRM-606 I had to revert the fix for FIRM-593

included is another fix for FIRM-593 that I asked Eric to test


jplexer:
Gotcha, updated the commit
I want to wait for Eric to confirm if this fixes FIRM-593 for him (since AFAIK) he was the only one that experienced that

gmarull (approved):

PR #276, opened 2025-09-19
FIRM-615: create on-device debugging options to generate a core dump

jwise:
As per FIRM-615, we surface two options to create a core dump: an immediate core dump option (previously hidden behind Settings -> System -> Information -> long-press Firmware), and a "core dump any time" option that can be enabled to trigger a core dump by rapidly clicking the back button 10 times in a row (intended for use with debugging FIRM-425).

We also create a 'debugging' menu in system settings (gated behind a confirmation dialog) to control these, and, while we're at it, if the user voluntarily creates a core dump, then we pop up a special dialog on reboot to explain why there was an unceremonious reboot, and to tell the user to go and upload the crash dump through CoreApp.

### DIFF
--- a/src/fw/shell/prf/stubs.c
+++ b/src/fw/shell/prf/stubs.c
@@ -192,6 +192,11 @@ int16_t shell_prefs_get_automatic_timezone_id(void) {
   return -1;
 }
 
+bool shell_prefs_can_coredump_on_request() {
+  // it would be good to have a core dump escape hatch in PRF
+  return true;
+}
+
 AlertMask alerts_get_mask(void) {
   return AlertMaskAllOff;
 }
--- a/src/fw/shell/prf/stubs.c
+++ b/src/fw/shell/prf/stubs.c
@@ -45,6 +45,9 @@ const char *app_custom_get_title(AppInstallId app_id) {
 void crashed_ui_show_worker_crash(AppInstallId install_id) {
 }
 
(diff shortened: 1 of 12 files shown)


jwise:
blgh, why does it build for me but not in CI

src/fw/shell/prf/stubs.c
```diff
@@ -192,6 +192,11 @@ int16_t shell_prefs_get_automatic_timezone_id(void) {
   return -1;
 }
 
+bool shell_prefs_can_coredump_on_request() {
+  // it would be good to have a core dump escape hatch in PRF
+  return true;
```
gmarull, line 200: `false` on PRF?

gmarull (commented):
lgtm, one question, also remember to move linear tags to commit body with Fixes FIRM-XXX

src/fw/shell/prf/stubs.c
jwise: I thought about that both directions and left it intentionally as `true`.  jamming on the back button in PRF is a very unusual thing to do, so it's unlikely to trigger under normal use.  and it would be nice to have a mechanism to generate a core dump in PRF also, for instance, when debugging weird bluetooth behavior on one system but not on another.

jwise:
Updated commit messages with Linear tags ("Associated with xxx" for all but the final one, then "Fixes xxx" for the final one).

gmarull (approved):

PR #277, opened 2025-09-19
kernel/event_loop: avoid "overkill"ing watchfaces if the KernelMain task has gone out to lunch [FIRM-425]

jwise:
The back-button-to-force-quit is handled on the KernelMain thread.  But if shutting down the Launcher and entering a watchface takes a long time, and the KernelMain thread gets starved, the back-button-to-force-quit sequence could be activated even for just a short press of the back button! Obviously it would be better if the system were not so darned slow, but in the mean time, if it has to be, we can add a second check later to see if the back-button-release event has fired by the time we actually get around to killing the newborn watchface.

I'm on a train right now, and so I've only compiled this, but I haven't actually tested it 🙂 Gerard, if you get a chance to test this tomorrow morning, go for it, and feel free to merge if it doesn't cause obvious regressions (test both a force-kill and some normal back button presses, I suppose).  But otherwise it "looks" correct to me, and if you don't get to it first, I'll check it on hardware tomorrow Joshua-morning.


src/fw/kernel/event_loop.c
```diff
@@ -147,11 +147,26 @@ bool launcher_popups_are_blocked(void) {
...
+  s_force_quit_was_cancelled = true;
   new_timer_stop(s_back_hold_timer);
 }
 
 static void launcher_force_quit_app(void *data) {
+  if (s_force_quit_was_cancelled) {
+    // If you see this in logs after FIRM-556 (general system sluggishness)
+    // is fixed, please file a bug!
+    PBL_LOG(LOG_LEVEL_ERROR, "NewTimer event fired for force quit, but the back button was released before we went to deal with it!  Wow, we must have been really slow!  Please file a bug!");
+    return;
```
gmarull: it would probably be good to track this on memfault, so we have an idea on how often this happens

gmarull (commented):
sounds reasonable. move linear tags to commit body, like "Fixes FIRM-XXX"

src/fw/kernel/event_loop.c
jwise: Seems right.  Glued that in.

PR #282, opened 2025-09-21
fw/display: Improve obelix display performance 

cameled:
This PR is focusing on obelix display performance.

Progress:
1.  the button timer same as pwm led, which cause button event more slow, or gui response more slow.
  -- fix: use the gptim2 for button instead

The `display_update` spend nearly 37 ms. `222_TO_332` convert about 3~4ms.

The display_update call stack as below.
```

 -> HAL_LCDC_SendLayerData2Reg_IT
  -> SendLayerData_IT
   -> _SendLayerData
    -> LayerUpdate
 -> wait HAL_LCDC_SendLayerDataCpltCbk.
```

### DIFF
--- a/src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
+++ b/src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
@@ -32,15 +32,11 @@
 
 #define BYTE_222_TO_332(data) ((((data) & 0x30) << 2) | (((data) & 0x0c) << 1) | ((data) & 0x03))
 
-static uint8_t s_framebuffer[DISPLAY_FRAMEBUFFER_BYTES];
+static uint8_t s_framebuffer[2][DISPLAY_FRAMEBUFFER_BYTES];
+static bool s_framebuffer_dirty[2];
+static uint8_t s_framebuffer_index;
 static PebbleMutex *s_update;
 static SemaphoreHandle_t s_write_done;
-
-static void prv_fb_222_to_332(const DisplayRow *row) {
-  for (uint16_t count = 0U; count < PBL_DISPLAY_WIDTH; count++) {
-    s_framebuffer[row->address * PBL_DISPLAY_WIDTH + count] = BYTE_222_TO_332(*(row->data + count));
-  }
-}
 
 // TODO(SF32LB52): Improve/clarify display on/off code
 static void prv_display_on() {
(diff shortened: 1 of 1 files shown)


src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
```diff
@@ -149,18 +160,32 @@ void display_init(void) {
 
 void display_clear(void) {
   DisplayJDIState *state = DISPLAY->state;
+  uint8_t index = 0xFF;
 
   mutex_lock(s_update);
 
-  memset(s_framebuffer, 0xFF, DISPLAY_FRAMEBUFFER_BYTES);
+  while (index == 0xFF) {
+    psleep(10);
```
gmarull: I'm a bit confused with this part; shouldn't we always have a free buffer available? There can only be 1 transfer ongoing to the display.

gmarull:
note: please squash, some commits seems to contain partial work (e.g. stop mode)

src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
cameled: Both frame buffer may be write by UI layer, But looks the psleep is not required in here. Will same as display update, rewrite the buffer directly.

gmarull: UI layer cannot write to the buffer being transferred to the display, am I missing something?

cameled: You are right. Only the idle(or pending?) buffer may be rewriten.

cameled: done.

src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
```diff
@@ -188,10 +212,20 @@ void display_update(NextRowCallback nrcb, UpdateCompleteCallback uccb) {
   uint16_t rows = 0U;
   uint16_t y0 = 0U;
   bool y0_set = false;
+  uint8_t index = 0xFF;
```
gmarull:
```suggestion
  uint8_t index;
```

src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
```diff
@@ -149,18 +160,31 @@ void display_init(void) {
 
 void display_clear(void) {
   DisplayJDIState *state = DISPLAY->state;
+  uint8_t index = 0xFF;
```
gmarull:
```suggestion
  uint8_t index;
```

gmarull (commented):
lgtm, just nitpicks

src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
cameled: done

src/fw/drivers/sf32lb52/jdi_lpm015m135a.c
cameled: done

gmarull (approved):

PR #295, opened 2025-09-25
drivers/imu/lsm6dso: enhance interrupt watchdog with recovery attempts and forced reinitialization

jplexer:
This pull request improves the robustness of the LSM6DSO IMU driver by enhancing its interrupt handling and recovery logic. It adds a multi-stage recovery mechanism for sensor recovery when interrupts are missed, escalating from simple recovery attempts to full reinitialization if necessary. This should help the system recover gracefully from sensor lockups and improve reliability.

Introduced a multi-stage watchdog recovery system. After a configurable number of simple recovery attempts (`LSM6DSO_SIMPLE_WATCHDOG_RECOVERY_ATTEMPTS`), the system escalates to a forced sensor reinitialization via the new `prv_lsm6dso_force_reinit()` function.
* Added the `s_watchdog_recovery_attempts` counter to track recovery attempts and control escalation.
* Improved logging for watchdog events, recovery attempts, and error conditions for debugging and monitoring.

Enhanced error handling when configuring interrupt routes logs failures and ensures the external interrupt is always re-enabled for future recovery, even if routing fails. Improved clearing of pending interrupt sources checks return values and logs warnings on failure. These changes make the IMU driver more resilient to sensor failures and easier to diagnose in the field.

### DIFF
--- a/src/fw/drivers/imu/lsm6dso/lsm6dso.c
+++ b/src/fw/drivers/imu/lsm6dso/lsm6dso.c
@@ -45,6 +45,7 @@ static void prv_lsm6dso_configure_shake(bool enable, bool sensitivity_high);
 static void prv_lsm6dso_interrupt_handler(bool *should_context_switch);
 static void prv_lsm6dso_process_interrupts(void);
 static void prv_lsm6dso_interrupt_watchdog_callback(void *data);
+static bool prv_lsm6dso_force_reinit(void);
 static bool prv_is_vibing(void);
 typedef struct {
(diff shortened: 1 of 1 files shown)


jplexer:
i am worried as well but I cannot for the life of it figure what the hell the problem is

gmarull:
> i am worried as well but I cannot for the life of it figure what the hell the problem is

we should not discard Asterix being too slow to respond, provided we have some "general system slowness.", and that causes issues.

PR #297, opened 2025-09-25
Fix dual slot report, add slot info

gmarull:


gmarull:
@jplexer merging as this is a trivial fix

PR #313, opened 2025-09-30
fw/main: do not initialize boot bits for Obelix

gmarull:
Bootloader takes care.


jplexer (approved):
> bootloader takes care

and we love it for that

PR #326, opened 2025-10-02
fix backlight option hiding itself

EladDv:
When turning the backlight off the option disappears and you are not able to turn it back on again
This is because we move from back to front to hide the options so we accidentally move by 1 too much

Tested on silk, spalding,snowy


jplexer (changes requested):
Please fix compliance check,
Format commit correctly (i.e. by amending) and also sign off your commit

EladDv:
I'm just gonna redo the pr, you can delete this
sorry :sweat_smile:

PR #340, opened 2025-10-07
fw/pfs: fix crash when removing file that's in use [FIRM-638]

jplexer:
The firmware would crash with `PBL_CROAK("Cannot delete app_comm, it is currently in use")` when there was concurrent access to PFS files. This occurred in the following scenario:

1. `settings_file.c` opens a file to validate its header
2. File has invalid/corrupt header, so code calls `pfs_close_and_remove()`
3. Meanwhile, another thread (e.g., `app_session_capabilities.c`) opens the same file
4. `pfs_remove()` finds the file is still open and calls `PBL_CROAK()`, crashing the system

## Solution
- **pfs.c**: Changed `pfs_remove()` to return `E_BUSY` error instead of croaking when a file is in use (lines 1493-1497)
- **settings_file.c**: Don't attempt to remove files with invalid magic headers - just close them and return an error (line 78)
- **settings_file.c**: Add error handling when removing files with wrong versions, so failures are logged but don't cause crashes (lines 86-92)

### DIFF
--- a/src/fw/services/normal/settings/settings_file.c
+++ b/src/fw/services/normal/settings/settings_file.c
@@ -19,6 +19,7 @@
 
 #include "drivers/rtc.h"
 #include "kernel/pbl_malloc.h"
+#include "kernel/util/sleep.h"
 #include "services/normal/filesystem/pfs.h"
 #include "system/logging.h"
 #include "system/passert.h"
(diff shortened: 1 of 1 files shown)


src/fw/services/normal/settings/settings_file.c
```diff
@@ -75,15 +75,21 @@ static status_t prv_open(SettingsFile *file, const char *name, uint8_t flags, in
...
   if (file_hdr.version > SETTINGS_FILE_VERSION) {
     PBL_LOG(LOG_LEVEL_WARNING,
-            "Unrecognized version %d for file %s, removing...",
+            "Unrecognized version %d for file %s, closing and removing...",
             file_hdr.version, name);
-    pfs_close_and_remove(fd);
+    status_t remove_status = pfs_close_and_remove(fd);
+    if (FAILED(remove_status)) {
+      PBL_LOG(LOG_LEVEL_WARNING,
+              "Failed to remove %s (status: %"PRId32"), retrying open without removal",
```
gmarull: where is retry happening?

jplexer: I forgot to push the retry logic, sorry about that

src/fw/services/normal/settings/settings_file.c
```diff
@@ -75,15 +76,44 @@ static status_t prv_open(SettingsFile *file, const char *name, uint8_t flags, in
...
+        break;
+      }
+    }
+
+    if (FAILED(remove_status)) {
+      PBL_LOG(LOG_LEVEL_WARNING,
+              "Failed to remove %s after %d attempts (status: %"PRId32")",
+              name, max_retries, remove_status);
+      return remove_status;
+    }
```
gmarull, line 115: Honestly, this looks like a workaround papering over real issues. First, why do we try to delete an open file? Do we have concurrent access to this file?

jplexer: yeah I guess you are right, let me look into this further and I'll open a new pr ^^

PR #341, opened 2025-10-07
fw/pfs: fix crash when removing file that's in use

jplexer:
Fix a race condition where pfs_remove() would crash with PBL_CROAK when attempting to delete a file that's currently open by another thread.

The issue occurred when the "app_comm" settings file was accessed concurrently without synchronization:
- Thread A: settings_file validation discovers invalid/corrupt file
- Thread B: app_session_capabilities queries the same file
- Thread A tries to remove the file and finds it's still open, crashes

Root cause: No mutex protecting concurrent access to app_comm file.

Changes:
- pfs_remove(): Return E_BUSY instead of croaking when file is in use
- settings_file: Don't try to remove files with invalid headers during open - just close and return error. File removal should only happen in controlled contexts (e.g., during init when holding a mutex)
- app_session_capabilities: Add mutex to synchronize all access to the app_comm file. Handle invalid files safely during init.

Fixes FIRM-638

### DIFF
--- a/src/fw/services/normal/comm_session/app_session_capabilities.c
+++ b/src/fw/services/normal/comm_session/app_session_capabilities.c
@@ -17,6 +17,7 @@
 #include "process_management/app_install_types.h"
 #include "process_management/app_manager.h"
 #include "services/common/comm_session/app_session_capabilities.h"
+#include "services/normal/filesystem/pfs.h"
 #include "services/normal/settings/settings_file.h"
 #include "util/units.h"
 
@@ -98,8 +99,16 @@ void comm_session_app_session_capabilities_evict(const Uuid *app_uuid) {
 
 void comm_session_app_session_capabilities_init(void) {
   SettingsFile settings_file;
-  if (PASSED(prv_open(&settings_file))) {
+  status_t open_status = prv_open(&settings_file);
+  if (PASSED(open_status)) {
     settings_file_rewrite(&settings_file, prv_rewrite_cb, NULL);
     settings_file_close(&settings_file);
+  } else if (open_status == E_INVALID_OPERATION) {
+    // File has invalid format (wrong magic or version), remove it and recreate
+    pfs_remove(APP_SESSION_CAPABILITIES_CACHE_FILENAME);
+    // Try to open again to create a fresh file
+    if (PASSED(prv_open(&settings_file))) {
+      settings_file_close(&settings_file);
+    }
   }
 }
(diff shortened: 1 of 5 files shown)


src/fw/services/normal/comm_session/app_session_capabilities.c
```diff
@@ -89,17 +97,35 @@ static void prv_rewrite_cb(SettingsFile *old_file,
...
     settings_file_close(&settings_file);
   }
+
+  mutex_unlock(s_app_comm_mutex);
 }
 
 void comm_session_app_session_capabilities_init(void) {
+  s_app_comm_mutex = mutex_create();
+
+  mutex_lock(s_app_comm_mutex);
```
gmarull: I guess it is safe to not lock here; you're just creating the mutex so any concurrent access at this point would be racey anyway.

src/fw/services/normal/settings/settings_file.c
```diff
@@ -75,16 +75,16 @@ static status_t prv_open(SettingsFile *file, const char *name, uint8_t flags, in
 
   if (memcmp(&file_hdr.magic, SETTINGS_FILE_MAGIC, sizeof(file_hdr.magic)) != 0) {
     PBL_LOG(LOG_LEVEL_ERROR, "Attempted to open %s, not a settings file.", name);
-    pfs_close_and_remove(fd);
+    pfs_close(fd);
     return E_INVALID_OPERATION;
```
gmarull, line 79: is there any other code paths relying on this behavior, ie, close and remove

gmarull: (app session caps now handles this gracefully, but what about others?)

jplexer: yes, there totally are! adding them

src/fw/services/normal/filesystem/pfs.c
```diff
@@ -1491,8 +1491,10 @@ status_t pfs_remove(const char *name) {
   status_t rv = get_avail_fd(name, &fd, false);
   if (rv >= FDAlreadyLoaded) { // the file is in the cache
     if (rv == FDBusy) {
-      PBL_CROAK("Cannot delete %s, it is currently in use",
-                PFS_FD(fd).file.name);
+      PBL_LOG(LOG_LEVEL_WARNING, "Cannot delete %s, it is currently in use",
+              PFS_FD(fd).file.name);
+      rv = E_BUSY;
+      goto cleanup;
```
jwise, line 1497: My $0.02: we absolutely *MUST NOT* remove the `CROAK` in an error case that indicates system inconsistency.  If something is specified to never happen, other parts of the system may depend on that not happening, and that might result in consequences more dire than this particular case if it does.

PR #342, opened 2025-10-07
Revert "third_party/memfault: wire up Pebble DEVICE analytics to memfault"

jplexer:
This reverts the analytics integration from commit b028e387.

The integration added all 185 Pebble analytics metrics to Memfault heartbeat data, which caused metrics to stop appearing in Memfault. The large payload (185 metrics vs the original ~15) likely exceeds size limits or schema expectations in the mobile app or Memfault backend.

Reverting to the original small set of metrics:
- Battery voltage
- Heap/memory usage
- Accelerometer diagnostics (LSM6DSO)
- Back button long press counter

We can revert this for now so that memfault works and see what is actually going wrong later


ericmigi:
ok to revert

ericmigi (approved):

PR #353, opened 2025-10-13
Implementation HAL_GetTick() for SifliSDK

Jiyun-Yang:
Add strong implementation HAL_GetTick() for SifliSDK, as the I2C HAL driver would always busy wait if communicate with a non-exist device.

### DIFF
--- a/third_party/hal_sifli/hal_glue.c
+++ b/third_party/hal_sifli/hal_glue.c
@@ -1,0 +1,22 @@
+/*
+ * Copyright 2025 Core Devices LLC
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <stdint.h>
+
+extern int rtc_get_ticks(void);
+
+uint32_t HAL_GetTick(void) {
+  return (uint32_t)rtc_get_ticks();
+}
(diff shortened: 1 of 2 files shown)


src/fw/drivers/sf32lb52/rtc.c
```diff
@@ -244,3 +244,7 @@ bool rtc_is_timezone_set(void) { return 0; }
 void rtc_enable_backup_regs(void) {}
 
 void rtc_calibrate_frequency(uint32_t frequency) {}
+
+uint32_t HAL_GetTick(void) {
+  return (uint32_t)rtc_get_ticks();
+}
```
gmarull: this needs to be implemented in `third_party/hal_sifli`

third_party/hal_sifli/hal_glue.c
```diff
@@ -0,0 +1,22 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#include <stdint.h>
+
+extern int rtc_get_ticks(void);
+
+uint32_t HAL_GetTick(void) {
```
gmarull, line 20: why is file named "ipc_os_port.c"?

gmarull: maybe just "hal_glue.c", and not sf32lb52 specific

third_party/hal_sifli/wscript
```diff
@@ -4,13 +4,15 @@ def configure(conf):
...
         conf.env.append_unique('DEFINES', 'HAL_NMI_HANLDER_OVERRIDED')
+        conf.env.append_unique('LINKFLAGS', '-Wl,--undefined=HAL_GetTick')
 
     if conf.env.MICRO_FAMILY == 'SF32LB52':
         conf.env.append_unique('DEFINES', 'SF32LB52X')
 
 
 def build(bld):
-    micro_sources = bld.path.ant_glob('SiFli-SDK/drivers/hal/*.c', excl=['**/bf0_hal_audcodec.c'])
+    micro_sources = ['sf32lb52/hal_glue.c']
```
gmarull: move to the root, not 52 specific, also adjust commit title

third_party/hal_sifli/wscript
```diff
@@ -4,13 +4,15 @@ def configure(conf):
         conf.env.append_unique('DEFINES', 'SOC_BF0_HCPU')
         conf.env.append_unique('DEFINES', 'HAL_TICK_PER_SECOND=1024')
         conf.env.append_unique('DEFINES', 'HAL_NMI_HANLDER_OVERRIDED')
+        conf.env.append_unique('LINKFLAGS', '-Wl,--undefined=HAL_GetTick')
```
gmarull, line 7: why is this needed?

Jiyun-Yang: force to use the strong implementation, it always link to the weak one if without this flag in the .a lib

PR #354, opened 2025-10-13
fw/services/normal/activity: fix race condition in prv_reset_state [FIRM-672]

jplexer:
Fixes a use-after-free race condition that causes a HardFault when plugging in the charger while activity tracking is running.


src/fw/services/normal/activity/kraepelin/kraepelin_algorithm.c
```diff
@@ -391,8 +391,14 @@ static void prv_reset_step_activity_state(KAlgStepActivityState *state) {
 
 // ----------------------------------------------------------------------------------------
 static void prv_reset_state(KAlgState *state) {
-  prv_reset_step_activity_state(&state->walk_state);
-  prv_reset_step_activity_state(&state->run_state);
+  // Only reset step activity states if they're not currently in progress to avoid
+  // race conditions with the minute handler (FIRM-672)
```
gmarull:
```suggestion
  // race conditions with the minute handler
```

PR #358, opened 2025-10-13
fw/services/common: conditionally initialize accel_manager [FIRM-656]

jplexer:
Only initialize accel_manager if the accelerometer driver successfully initialized.

### DIFF
--- a/src/fw/services/common/services_common.c
+++ b/src/fw/services/common/services_common.c
@@ -27,6 +27,7 @@
 #ifdef MICRO_FAMILY_STM32F2
 #include "services/common/legacy/factory_registry.h"
 #endif
+#include "drivers/accel.h"
 #include "services/common/accel_manager.h"
 #include "services/common/bluetooth/bluetooth_persistent_storage.h"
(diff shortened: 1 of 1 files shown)


src/fw/services/common/services_common.c
```diff
@@ -42,13 +43,25 @@
...
   firmware_update_init();
   put_bytes_init();
   poll_remote_init();
-  accel_manager_init();
+
+  // Only initialize the accelerometer manager if the accelerometer driver
+  // successfully initialized. We test this by attempting to peek at a sample.
+  AccelDriverSample probe_sample;
+  if (accel_peek(&probe_sample) == 0) {
+    accel_manager_init();
```
gmarull, line 60: feels fragile, should we have a specific error code, or an API ?

jplexer: eh honestly this should be fine since most drivers return -1 if the accel is not initialized

gmarull: fair enough, our error handling is pretty bad everywhere :')

PR #360, opened 2025-10-13
Support for choosing JavaScript engine [FIRM-686]

jplexer:
Prerequisite that allows us to select what JavaScript engine to compile the firmware with (once we add others)

### DIFF
--- a/docs/development/options.md
+++ b/docs/development/options.md
@@ -5,8 +5,9 @@ Below you will find a list of the most relevant ones.
 
 ## Main features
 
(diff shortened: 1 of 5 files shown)


docs/development/options.md
```diff
@@ -5,8 +5,9 @@ Below you will find a list of the most relevant ones.
 
 ## Main features
 
-:`--nojs`:
-  Disable Javascript support
+:`--js_engine`:
```
gmarull: `--js-engine`

PR #362, opened 2025-10-14
fw/services/normal/comm_session: improve mutex handling and file operations [FIRM-638]

jplexer:
Adds proper mutex synchronization to the app session capabilities cache to prevent race conditions when accessing the shared settings file.

### DIFF
--- a/src/fw/services/normal/comm_session/app_session_capabilities.c
+++ b/src/fw/services/normal/comm_session/app_session_capabilities.c
@@ -18,15 +18,34 @@
 #include "process_management/app_manager.h"
 #include "services/common/comm_session/app_session_capabilities.h"
 #include "services/normal/settings/settings_file.h"
+#include "os/mutex.h"
+#include "system/logging.h"
+#include "system/passert.h"
 #include "util/units.h"
 
 #define APP_SESSION_CAPABILITIES_CACHE_FILENAME "app_comm"
 
 #define APP_SESSION_CAPABILITIES_CACHE_FILE_MAX_USED_SPACE (KiBYTES(2))
 
+static PebbleMutex *s_mutex;
+
 static status_t prv_open(SettingsFile *settings_file) {
   return settings_file_open(settings_file, APP_SESSION_CAPABILITIES_CACHE_FILENAME,
                             APP_SESSION_CAPABILITIES_CACHE_FILE_MAX_USED_SPACE);
+}
+
+static status_t prv_open_locked(SettingsFile *settings_file) {
+  mutex_lock(s_mutex);
+  status_t status = prv_open(settings_file);
+  if (FAILED(status)) {
+    mutex_unlock(s_mutex);
+  }
+  return status;
+}
+
+static void prv_close_and_unlock(SettingsFile *settings_file) {
+  settings_file_close(settings_file);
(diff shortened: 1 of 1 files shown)


src/fw/services/normal/comm_session/app_session_capabilities.c
```diff
@@ -18,27 +18,74 @@
...
+  }
+
+  PebbleMutex *new_mutex = mutex_create();
+  if (new_mutex == NULL) {
+    return NULL;
+  }
+
+  if (!__sync_bool_compare_and_swap(&s_state.mutex, NULL, new_mutex)) {
+    mutex_destroy(new_mutex);
+  }
```
gmarull: why isn't mutex created once?

jplexer: claude...

src/fw/services/normal/comm_session/app_session_capabilities.c
```diff
@@ -90,16 +109,23 @@ static void prv_rewrite_cb(SettingsFile *old_file,
...
   }
 }
 
 void comm_session_app_session_capabilities_init(void) {
+  s_mutex = mutex_create();
+  if (s_mutex == NULL) {
+    PBL_LOG(LOG_LEVEL_ERROR,
+            "Failed to initialize mutex for app session capabilities cache");
+    return;
+  }
```
gmarull: just assert

src/fw/services/normal/comm_session/app_session_capabilities.c
```diff
@@ -18,27 +18,51 @@
...
+
 static status_t prv_open(SettingsFile *settings_file) {
   return settings_file_open(settings_file, APP_SESSION_CAPABILITIES_CACHE_FILENAME,
                             APP_SESSION_CAPABILITIES_CACHE_FILE_MAX_USED_SPACE);
 }
 
+static status_t prv_open_locked(SettingsFile *settings_file) {
+  if (s_mutex == NULL) {
+    return E_INVALID_OPERATION;
+  }
```
gmarull: not needed if we assert on creation

src/fw/services/normal/comm_session/app_session_capabilities.c
```diff
@@ -49,28 +73,23 @@ bool comm_session_current_app_session_cache_has_capability(CommSessionCapability
...
   } else {
     // Not connected, use cached data.
-
-    if (FAILED(open_status)) {
+    if (!file_open) {
       // File open failed, no cache available
-      goto done;
+      return ((new_capabilities & capability) != 0);
     }
   }
```
gmarull: else branch seems redundant now?

src/fw/services/normal/comm_session/app_session_capabilities.c
gmarull: Vibe coding provides consistently bad results with anything that requires a minimal level of reasoning. They're just advanced autocompleters, with the danger of seeming good enough.

src/fw/services/normal/comm_session/app_session_capabilities.c
jplexer: image

PR #363, opened 2025-10-14
Reduce latency of voice session init + UI

crc-32:
- Removes progress bar and associated intermediate state from the startup transition of the voice window
 - Audio session setup no longer waits on responsiveness granted callback (connection interval reduction), also now just returns from setup function rather than taking a callback for session setup success (nothing blocking is waited on now)

Associated tests were edited/removed to match new expected functionality

### DIFF
--- a/src/fw/services/normal/audio_endpoint.c
+++ b/src/fw/services/normal/audio_endpoint.c
@@ -34,7 +34,6 @@ _Static_assert(ACTIVE_MODE_TIMEOUT > ACTIVE_MODE_START_BUFFER,
 
 typedef struct {
   AudioEndpointSessionId id;
-  AudioEndpointSetupCompleteCallback setup_completed;
   AudioEndpointStopTransferCallback stop_transfer;
   TimerID active_mode_trigger;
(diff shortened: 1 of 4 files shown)


src/fw/applib/voice/voice_window.c
```diff
@@ -1149,12 +1133,7 @@ static void prv_do_transition(VoiceUiData *data, VoiceUiState state) {
       // Stop fly in animation. Start pulsing dot animation. Start progress bar animation
       prv_stop_fly_dot(data);
       prv_show_mic_dot_pulse(data);
-      prv_show_progress_bar(data, true /* animated */);
-      break;
-
-    case StateStopWaitForReady:
-      // Shrink progress bar
-      prv_shrink_progress_bar(data);
+      //prv_show_progress_bar(data, true /* animated */);
```
gmarull: remove commented code?

jplexer (approved):

jwise:
I would have preferred if these were ifdeffed to PLATFORM_ASTERIX | PLATFORM_OBELIX...

PR #364, opened 2025-10-15
Improve sf32lb52 low power mode stability.

cameled:
This improve the sf32lb52 low power mode stability.

Low power current on big board is mainly from the blue led, nearly 2 mA. Also find current leak on HR pin, about 150uA. Current low power performance as below. (still have some IO leak?)

Screenshot from 2025-10-20 01-49-48

In deepsleep mode, HXT48 clock should be enable. its effective the LCPU/BLE and wakeup time. After HXT48 clock enable, BLE worked (not full tested), and wakeup time drop to 1ms.

Remain work:
1. Should add button to AON wakeup source, or the UI response is slow.
2.

### DIFF
--- a/src/fw/board/boards/board_obelix.c
+++ b/src/fw/board/boards/board_obelix.c
@@ -647,13 +647,11 @@ void board_early_init(void) {
   BSP_System_Config();
 
   HAL_HPAON_StartGTimer();
-#ifdef SF32LB52_USE_LXT
+
   HAL_PMU_EnableRC32K(1);
 
   HAL_PMU_LpCLockSelect(PMU_LPCLK_RC32);
-#else
-  HAL_PMU_LpCLockSelect(PMU_LPCLK_RC10);
-#endif
+
   HAL_PMU_EnableDLL(1);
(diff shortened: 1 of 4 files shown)


src/fw/board/boards/board_obelix.c
```diff
@@ -675,6 +675,11 @@ void board_early_init(void) {
 
   __HAL_SYSCFG_CLEAR_SECURITY();
   HAL_EFUSE_Init();
+
+  // Reset HR pin, clear IE, PE
+  hwp_pinmux1->PAD_PA09 &= ~((1 << 6) | (1 << 4));
+  hwp_pinmux1->PAD_PA20 &= ~((1 << 6) | (1 << 4));
+  hwp_pinmux1->PAD_PA25 &= ~((1 << 6) | (1 << 4));
```
cameled, line 685: Leak is from PA09 or PA20(forget which one). As I tested, after reset that pin, current drop 150uA.

gmarull:
> @gmarull The black issue fixed.

I can see firmware is stable now, average current is:

image

PR #367, opened 2025-10-15
drivers/nrf5/qspi: disable the IRQ when polling for very short reads

jwise:
In 2480779da52a05be852543d8aff0204afbbb5187, we found that FreeRTOS semaphores and context switches were very expensive.  They are, but it is also very expensive just to have an IRQ at all (4us or so).  For that, we avoid the nrfx_ API (which is inexorably committed to an IRQ, once set up with a handler), and instead use the nrf_ API to turn off the IRQ, set up the read, and spin-wait until it completes.

Performance improvements:

 * 4 bytes: 30us -> 25us (-5us)
 * 5 bytes: 42us -> 34us (-8us)
 * 16 bytes: 33us -> 29us (-4us)
 * 64 bytes: 45us -> 41us (-4us)
 * 256 bytes: 93us -> 89us (-4us)
 * 1024 bytes: 316us -> 315us (-1us; presumably noise)

Presumably the next improvement is to speed up the mutex lock / unlock, now that the 'true cost' of a read is only 9us.

### DIFF
--- a/src/fw/drivers/nrf5/qspi.c
+++ b/src/fw/drivers/nrf5/qspi.c
@@ -241,6 +241,26 @@ void qspi_flash_set_lower_power_mode(QSPIFlash *dev, bool active) {
 
 /*** Underlying small flash I/O helpers. ***/
 
+/** Read from QSPI, but poll instead of waiting for the ISR.
+ *
+ * For very short reads, the interrupt overhead is really quite substantial.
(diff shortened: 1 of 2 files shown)


gmarull (changes requested):
Let's wait for the refactored driver (hopefully tomorrow if Obelix does not come up again!); we have a release that is stable enough for internal testing, so any work now should go toward merging things back to main.

PR #370, opened 2025-10-15
treewide: add support for "silk_flint" platform

jplexer:
because there is no "nrf52" version of QEMU, we hack together a "Silk and It's Completely Different but Also Still Silk" that has all the APIs we create for flint


gmarull (commented):
maybe build this on CI instead of silk then?

gmarull (approved):

PR #376, opened 2025-10-17
nRF QSPI flash driver refactor

gmarull:
Notes:

- Enable/disable is synchronous, as it appears to be really fast
- HFXO enable/disable to cover anomaly 244 is not added; at 8 MHz it does not appear to happen (Nordic confirmed at low speeds it should not be observed). 

Measures with this PR:

```
benchmarking 4 bytes...
  -> 4 bytes: 16384 iters in 392 ticks = 23 us/iter
benchmarking 5 bytes...
  -> 5 bytes: 16384 iters in 526 ticks = 31 us/iter
benchmarking 16 bytes...
  -> 16 bytes: 16384 iters in 449 ticks = 26 us/iter
benchmarking 64 bytes...
  -> 64 bytes: 8192 iters in 326 ticks = 38 us/iter
benchmarking 128 bytes...
  -> 128 bytes: 8192 iters in 463 ticks = 55 us/iter
benchmarking 256 bytes...
  -> 256 bytes: 4096 iters in 368 ticks = 87 us/iter
benchmarking 512 bytes...
  -> 512 bytes: 4096 iters in 780 ticks = 185 us/iter
benchmarking 1024 bytes...
  -> 1024 bytes: 4096 iters in 1319 ticks = 314 us/iter
```

TODO:

- [x] Consider enabling/disabling QSPI when exiting/entering stop mode (with XIP we will need it anyway).
- [ ] Add support for anomaly 215 (needed for XIP)

### DIFF
--- a/src/fw/board/boards/board_asterix.c
+++ b/src/fw/board/boards/board_asterix.c
@@ -239,7 +239,4 @@ void board_init(void) {
   i2c_use(I2C_DA7212);
   i2c_write_block(I2C_DA7212, 2, da7212_powerdown);
   i2c_release(I2C_DA7212);
-  
-  // XXX: FIRM-264: stop mode breaks NimBLE
-  stop_mode_disable(InhibitorMain);
 }


src/fw/drivers/nrf5/qspi.c
```diff
@@ -1,138 +1,207 @@
...
-  prv_poll_bit(dev->qspi, dev->state->part->instructions.rdsr1,
-               dev->state->part->status_bit_masks.write_enable, true /* set */, QSPI_NO_TIMEOUT);
+static inline void prv_cinstr_write(QSPIFlash *dev, uint8_t instr, const void *data, size_t len) {
+  prv_cinstr_write_read(dev, instr, data, NULL, len);
 }
 
-static void prv_wait_for_completion(QSPIFlash *dev) {
-  if (!dev->state->coredump_mode) {
-    xSemaphoreTake(dev->qspi->state->dma_semaphore, portMAX_DELAY);
+static void prv_read(QSPIFlash *dev, void *buf, size_t len, uint32_t addr) {
```
jwise, line 151: assert on buf alignment?

src/fw/console/prompt_commands.c
```diff
@@ -654,7 +654,9 @@ void command_flash_benchmark() {
   s_flash_benchmark(5);
   s_flash_benchmark(16);
   s_flash_benchmark(64);
+  s_flash_benchmark(128);
   s_flash_benchmark(256);
+  s_flash_benchmark(512);
```
jwise, line 659: what are the results for the refactor? :)

src/fw/drivers/nrf5/qspi.c
```diff
@@ -1,138 +1,207 @@
...
 #include "kernel/util/delay.h"
 #include "kernel/util/sleep.h"
-#include "kernel/util/stop.h"
-#include "mcu/cache.h"
-#include "system/logging.h"
 #include "system/passert.h"
 #include "util/math.h"
 
-#define NRF5_COMPATIBLE
+#define CMSIS_COMPATIBLE
```
jwise, line 31: hm?

src/fw/drivers/nrf5/qspi.c
```diff
@@ -1,138 +1,207 @@
...
-      return false;
-    }
-    delay_us(1);
+static void prv_put(QSPIFlash *dev) {
+  nrf_qspi_int_disable(NRF_QSPI, NRF_QSPI_INT_READY_MASK);
+  nrf_qspi_task_trigger(NRF_QSPI, NRF_QSPI_TASK_DEACTIVATE);
+  nrf_qspi_event_clear(NRF_QSPI, NRF_QSPI_EVENT_READY);
+
+  nrf_qspi_disable(NRF_QSPI);
+}
```
jwise: Put flash in DPM?

src/fw/drivers/nrf5/qspi.c
```diff
@@ -333,32 +479,31 @@ void qspi_flash_read_blocking(QSPIFlash *dev, uint32_t addr, void *buffer, uint3
   buf_suf = (uint8_t)((length - buf_pre) % 4U);
   buf_mid = length - buf_pre - buf_suf;
 
+  prv_get(dev);
+
```
jwise: this is quite expensive around every read

gmarull: it is actually not that expensive compared to e.g. any mutex lock, NT etc.

src/fw/drivers/nrf5/qspi.c
gmarull: we just need CMSIS APIs from mcu.h (mcu.h should really disappear)

src/fw/console/prompt_commands.c
gmarull: see updated comment

src/fw/drivers/nrf5/qspi.c
gmarull: callers already take care of that

gmarull:
> > -> 4 bytes: 8192 iters in 307 ticks = 36 us/iter
> > -> 5 bytes: 8192 iters in 379 ticks = 45 us/iter
> > -> 16 bytes: 8192 iters in 341 ticks = 40 us/iter
> 
> Hm, this is substantially slower than #367 (25us; 34us; 29us) and even slower than the old implementation (30us; 42us; 33us), on what is a pretty hot path inside PebbleOS that was already causing WDT issues...

I prefer to do any optimization on top of clean code; the previous driver was becoming unreadable, with too many dirty hacks. WDT was also triggered when ops took much longer than what we have with this PR. We'll also likely need to move enable/disable to the stop mode path when we enable XIP. I need to test that as well before merging anything.

gmarull:
ops, updated benchmarks, they were from an old run :-( it's a bit better now

PR #379, opened 2025-10-18
fw/apps/system_apps/settings: add accent colors

EladDv:
Added an accent color settings module for both apps screen and the settings app

tested on silk (just doesn't show up and uses the defaults), snowy and spalding in QEMU

for it to be expanded in the future for other apps will probably need a rewrite (unclear how dark/light/default modes will integrate into this too, might be able to squeeze it in this structure)


jplexer (changes requested):
you shouldnt be able to set colors that text isnt readable on imho.

EladDv:
> you shouldnt be able to set colors that text isnt readable on imho.

Which colors are you thinking about? 
Changing it is super easy - just take them out of the list of colors in the settings_themes.h file (unless you want to separate the 2 lists but then it's just duplicating them and changing a few variables)

personally I think user freedom is always a plus (especially on a more "enthusiast" platform like this)

jplexer:
yeah I guess that would make sense!

jplexer:
well you dont have to change the entire background, just the text color :grin:

EladDv:
BTW still left in the different menus for light and dark (should probably rename them to light accent and dark accent though)

EladDv:
Ok yeah all those git shenanigans reverted me back to a bad version, now should be fixed
I also had dark and light accidentally reversed and moved also moved the menu item to just below display

src/fw/apps/system_apps/settings/settings_menu.c
```diff
@@ -55,6 +57,9 @@ static const SettingsModuleGetMetadata s_submodule_registry[] = {
 #endif
   [SettingsMenuItemDisplay]       = settings_display_get_info,
   [SettingsMenuItemSystem]        = settings_system_get_info,
+#if PBL_COLOR
+  [SettingsMenuItemThemes]        = settings_themes_get_info,
```
jplexer, line 60: nit: adapt to actual order in settings

jplexer (approved):
codewise looks good, just one nitpick :)

src/fw/apps/system_apps/settings/settings_menu.c
EladDv: sure, I did that originally then reverted before I sent the commit because I thought maybe I shouldnt :P

PR #396, opened 2025-10-28
fw/shell/normal: disable welcome message

jplexer:
Disables the "Pebble Updated!" welcome message


ericmigi:
Oh yeah! Definitely don't need that one

PR #402, opened 2025-10-31
add .claude to gitignore

jplexer:


jplexer:
@gmarull merge requirements check for build but ofc this is not building something due to the rules in the action (as it should)
can you override pls :)

PR #405, opened 2025-10-31
comm/ble: Fix deadlock between NimbleHost and KernelMain threads [FIRM-646]

jplexer:
Fix a lock ordering violation that causes deadlock between the NimbleHost and KernelMain threads. The deadlock occurs when:

1. KernelMain holds bt_lock and tries to acquire TaskTimerManager mutex (by calling new_timer_create()) 
2. NimbleHost holds TaskTimerManager mutex and tries to acquire bt_lock (when calling back into firmware code)

### DIFF
--- a/src/fw/comm/ble/kernel_le_client/ppogatt/ppogatt.c
+++ b/src/fw/comm/ble/kernel_le_client/ppogatt/ppogatt.c
@@ -352,14 +352,14 @@ static void prv_timer_callback(void *unused) {
 
 // -------------------------------------------------------------------------------------------------
 
-static PPoGATTClient *prv_create_client(void) {
+static PPoGATTClient *prv_create_client(TimerID timer) {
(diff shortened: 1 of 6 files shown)


src/fw/comm/ble/kernel_le_client/ppogatt/ppogatt.c
```diff
@@ -861,10 +861,23 @@ void ppogatt_invalidate_all_references(void) {
 }
 
 void ppogatt_handle_service_discovered(BLECharacteristic *characteristics) {
+  // Create timer outside of bt_lock to avoid deadlock with NimbleHost.
+  // new_timer_create() acquires TaskTimerManager mutex, which may be held by NimbleHost
+  // when it's trying to acquire bt_lock, leading to a lock ordering deadlock.
+  TimerID timer = new_timer_create();
+  if (!timer) {
+    return;
+  }
```
gmarull: maybe assert?

PR #408, opened 2025-10-31
fw/console: enable 'rgb' command on Obelix

gmarull:
Usage is `rgb 0xRRGGBB`.


jplexer (approved):
very compilicated change, I need a minute to think about this

PR #409, opened 2025-10-31
services/bluetooth: Defer bonding callbacks to KernelMain to avoid deadlock [FIRM-723]

jplexer:
- schedule BLE bonding change notifications on the launcher task when invoked from NimBLE
- queue bonding creation finalization to KernelMain so bt_lock is taken outside the host stack

### DIFF
--- a/src/fw/services/common/bluetooth/bonding.c
+++ b/src/fw/services/common/bluetooth/bonding.c
@@ -20,11 +20,45 @@
 
 #include "services/common/bluetooth/bluetooth_persistent_storage.h"
 #include "services/common/bluetooth/local_addr.h"
+#include "kernel/event_loop.h"
+#include "kernel/pbl_malloc.h"
 #include "system/logging.h"
 
 #include <bluetooth/bonding_sync.h>
 #include <bluetooth/bluetooth_types.h>
 #include <bluetooth/sm_types.h>
+
+typedef struct {
+  BTBondingID bonding_id;
+  BTDeviceAddress addr;
+  bool is_gateway;
+} CreateBondingContext;
+
+static void prv_finalize_create_bonding(BTBondingID bonding_id,
+                                        const BTDeviceAddress *addr,
+                                        bool is_gateway) {
+  bt_lock();
+  GAPLEConnection *connection = gap_le_connection_by_addr(addr);
+  if (connection) {
+    connection->bonding_id = bonding_id;
+    connection->is_gateway = is_gateway;
+
+    if (!connection->is_gateway) {
+      PBL_LOG(LOG_LEVEL_DEBUG, "New bonding is not gateway?");
+    }
+
+    gap_le_device_name_request(&connection->device);
+  } else {
+    PBL_LOG(LOG_LEVEL_ERROR, "Couldn't find connection for bonding!");
+  }
+  bt_unlock();
+}
+
+static void prv_finalize_create_bonding_cb(void *data) {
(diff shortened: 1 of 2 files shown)


src/fw/services/normal/bluetooth/bluetooth_persistent_storage.c
```diff
@@ -597,6 +603,33 @@ static void prv_call_ble_bonding_change_handlers(BTBondingID bonding,
...
+                                                 BtPersistBondingOp op) {
+  if (launcher_task_is_current_task()) {
+    prv_call_ble_bonding_change_handlers_impl(bonding, op);
+    return;
+  }
+
+  BleBondingChangeContext *context = kernel_malloc(sizeof(*context));
+  if (!context) {
+    PBL_LOG(LOG_LEVEL_ERROR, "Failed to alloc bonding change ctx; running inline");
+    prv_call_ble_bonding_change_handlers_impl(bonding, op);
```
gmarull: is this safe?

jplexer: Actually if the kernel_malloc fails i guess we have some other problems. so should we assert?

gmarull: well, it can or cannot be a fatal failure, depending on the context. In pebbleos we (ab)use asserts to handle these kind of failures instead of gracefully propagating things up and crashing somewhere.

jplexer: ok so to answer your original question: kinda. it’s the original behaviour and so safe as a last‑resort fallback, but it can reintroduce the deadlock you I fixed if the malloc fails while running on the NimBLE host thread

gmarull: Any path that can deadlock should be deleted, better crash immediately (or WDT would anyway do it for you)

src/fw/services/normal/bluetooth/bluetooth_persistent_storage.c
```diff
@@ -597,6 +603,29 @@ static void prv_call_ble_bonding_change_handlers(BTBondingID bonding,
...
+
+static void prv_call_ble_bonding_change_handlers(BTBondingID bonding,
+                                                 BtPersistBondingOp op) {
+  if (launcher_task_is_current_task()) {
+    prv_call_ble_bonding_change_handlers_impl(bonding, op);
+    return;
+  }
+
+  BleBondingChangeContext *context = kernel_malloc(sizeof(*context));
+  PBL_ASSERTN(context);
```
gmarull: actually, use `kernel_malloc_check`

src/fw/services/common/bluetooth/bonding.c
```diff
@@ -44,24 +79,22 @@ void bt_driver_cb_handle_create_bonding(const BleBonding *bonding,
...
-      PBL_LOG(LOG_LEVEL_ERROR, "Couldn't find connection for bonding!");
-    }
+  if (launcher_task_is_current_task()) {
+    prv_finalize_create_bonding(bonding_id, addr, bonding->is_gateway);
+    return;
   }
-  bt_unlock();
+
+  CreateBondingContext *context = kernel_malloc(sizeof(*context));
+  PBL_ASSERTN(context);
```
gmarull: same: kernel_malloc_check

PR #419, opened 2025-11-04
fw: Add Motion Sensitivity setting and LSM6DSO mapping

jplexer:
Add an Asterix-only settings menu and preference (0-100%, default 100%) to control accelerometer shake sensitivity. Expose shell prefs and UI labels, and store the value in prefs_values. Introduce accel_manager_update_sensitivity and a driver API to set percent. Driver interpolates between board low/high wake thresholds (100% = most sensitive = low threshold), clamps values, logs changes, and reconfigures shake detection immediately.

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_system.c
+++ b/src/fw/apps/system_apps/settings/settings_system.c
@@ -16,6 +16,7 @@
 
 #include "settings_factory_reset.h"
 #include "settings_menu.h"
+#include "settings_option_menu.h"
 #include "settings_system.h"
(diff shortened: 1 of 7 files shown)


gmarull:
> Don't think that there is an right threshold. Peoples have different preferences. The users should be able to have some options for this. A carpenter and a office worker have probably different needs for thresholds. I personally would prefer if the threshold setting where easily accessible.

Agreed, but this first needs some debugging to determine which options make sense to expose to the users.

PR #420, opened 2025-11-04
treewide: improve Obelix board build strategy

gmarull:
Add 4 board variants:

- EVT and BB
- DVT and BB2

Each variant will report a different code in firmware metadata. In the release process (or CI), only EVT/DVT are built.


src/include/pebbleos/firmware_metadata.h
```diff
@@ -63,8 +64,11 @@ typedef enum FirmwareMetadataPlatform {
   FirmwareMetadataPlatformPebbleRobertBigboard = 0xf9,
   FirmwareMetadataPlatformPebbleSilkBigboard2 = 0xf8,
   FirmwareMetadataPlatformPebbleRobertBigboard2 = 0xf7,
-  FirmwareMetadataPlatformPebbleSilkFlint = 0xf6,
-  FirmwareMetadataPlatformPebbleSnowyEmery = 0xf5,
+  FirmwareMetadataPlatformPebbleObelixBigboard = 0xf6,
+  FirmwareMetadataPlatformPebbleObelixBigboard2 = 0xf5,
+
+  FirmwareMetadataPlatformPebbleSilkFlint = 0xa0,
+  FirmwareMetadataPlatformPebbleSnowyEmery = 0xa1,
```
jplexer: we should not be reassigning firmware metadata that is already used in places (i.e. a currently already released SDK)

jplexer (changes requested):

src/include/pebbleos/firmware_metadata.h
gmarull: argh, that's unfortunate as the 0xff and down are used for Bigboards...

gmarull: is there any chance to still correct this?

gmarull: not a big deal, just left them untouched.

jplexer (approved):

PR #437, opened 2025-11-06
fw: Add dynamic backlight feature with ALS-based intensity control

jplexer:
Implements user-toggleable dynamic backlight that automatically adjusts backlight intensity based on ambient light sensor readings. When enabled, backlight intensity scales linearly between 5% (Low) and the user's configured max intensity based on ambient light condition.

Note: Threshold values (20/120) are initial estimates and will likely need to be adjusted based on real-world testing and user feedback

### DIFF
--- a/src/fw/board/boards/board_obelix.c
+++ b/src/fw/board/boards/board_obelix.c
@@ -546,6 +546,8 @@ const BoardConfig BOARD_CONFIG = {
   .backlight_on_percent = 25,
   .ambient_light_dark_threshold = 150,
   .ambient_k_delta_threshold = 25,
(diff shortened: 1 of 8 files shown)


src/fw/board/boards/board_obelix.c
```diff
@@ -546,6 +546,10 @@ const BoardConfig BOARD_CONFIG = {
   .backlight_on_percent = 25,
   .ambient_light_dark_threshold = 150,
   .ambient_k_delta_threshold = 25,
+#if CAPABILITY_HAS_DYNAMIC_BACKLIGHT
```
gmarull: ifdef not needed here

PR #445, opened 2025-11-09
feat: add display orientation setting for asterix devices

### LINKED ISSUE #429: Left-handed mode (Display rotation configurability)

### Proposed feature request

Hello,

As a left-hander, I would really appreciate a setting to wear the watch on the right hand. The watch obviously works with the right hand as well, but it is quite uncomfortable, especially since the new buttons are a bit more stiff.

My idea is to enable a "left-handed mode" that would cause two things:

- Rotate the display by 180%
- Remap (swap) the up and down buttons

### Alternatives

_No response_

### Anything else?

I don't know how difficult it is to modify the display driver to make this possible, but know that us left-handers (about every tenth person ;) ) would really appreciate it.

rojas-claudio:
This pull request introduces a new 'Orientation' display setting for asterix devices, allowing the user to more comfortably wear their watch on the right wrist. When enabled, the display will rotate 180º so that Up/Select/Down will be positioned on the left of the watch. As opposed to reaching around with the left thumb to press Up/Select/Down, the navigation buttons can be pressed with the left index finger.

Fixes #429

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_display.c
+++ b/src/fw/apps/system_apps/settings/settings_display.c
@@ -95,6 +95,43 @@ static void prv_intensity_menu_push(SettingsDisplayData *data) {
       true /* icons_enabled */, s_intensity_labels, data);
 }
 
+#if PLATFORM_ASTERIX
+// Orientation Settings
+/////////////////////////////
+static const char *s_display_orientation_labels[] = {
+    i18n_noop("Default"),
+    i18n_noop("Left-Handed"),
+};
+
+static int prv_display_orientation_get_selection_index() {
+  return display_orientation_is_left() ? 1 : 0;
+}
+
+static void prv_display_orientation_menu_select(OptionMenu *option_menu, int selection,
+                                                void *context) {
+  if (prv_display_orientation_get_selection_index() == selection) {
+    // No change
+    app_window_stack_remove(&option_menu->window, true /* animated */);
+    return;
+  }
+
+  display_orientation_set_left(!display_orientation_is_left());
+  app_window_stack_remove(&option_menu->window, true /* animated */);
+}
+
+static void prv_display_orientation_menu_push(SettingsDisplayData *data) {
+  const int index = prv_display_orientation_get_selection_index();
+  const OptionMenuCallbacks callbacks = {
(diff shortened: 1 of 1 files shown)


src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
```diff
@@ -200,12 +214,16 @@ void display_update(NextRowCallback nrcb, UpdateCompleteCallback uccb) {
   desc.tx_length++;
 
   while (nrcb(&row)) {
-    // write row address, data and trailing dummy
```
gmarull, line 203: preserve comment

src/fw/drivers/display/sharp_ls013b7dh01/sharp_ls013b7dh01_nrf5.c
```diff
@@ -27,6 +27,13 @@
 #include "kernel/util/stop.h"
 #include "os/mutex.h"
 #include "system/passert.h"
+
+#include "util/logging.h"
+#include "util/bitset.h"
+#include "util/net.h"
+#include "util/reverse.h"
+#include "util/units.h"
```
gmarull: not all needed?

src/fw/drivers/display/display.h
```diff
@@ -41,6 +41,10 @@ void display_clear(void);
 
 void display_set_enabled(bool enabled);
 
+#if PLATFORM_ASTERIX
+void display_set_rotated(bool rotated);
+#endif
```
gmarull: this API should not be asterix specific, can be generic

gmarull (commented):
overall lgtm, just make it non-asterix specific as this can apply to all watches (and stub drivers if needed)

src/fw/drivers/display/display.h
rojas-claudio: do we want the feature fully implemented for other displays/watches before merging?

gmarull: it can be stubbed for others

gmarull (approved):

PR #448, opened 2025-11-10
Make single click UP/DOWN mappable [FIRM-788]

jplexer:
Tested in QEMU and on Asterix :)


src/fw/shell/normal/prefs.c
```diff
@@ -1055,6 +1080,84 @@ uint8_t quick_launch_get_quick_launch_setup_opened(void) {
...
+      break;
+    case BUTTON_ID_DOWN:
+      key = PREF_KEY_QUICK_LAUNCH_SINGLE_CLICK_DOWN;
+      break;
+    case BUTTON_ID_SELECT:
+    case BUTTON_ID_BACK:
+    case NUM_BUTTONS:
+      break;
+  }
+  PBL_ASSERTN(key);
```
gmarull: default: assert?

src/fw/shell/normal/watchface.c
```diff
@@ -121,29 +132,13 @@ static void prv_launch_launcher_app(ClickRecognizerRef recognizer, void *data) {
   }, recognizer);
 }
 
-#if CAPABILITY_HAS_CORE_NAVIGATION4
```
gmarull, line 124: cap no longer used? maybe delete?

src/fw/apps/system_apps/settings/settings_quick_launch_app_menu.c
```diff
@@ -48,13 +51,42 @@ typedef struct {
...
+      // Tap Up: Only show Timeline Past, hide Timeline Future
+      if (uuid_equal(&entry->uuid, &timeline_uuid)) {
+        return false;
+      }
+    } else if (data->button == BUTTON_ID_DOWN) {
+      // Tap Down: Only show Timeline Future, hide Timeline Past
+      if (uuid_equal(&entry->uuid, &timeline_past_uuid)) {
+        return false;
+      }
+      // We also only want the Health app for Tap Up
```
matejdro, line 79: Is there a reason to limit these? Why wouldn't we allow users to map health and timeline to any button they want?

src/fw/apps/system_apps/launcher/default/launcher_app.c
```diff
@@ -64,6 +66,13 @@ static void prv_will_focus(bool in_focus) {
 // AppMenuDataSource callbacks
 
 static bool prv_app_filter_callback(PBL_UNUSED AppMenuDataSource *source, AppInstallEntry *entry) {
+  // Skip Health app if it's set as the quick launch single-click up action
+  const Uuid health_uuid = UUID_HEALTH_DATA_SOURCE;
```
matejdro, line 70: Can we do the same for both timelines? That way they are accessible even when unbound

src/fw/apps/system_apps/settings/settings_quick_launch_app_menu.c
jplexer: only health is blocked from every button (although I do want to remove this limitation once the app itself is a bit cleaned up, currently it will always be down to exit etc)

Timeline is not being limited for other buttons! its just that timeline future can not be set to up and timeline past cannot be set to down, this prevents unexpected behavior

src/fw/apps/system_apps/launcher/default/launcher_app.c
jplexer: I wouldnt be against this, although I would maybe talk with LCP about this on the discord

jplexer: Requires some changes according to LCP, will be done at a later point

PR #465, opened 2025-11-13
Add support for TZ handling to SF32LB52 RTC driver

gmarull:
Fixes FIRM-735


ericmigi:
it's always a tz problem!

PR #480, opened 2025-11-18
configure sifli chipset power key reset time to 15s

Jiyun-Yang:
So it always use PMIC cool reboot for long press the left button

Fixes FIRM-826

### DIFF
--- a/src/fw/board/boards/board_obelix.c
+++ b/src/fw/board/boards/board_obelix.c
@@ -28,6 +28,7 @@
 
 
 #define HCPU_FREQ_MHZ 240
+#define PWRKEY_RESET_CNT (32000 * 15)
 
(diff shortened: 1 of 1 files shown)


src/fw/board/boards/board_obelix.c
```diff
@@ -28,6 +28,7 @@
 
 
 #define HCPU_FREQ_MHZ 240
+#define PWRKEY_RESET_CNT (32000 * 15)
```
gmarull, line 31: where does 32000 come from? 

Image

don't we clock WDT with RC10K?

gmarull: @cameled ?

cameled: Sorry, I'm missing this issue, will check it tonight.

PR #501, opened 2025-11-25
treewide: move to SPDX tags

gmarull:


jwise:
haha I just did this by hand and had planned to submit it back in this direction. that's what I get for being a day out of date!

PR #503, opened 2025-11-25
services/normal/activity_insights: remove pebble health nag

jplexer:


ericmigi:
I think the problem with removing this is that people will complain when they don't see their step count. So as long as we've got something in the firmware to remind people to turn it on in the app, we can remove this. Ie if you open the help app and you haven't enabled it in mobile app then explain that they have to go into the mobile app and check them

gmarull (approved):

PR #525, opened 2025-11-28
apps/music: Add new Music settings for Volume Controls and Progress Bar

asyba:
Add "Show Volume Controls" and "Show Progress Bar" options to the Music Settings menu.

When "Show Volume Controls" is hidden:
- UP/DOWN buttons perform next/previous track.
- The SELECT button long-press is disabled to prevent entering volume mode.

When "Show Progress Bar" is hidden:
- The progress bar and time labels are hidden.
- The app unsubscribes from the tick timer to save power.

I wanted to this since I'm iOS user and we don't have an alternative like MusicBoss.
I wanted to have a way to have the music player just with the play/pause next/previus buttons for fast click and not deal with the ellipsis button.

I tested on QEMU andmy pebble duo 2.

img2
img3
img4

### DIFF
--- a/src/fw/apps/system_apps/music_app.c
+++ b/src/fw/apps/system_apps/music_app.c
@@ -35,6 +35,7 @@
 #include "process_management/app_manager.h"
 #include "process_state/app_state/app_state.h"
 #include "resource/resource_ids.auto.h"
+#include "shell/prefs.h"
 #include "system/logging.h"
 #include "system/passert.h"
 #include "util/math.h"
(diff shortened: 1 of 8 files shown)


asyba:
> I actually didn't know this until like 2020, but did you know that if you long hold the up or down button, it will change volume without having to enter 'volume'? Or do you mean something else by 'fast click'?

Sorry, perhaps I wasn't clear enough.

Yes a fast click. I want a way to play/pause and rewind/next with just one click.

I know you can hold down the middle ellipsis and/or the up/down volume buttons, but having to press and hold for this I don't like it much.

Obviously, wanting it with just one click means losing the volume control functionality, but in practice, I've rarely needed to change the volume.

But it's all optional; by default, I set it to be as it always was, but I added a settings menu to hide those options if desired.

ericmigi:
probably won't be something we merge - sorry. I think double click to pause is good enough for now. I do love settings…but this feels like a step too far for now. I think once we have the ability to control settings from the phone, we'll be able to add more settings without overwhelming the UI on the watch

PR #529, opened 2025-11-29
settings/settings_quiet_time: Add "Notifications" toggle option to Quiet Time settings

asyba:
Add a new "Notifications" option to the Quiet Time settings menu that allows users to control whether notifications are displayed during DND.

When "Notifications" is set to "Hide":
- All notifications are suppressed during Quiet Time.
- Analytics are still recorded for incoming notifications.
- The notification window is not displayed.

When "Notifications" is set to "Show" (default):
- Notifications behave normally during Quiet Time.
- Existing interruption mask settings still apply.

This provides users with finer control over their Quiet Time experience, allowing them to completely silence notifications while still allowing important interruptions (calls, alarms) based on the interruption mask.

For example: I don't want on QT mode to display notifications while I'm on the cinema, gym, dinner with friends, etc.

CleanShot 2025-11-29 at 17 41 33@2x

Discord conversations: https://discord.com/channels/221364737269694464/1444175656500854925

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_quiet_time.c
+++ b/src/fw/apps/system_apps/settings/settings_quiet_time.c
@@ -54,6 +54,7 @@ enum QuietTimeItem {
   QuietTimeItemWeekdayScheduled,
   QuietTimeItemWeekendScheduled,
   QuietTimeItemInterruptions,
+  QuietTimeItemNotifications,
   QuietTimeItem_Count,
 };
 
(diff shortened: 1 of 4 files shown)


jplexer (changes requested):
looks good to me but please ammend commit title to make sure that the compliance check doesnt fail (for reference you can look at other commits in the repo!)

ericmigi:
I'm not super excited about this one - I foresee bug reports about people not getting notifications and don't know why. We already get enough about quiet time.

I think this thing might be best hidden behind some sort of advanced settings mode (controlled from mobile app) in future.

Maybe a better setting (I'm guessing your goal is so notifications don't stick on screen) would be an option to disable that?

asyba:
> looks good to me but please amend commit title to make sure that the compliance check doesnt fail (for reference you can look at other commits in the repo!)

sorry I forgot that in this PR. done

src/fw/apps/system_apps/settings/settings_quiet_time.c
```diff
@@ -79,6 +93,36 @@ static const char *prv_get_dnd_mask_subtitle(void *i18n_key) {
...
+}
+
+static const char *prv_get_dnd_notifications_enable(void *i18n_key) {
+  switch (alerts_preferences_dnd_get_show_notifications()) {
+    case DndNotificationModeShow:
+      return i18n_get("Show", i18n_key);
+    case DndNotificationModeHide:
+      return i18n_get("Hide", i18n_key);
+    default:
+      return "???";
```
gmarull, line 122:
```suggestion
      WTF;
```

gmarull:
> I actually like this one considering that the default option is unchanged behavior

Same, I actually would like to have this

src/fw/apps/system_apps/settings/settings_quiet_time.c
asyba: I did the same way as the method _prv_cycle_dnd_mask_

gmarull:
We could create a new “experimental” settings section, with no guarantees they’ll stay.

ericmigi:
yeah I like that idea

PR #533, opened 2025-11-29
fw/drivers/vibe/aw86225: Disable stop mode while vibrator is on

jplexer:


gmarull (commented):
why is this needed?

jplexer:
There is a bug when the watch gets out of stop mode. The first vibration is longer than it should be. This fixes that

gmarull (changes requested):
We need to understand better why disabling stop mode resolves a problem here. Stop mode is just used for cases where the CPU can't enter some deep sleep modes while a peripheral is doing something (as state would be lost, clock stopped, etc.). The LRA driver is an external component.

gmarull:
@jplexer according to Sifli, `HAL_HPAON_DISABLE_PAD` is not needed in fact. Could you try that first?

jplexer:
Sure I'll try

PR #537, opened 2025-11-30
drivers/sf32lb52: Eliminate redundant display framebuffer to save 44KB RAM

jplexer:
The sf32lb52 display driver maintained its own 45,600-byte static framebuffer (`s_framebuffer[DISPLAY_FRAMEBUFFER_BYTES]`) separate from the compositor's framebuffer. This was used to convert pixels from the compositor's ARGB 2-2-2 format to the display's RGB 3-3-2 format.

With the compositor already having its own framebuffer, this resulted in **~88KB total** being used for display buffers. This caused OOM failures when initializing the Speex voice encoder.

Instead of maintaining a separate buffer and copying with conversion, we now:

1. **Convert in-place**: During `display_update()`, convert pixels directly in the compositor's framebuffer from 222→332 format
2. **Restore after DMA**: After the LCDC finishes transferring data, convert the pixels back from 332→222 format
3. **Temporary allocation for boot**: `display_clear()` and `display_show_splash_screen()` (only called during early boot) use temporary heap allocations which are freed immediately after use

### DIFF
--- a/src/fw/drivers/sf32lb52/display_jdi.c
+++ b/src/fw/drivers/sf32lb52/display_jdi.c
@@ -8,6 +8,7 @@
 #include "drivers/display/display.h"
 #include "drivers/gpio.h"
 #include "kernel/events.h"
+#include "kernel/pbl_malloc.h"
 #include "kernel/util/sleep.h"
 #include "kernel/util/stop.h"
 #include "os/mutex.h"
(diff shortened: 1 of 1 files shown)


src/fw/drivers/sf32lb52/display_jdi.c
```diff
@@ -257,26 +281,37 @@ bool display_update_in_progress(void) {
...
-  uint16_t rows = 0U;
-  uint16_t y0 = 0U;
-  bool y0_set = false;
+  bool first_row = true;
 
   PBL_ASSERTN(!s_updating);
 
-  // convert all rows requiring an update to 332 format
+  // Convert rows in-place from 222 to 332 format
+  // We use the compositor's framebuffer directly to save 44KB RAM
```
gmarull:
```suggestion
  // We use the compositor's framebuffer directly to save RAM
```

PR #538, opened 2025-11-30
Low power opt

cameled:
Progress:
1. vddio drop to 200uA, (original 300~400uA) [11/30]
2. After remove uart connection, vddio drop 9uA(from 200uA) [11/30]
3. Just remove uart connection is ok, vddio drop to 19uA.[12/01]

After add Display back, low power current is about 350uA. [12/10]


cameled:
> My measurements (DVT, battery terminals, `./waf configure --board obelix_dvt --release`, `./waf build`, airplane mode on (no BLE))
> image

I just turn off the psram, please help to check the new one. low power current have a significant change.

PR #568, opened 2025-12-05
fw/drivers/sf32lb52/rtc: fix calibration applying bogus corrections [FIRM-907]

jplexer:
The RC10K RTC calibration algorithm uses static variables to track reference times and accumulated drift. When rtc_set_time() was called externally (e.g., phone time sync), these variables weren't reset, causing the algorithm to compute massive corrections based on stale reference times.

This resulted in the clock jumping forward or backward by 1-2 days when calibration ran after an external time sync.

Fix by:
- Reset calibration state when rtc_set_time() is called externally
- Add sanity check to reject corrections exceeding 60 seconds
- Split rtc_set_time() so calibration can apply small corrections without triggering a state reset


gmarull (approved):
good finding!

PR #573, opened 2025-12-06
Switch lptim_systick to use overflow-based wakeup during stop mode [FIRM-897]

jplexer:
Fixes the notification issue

### DIFF
--- a/src/fw/freertos_application.c
+++ b/src/fw/freertos_application.c
@@ -257,10 +257,23 @@ extern void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime ) {
         counter_stop += 0x10000;
       }
       uint32_t counter_elapsed = counter_stop - counter_start;
-      uint32_t cycles_elapsed = (counter_elapsed * RTC_TICKS_HZ) / 9000;
 
-      s_analytics_device_sleep_cpu_cycles += cycles_elapsed;
-      s_analytics_app_sleep_cpu_cycles += cycles_elapsed;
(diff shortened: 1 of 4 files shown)


rabbitsaviola (approved):

src/fw/freertos_application.c
```diff
@@ -257,10 +257,22 @@ extern void vPortSuppressTicksAndSleep( TickType_t xExpectedIdleTime ) {
         counter_stop += 0x10000;
       }
       uint32_t counter_elapsed = counter_stop - counter_start;
-      uint32_t cycles_elapsed = (counter_elapsed * RTC_TICKS_HZ) / 9000;
 
-      s_analytics_device_sleep_cpu_cycles += cycles_elapsed;
-      s_analytics_app_sleep_cpu_cycles += cycles_elapsed;
+      // Calculate elapsed ticks and step FreeRTOS tick count
+      // LPTIM runs at ~9000 Hz, RTC_TICKS_HZ is 8 Hz
+      uint32_t ticks_elapsed = (counter_elapsed * RTC_TICKS_HZ) / 9000;
```
gmarull: the 9000hz is not true (far from true in some devices), we now have the real frequency measured agasinst HXT48, maybe add an internal API to retrieve it.

gmarull (changes requested):

src/fw/freertos_application.c
jplexer: amended to use measued frequency

gmarull (approved):

PR #578, opened 2025-12-09
sdk: allow BDF fonts to be added as resources

soxfox42:
This makes it easier to add custom bitmap fonts to watchapps. The font conversion in the SDK uses freetype, which already supports BDF bitmap fonts -- it is in fact already possible to use a BDF font by renaming it to *.ttf.

Tested with Terminus 16px and the sample app:
image


jplexer (approved):
Awesome change!

PR #587, opened 2025-12-11
fw/apps/settings: Add new Health menu with Distance Unit setting toggler

MrXANA91:
As there is currently no way to change this setting for the moment, this PR adds a Distance Unit setting toggler linked to the already existing Units Distance preference, inside a new Health menu.

We can now easily toggle between Kilometers and Miles.

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_health.c
+++ b/src/fw/apps/system_apps/settings/settings_health.c
@@ -1,0 +1,111 @@
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "settings_health.h"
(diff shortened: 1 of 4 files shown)


gmarull:
https://github.com/coredevices/mobileapp/pull/54

MrXANA91:
It might be relevant to have this setting both on the mobileapp and on the watch.

Or, at least, having access to this setting on the watch while the mobileapp health integration is being built.

ericmigi (approved):
We can move this to the phone side when we have Health fully integrated. Thanks for your contribution!

PR #597, opened 2025-12-15
Fix watchface metrics event queue overflow and time jump issues

jplexer:


src/fw/shell/normal/watchface_metrics.c
```diff
@@ -37,6 +42,20 @@ static bool s_initialized = false;
...
+static uint32_t prv_get_elapsed_secs(uint32_t start_ticks) {
+  uint32_t now_ticks = rtc_get_ticks();
+  uint32_t elapsed_ticks = now_ticks - start_ticks;
+  uint32_t elapsed_secs = elapsed_ticks / RTC_TICKS_HZ;
+
+  // Cap to guard against RTC time jumps (e.g., when time syncs from phone)
+  if (elapsed_secs > WATCHFACE_METRICS_MAX_ELAPSED_SECS) {
+    elapsed_secs = WATCHFACE_METRICS_MAX_ELAPSED_SECS;
+  }
+  return elapsed_secs;
```
gmarull, line 56: we should really have a monotonic clock source, using RTC is a really bad idea.

gmarull (approved):

PR #619, opened 2025-12-18
Backlight changes

jplexer:


src/fw/drivers/led_controller/aw2016.c
```diff
@@ -17,7 +17,7 @@
 #define AW2016_REG_GCR1_CHIPEN_DIS 0U
 
 #define AW2016_REG_GCR2 0x04U
-#define AW2016_REG_GCR2_IMAX_5MA 2U
+#define AW2016_REG_GCR2_IMAX_15MA 0U
```
gmarull, line 20: @Jiyun-Yang does this place RGB LED within spec?

gmarull: It will be tested later, let's merge for now.

PR #627, opened 2025-12-20
fw/console/dbgserial: stub out on nRF52840 release builds [FIRM-1061]

jplexer:
uart_init_tx_only() is unimplemented on nRF5 and causes an immediate crash on release builds. Stub out all dbgserial functions on nRF52840 release builds to avoid the crash.


gmarull (commented):
Can it be implemented instead?

PR #636, opened 2025-12-23
Don't reset vibration timer if vibration didn't occur

FlynnD273:
When you get a notification, previously it would reset the vibration cooldown even if no vibration happened. This PR changes that so the vibration cooldown timer only gets reset if the notification vibrates the watch.


jplexer (approved):
fix commit title (too long), otherwise LGTM

PR #637, opened 2025-12-23
MenuLayer: scroll wrap around & vibes

### LINKED ISSUE #414: Improve main menu scrolling

### Proposed feature request

A copy from the https://github.com/pebble-dev/pebble-firmware/issues/284:

Currently, main menu scrolling is very basic, it is a single basic scroller from the top to the bottom, which makes reaching apps needlessly slow.

Possible improvements:

* Main menu scrolling could wrap around e.g. when you reach the top and want to go up, you should reach the bottom and vice versa.
* Start in the middle of the menu, not at the top. That way user can quickly reach apps above and below the middle (instead of just below the top)
* First option, but if you are holding down the button, the wrap around does not happen. That way, user can still hold the button to get to the bottom.

### Alternatives

_No response_

### Anything else?

_No response_

MrXANA91:
This PR adds a scroll wrap around capability to the `MenuLayer` ui object, alongside some vibe functionalities.

https://github.com/user-attachments/assets/89f6ddad-ab90-4c5d-8b58-78cda4a9a180

### Scroll wrap-around
Pressing the 'up' button when the first element of the `MenuLayer` is selected will wrap around and bring the cursor to the last element.
Pressing the 'down' button when the last element of the `MenuLayer` is selected will wrap around and bring the cursor to the first element.
However, wrap around will not be applied when holding down the 'up' or 'down' button, and cursor will still lock on the first or last element.

This wrap-around functionality is disabled by default but can be enabled with the `menu_layer_set_scroll_wrap_around` function.

### Vibe on cursor blocked / vibe on wrap-around
A small 50ms vibe can be activated on wrap-around, or when blocked. Those two behaviors are mutually exclusive (setting one to true will set the other to false).

Those are also disabled by default, but can be enabled using the `menu_layer_set_vibe_on_wrap` and `menu_layer_set_vibe_on_blocked` functions.

## Menu prefs
This PR also implements some new prefs to enable those new capabilties on every `MenuLayer` used accross the firmware system apps. No UI to change those prefs have been implemented in this PR though.
<details>
<summary>Here is a list of the apps affected</summary>

- notifications_app
- watchfaces
- alarm_editor
- alarms
- health_detail_card
- launcher_menu_layer
- send_text
- settings_system
- settings_window
- settings
- workout_selection
</details>

## Known bug(s)
- ~~Scroll wrap-around: if holding down the 'up' button when cursor is at the 2nd element, wrap around will still be applied. Same when holding down the 'down' button when cursor is at the penultimate element.~~ **<-- No longer observed with last build**

## Related issue(s)
Fix #414

### DIFF
--- a/tests/fw/apps/system_apps/health/test_health_app_includes.h
+++ b/tests/fw/apps/system_apps/health/test_health_app_includes.h
@@ -42,6 +42,7 @@
 #include "stubs_status_bar_layer.h"
 #include "stubs_syscalls.h"
 #include "stubs_task_watchdog.h"
+#include "stubs_vibes.h"
 #include "stubs_window_manager.h"
 #include "stubs_window_stack.h"
 
--- a/tests/fw/apps/system_apps/timeline/test_timeline_app_includes.h
+++ b/tests/fw/apps/system_apps/timeline/test_timeline_app_includes.h
@@ -75,6 +75,7 @@
 #include "stubs_timeline_layout_animations.h"
 #include "stubs_timeline_peek.h"
 #include "stubs_timezone_database.h"
+#include "stubs_vibes.h"
 #include "stubs_wakeup.h"
 #include "stubs_watchface.h"
 #include "stubs_window_manager.h"
--- a/tests/fw/services/timeline/test_timeline_layouts.c
+++ b/tests/fw/services/timeline/test_timeline_layouts.c
@@ -87,6 +87,7 @@ void clock_get_since_time(char *buffer, int buf_size, time_t timestamp) {
 #include "stubs_timeline_item.h"
 #include "stubs_timeline_layer.h"
 #include "stubs_timeline_peek.h"
+#include "stubs_vibes.h"
 #include "stubs_window_manager.h"
 #include "stubs_window_stack.h"
 
--- a/tests/fw/ui/test_menu_layer.c
+++ b/tests/fw/ui/test_menu_layer.c
@@ -10,6 +10,7 @@
 // Stubs
 /////////////////////
 #include "stubs_app_state.h"
+#include "stubs_click.h"
 #include "stubs_graphics.h"
 #include "stubs_heap.h"
 #include "stubs_logging.h"
@@ -19,6 +20,7 @@
 #include "stubs_ui_window.h"
 #include "stubs_process_manager.h"
 #include "stubs_unobstructed_area.h"
+#include "stubs_vibes.h"
 
 
 // Fakes
--- a/tests/fw/ui/test_option_menu_window.c
+++ b/tests/fw/ui/test_option_menu_window.c
@@ -45,6 +45,7 @@
 #include "stubs_syscalls.h"
 #include "stubs_task_watchdog.h"
 #include "stubs_unobstructed_area.h"
+#include "stubs_vibes.h"
 #include "stubs_window_manager.h"
 #include "stubs_window_stack.h"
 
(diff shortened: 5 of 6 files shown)


src/fw/applib/ui/menu_layer.c
```diff
@@ -598,6 +641,7 @@ static void prv_set_center_focused(MenuLayer *menu_layer, bool center_focused) {
 void menu_layer_init(MenuLayer *menu_layer, const GRect *frame) {
   *menu_layer = (MenuLayer) {
     .pad_bottom = true,
+    .scroll_wrap_around = true,
```
jplexer: don't enable by default - this will break behavior with existing apps

jplexer (changes requested):
I have one note, otherwise I am fine with this change though, well implemented!

src/fw/applib/ui/menu_layer.c
MrXANA91: Can I still enable it for all menus and submenus of the firmware?
Like launcher, settings, watchface selection, etc

jplexer: I would like that! final say would be from @ericmigi though

MrXANA91: Or is there a way to disable it for user apps, but still keeping it enabled by default for firmware apps?

jplexer: not that I know of

MrXANA91: I just force pushed, disabled the option by default and enabled it in all the firmware apps using specifically `MenuLayer`. Other firmware apps using layer objects inheriting from `MenuLayer` won't enable the scroll wrap around function.
<details>
<summary>Here is a list of the apps affected by this change</summary>

- notifications_app
- watchfaces
- alarm_editor
- alarms
- health_detail_card
- launcher_menu_layer
- send_text
- settings_system
- settings_window
- settings
- workout_selection
</details>

jplexer (approved):
LGTM, but requesting review from Eric because it changes the menu behavior

ericmigi:
thanks so much for testing it so much and gathering feedback! I know it takes more work but it's really appreciated.<br><br>I think it would be best to keep it in draft until we can implement a system in mobile app for settings/prefs so we don't keep adding things to the watch side.

jplexer (changes requested):
Any reason why you created menu_preferences.c and didn't just include it in prefs.c?
Including it in prefs.c would allow us to really really easily integrate it into mobile settings once we ship that!!

MrXANA91:
> Any reason why you created menu_preferences.c and didn't just include it in prefs.c? Including it in prefs.c would allow us to really really easily integrate it into mobile settings once we ship that!!

I got scared by this comment in `prefs.h:13`, especially the last line:
```c
// Shell Preferences
//
// These are preferences which must be available for querying across all shells
// and which must be implemented differently depending on the shell compiled in.
// Only preferences which are used by common services and kernel code meet these
// criteria.
//
// NEW PREFERENCES DO __NOT__ BELONG HERE WITHOUT A VERY GOOD REASON.
```
I felt like the new prefs I wanted to implement didn't really belong here. If you tell me that it would be better for everyone for those prefs to be here, I can do that.

MrXANA91:
> looks good! please add the prefs to the settings_blob_db whitelist :)

Done!

jplexer:
Yeah the main branch is missing #830

PR #652, opened 2026-01-01
docs, tools: update development documentation

asyba:
Update debugging documentation.

### DIFF
--- a/docs/development/debug.md
+++ b/docs/development/debug.md
@@ -1,0 +1,59 @@
+# Debugging
+
+#### Prerequisites
+
+```bash
+source .venv/bin/activate
+pip install pebble-tool
+```
+
+**Explanation:**
+- `pbl` = **internal** tool in this project (located in `python_libs/pbl/`)
+- `pebble-tool` = **public** package from PyPI (https://github.com/coredevices/pebble-tool)
+- `pbl` requires `pebble-tool` to work (see `python_libs/pbl/pbl/__init__.py` line 3)
+- Only install `pebble-tool`, `pbl` is already in the project
+
+```{tip}
+Remember to have Pebble core mobile app with LAN developer connection enabled.
+```
+
+## Logs
+
+#### Usage
+
+```bash
+pebble logs --phone 192.168.1.100
+```
+
+## Getting Coredumps
+
+#### Usage
+
+```bash
+pbl coredump --phone 192.168.1.100
+```
+
+#### Notes
+
+- Coredump is saved as `pebble_coredump_YYYY-MM-DD_HH-MM-SS.core`
+
+## Read Coredumps
+
+#### Convert a Pebble core dump to an ELF file.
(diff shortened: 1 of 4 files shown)


docs/development/debug.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash
+tools/readcore.py <coredump_file>.core core.elf
+```
+
+#### Debug with GDB
+
+```bash
+arm-none-eabi-gdb build/src/fw/tintin_fw.elf core.elf
+```
```
gmarull, line 52: `./waf debug`?

tools/analyze_coredump.py
```diff
@@ -18,7 +18,7 @@
 
 def find_gdb_executable():
     """Find an available arm-none-eabi-gdb with Python support."""
-    candidates = ["arm-none-eabi-gdb-py", "arm-none-eabi-gdb-py3"]
+    candidates = ["arm-none-eabi-gdb-py", "arm-none-eabi-gdb-py3", "arm-none-eabi-gdb"]
```
gmarull, line 21: pls make these changes into a separate commit (explaining why this change)

docs/development/debug.md
asyba: > `./waf debug`?

never use it before, but this required to be connected to the pebble by a board or something right?
My debug instructions are specific for remote, for people that doesn't have a way to connect directly to the board, so they can only get logs by IP or get a coredump.

I can update to say that is intended for remote debugging mostly.

tools/analyze_coredump.py
asyba: I forgot to add to the docs another useful command.
_tools/analyze_coredump.py build/src/fw/tintin_fw.elf core.elf -o report.txt_
but I got an error, probably because I have only installed with brew arm-none-eabi-gdb

```
(.venv)  PebbleOS % tools/analyze_coredump.py build/src/fw/tintin_fw.elf core1.elf -o report.txt

Traceback (most recent call last):
  File "/tools/analyze_coredump.py", line 240, in <module>
    main()
    ~~~~^^
  File "/tools/analyze_coredump.py", line 233, in main
    analyzer = CoredumpAnalyzer(elf_file, coredump_file, args.output)
  File "/Users/federicobechini/Documents/Droplet/GIT/PebbleOS/tools/analyze_coredump.py", line 38, in __init__
    raise RuntimeError(
    ...<2 lines>...
    )
RuntimeError: Could not find arm-none-eabi-gdb-py or arm-none-eabi-gdb-py3. Please ensure one of these is installed and in PATH.
(.venv) PebbleOS % tools/analyze_coredump.py build/src/fw/tintin_fw.elf core1.elf -o report.txt

Analyzing coredump: core1.elf
Using ELF file: build/src/fw/tintin_fw.elf
Output will be saved to: report.txt
```

PR #692, opened 2026-01-12
notifications: support iOS custom icon/vibe/color

asyba:
iOS support for custom icon/vibe/color notifications.

The custom vibe pattern is not supported for the moment, it requires more work on the mobileapp side. FW should be fine.

PR mobileapp: https://github.com/coredevices/mobileapp/pull/99

Tested on iOS only with Duo2.

### DIFF
--- a/src/fw/services/normal/notifications/ancs/ancs_item.c
+++ b/src/fw/services/normal/notifications/ancs/ancs_item.c
@@ -7,8 +7,10 @@
 
 #include "applib/graphics/utf8.h"
 #include "kernel/pbl_malloc.h"
+#include "resource/resource_storage_impl.h"
 #include "services/common/i18n/i18n.h"
 #include "system/logging.h"
+#include "services/normal/timeline/timeline_resources.h"
 #include "util/string.h"
 
 #include <stdio.h>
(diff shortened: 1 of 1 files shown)


src/fw/services/normal/notifications/ancs/ancs_item.c
```diff
@@ -246,17 +248,53 @@ static void prv_populate_attributes(TimelineItem *item,
     attr_idx++;
   }
 
-  // add the icon attribute
+  uint32_t icon_id = app_metadata->icon_id;
+  if (notif_prefs) {
+    uint32_t pref_icon = attribute_get_uint32(&notif_prefs->attr_list, AttributeIdIcon, 0);
+    if (pref_icon != 0) {
+      if (pref_icon < 512) {
```
gmarull: please avoid hardcoding magic numbers ;-)

asyba: done

src/fw/services/normal/notifications/ancs/ancs_item.c
```diff
@@ -18,6 +19,10 @@
 #define MULTIMEDIA_INDICATOR_LENGTH 64
 #define MULTIMEDIA_EMOJI "🎁"
 
+//! Maximum resource ID value that is considered a system resource.
+//! Resource IDs below this threshold need SYSTEM_RESOURCE_FLAG set.
+#define MAX_SYSTEM_RESOURCE_ID 512
```
gmarull: how is that related to e.g.

```
#define MAX_RESOURCES_PER_STORE 256
#define MAX_RESOURCES_FOR_SYSTEM_STORE 768
```
?

asyba: done added import from resource_storage_impl

PR #693, opened 2026-01-12
fw/drivers/sf32lb52: add boot animation [FIRM-1154]

jplexer:
Add an animated boot splash screen showing the Pebble logo with a rotating spinner below it. The animation runs in a background FreeRTOS task until the compositor is ready to display content.

- 8-dot spinner with one highlighted dot rotating clockwise
- Uses absolute timing (vTaskDelayUntil) for consistent frame rate
- High task priority ensures smooth animation during boot
- Automatically stops when display_update() is first called

### DIFF
--- a/src/fw/drivers/display/display.h
+++ b/src/fw/drivers/display/display.h
@@ -17,8 +17,11 @@ typedef struct {
 typedef bool(*NextRowCallback)(DisplayRow* row);
 typedef void(*UpdateCompleteCallback)(void);
 
-//! Show the splash screen before the display has been fully initialized.
-void display_show_splash_screen(void);
+//! Update the display with a boot animation frame.
+//! This is a simple interface for the boot animation service to send frames
+//! to the display before the compositor is initialized.
+//! @param framebuffer Pointer to the framebuffer data (DISPLAY_FRAMEBUFFER_BYTES)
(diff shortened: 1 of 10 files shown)


gmarull (changes requested):
Display driver should not have this logic, this is driver agnostic

gmarull (commented):
can we combine splash screen functionality into a single feature, called boot splash that can be externally controlled by the service?

src/fw/drivers/display/display.h
```diff
@@ -20,6 +20,12 @@ typedef void(*UpdateCompleteCallback)(void);
 //! Show the splash screen before the display has been fully initialized.
 void display_show_splash_screen(void);
```
gmarull: my previous comment meant to remove that and consolidate to a single splash write call, so it can be updated anytime from an external service.

gmarull: so, this should be removed

jplexer:
FPGA display route tested using emulators, rest confirmed on real hw

gmarull (approved):
lgtm, I'd probably center logo+spinner vertically, or use an "infinite progressbar" which is thinner and makes centering less critical visually speaking.

gmarull (approved):
like the bar!

PR #694, opened 2026-01-12
platform: add spalding_gabbro variant and fix snowy_emery platform type

jplexer:
Add spalding_gabbro, a new board variant combining Spalding hardware (STM32F4) with Getafix's 260x260 round display. This is analogous to how snowy_emery combines Snowy hardware with Robert's display.

Key changes:
- Add PLATFORM_SPALDING_GABBRO and PLATFORM_SNOWY_EMERY defines for display-specific code overrides while maintaining base platform compatibility
- Fix platform.h so snowy_emery correctly uses PlatformTypeEmery and spalding_gabbro uses PlatformTypeGabbro
- Use Getafix's display row infos and framebuffer handling for spalding_gabbro
- Exclude display calibration (not needed for this display)
- Configure 512K RAM to match gabbro platform requirements


src/fw/applib/graphics/framebuffer.c
```diff
@@ -19,10 +19,10 @@ void framebuffer_init(FrameBuffer *fb, const GSize *size) {
 
 GBitmap framebuffer_get_as_bitmap(FrameBuffer *fb, const GSize *size) {
   PBL_ASSERTN(!gsize_equal(size, &GSizeZero));
-#if PLATFORM_SPALDING
```
gmarull, line 22: isn't this redundant? Or is spalding defined as well?

src/fw/shell/normal/prefs.c
```diff
@@ -1546,7 +1546,7 @@ void shell_prefs_set_coredump_on_request(bool enabled) {
   prv_pref_set(PREF_KEY_COREDUMP_ON_REQUEST, &enabled, sizeof(enabled));
 }
 
-#if PLATFORM_OBELIX || PLATFORM_GETAFIX
+#if PLATFORM_OBELIX || PLATFORM_GETAFIX || PLATFORM_SPALDING_GABBRO
```
gmarull: Isn't snowy emery missing then?

src/fw/applib/graphics/framebuffer.c
jplexer: Spalding is defined as well

src/fw/shell/normal/prefs.c
jplexer: look, yes it would be, but it compiled without it so I missed it LMAO, adding it in a moment

gmarull (approved):

PR #695, opened 2026-01-12
fw/drivers/led_controller/aw2016: add reset detection and recovery [FIRM-1007]

jplexer:
The AW2016 can spontaneously reset (PUIS=1 in ISR) due to VBAT dipping below 1.25V or SCL held low >130ms. When this happens, all registers reset to defaults and the backlight turns off while firmware state thinks it's still on.

Add periodic health checking while backlight is on:
- Poll ISR register every 100ms to detect PUIS (power-up reset)
- If reset detected, re-enable chip and reconfigure registers
- Log warning when recovery occurs for debugging
- Timer only runs while backlight is active (no overhead when off)

Also check for reset before/after brightness changes to catch resets that could occur during configuration.

### DIFF
--- a/src/fw/drivers/led_controller/aw2016.c
+++ b/src/fw/drivers/led_controller/aw2016.c
@@ -5,7 +5,11 @@
 
 #include "board/board.h"
 #include "drivers/i2c.h"
+#include "services/common/new_timer/new_timer.h"
+#include "system/logging.h"
 #include "system/passert.h"
+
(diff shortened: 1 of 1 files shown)


gmarull (commented):
This may be a hardware problem. Why can this actually happen in the first place?

jplexer:
yep, just confirmed it also happens on a build with --nostop

jplexer:
Superseded by #702

PR #704, opened 2026-01-14
fw/apps/system/settings: tweak ISED/FCC ids

gmarull:


gmarull:
merging, needed for cert

PR #707, opened 2026-01-14
src/fw/shell/normal: new combo back + up quick launch

asyba:
Added a new Quick Launch button using a Hold with BACK+UP.

### DIFF
--- a/src/fw/shell/normal/prefs.c
+++ b/src/fw/shell/normal/prefs.c
@@ -143,6 +143,7 @@ static QuickLaunchPreference s_quick_launch_back = {
 
 #define PREF_KEY_QUICK_LAUNCH_SINGLE_CLICK_UP "qlSingleClickUp"
 #define PREF_KEY_QUICK_LAUNCH_SINGLE_CLICK_DOWN "qlSingleClickDown"
+#define PREF_KEY_QUICK_LAUNCH_COMBO_BACK_UP "qlComboBackUp"
 
 static QuickLaunchPreference s_quick_launch_single_click_up = {
   .enabled = true,
@@ -152,6 +153,11 @@ static QuickLaunchPreference s_quick_launch_single_click_up = {
 static QuickLaunchPreference s_quick_launch_single_click_down = {
   .enabled = true,
   .uuid = TIMELINE_UUID_INIT,
+};
+
+static QuickLaunchPreference s_quick_launch_combo_back_up = {
+  .enabled = false,
+  .uuid = UUID_INVALID_INIT,
 };
 
 #define PREF_KEY_QUICK_LAUNCH_SETUP_OPENED "qlSetupOpened"
@@ -424,6 +430,12 @@ static bool prv_set_s_quick_launch_single_click_up(QuickLaunchPreference *pref) 
 static bool prv_set_s_quick_launch_single_click_down(QuickLaunchPreference *pref) {
   prv_normalize_quick_launch_pref(pref);
   s_quick_launch_single_click_down = *pref;
+  return true;
+}
+
+static bool prv_set_s_quick_launch_combo_back_up(QuickLaunchPreference *pref) {
+  prv_normalize_quick_launch_pref(pref);
+  s_quick_launch_combo_back_up = *pref;
   return true;
 }
 
@@ -1316,6 +1328,28 @@ void quick_launch_single_click_set_enabled(ButtonId button, bool enabled) {
   }
   pref.enabled = enabled;
   prv_pref_set(key, &pref, sizeof(pref));
+}
+
+bool quick_launch_combo_back_up_is_enabled(void) {
+  return s_quick_launch_combo_back_up.enabled;
+}
+
+AppInstallId quick_launch_combo_back_up_get_app(void) {
+  return app_install_get_id_for_uuid(&s_quick_launch_combo_back_up.uuid);
+}
+
+void quick_launch_combo_back_up_set_app(AppInstallId app_id) {
+  QuickLaunchPreference pref = (QuickLaunchPreference) {
+    .enabled = true,
+  };
+  app_install_get_uuid_for_install_id(app_id, &pref.uuid);
+  prv_pref_set(PREF_KEY_QUICK_LAUNCH_COMBO_BACK_UP, &pref, sizeof(pref));
+}
+
+void quick_launch_combo_back_up_set_enabled(bool enabled) {
+  QuickLaunchPreference pref = s_quick_launch_combo_back_up;
+  pref.enabled = enabled;
+  prv_pref_set(PREF_KEY_QUICK_LAUNCH_COMBO_BACK_UP, &pref, sizeof(pref));
 }
 
 void watchface_set_default_install_id(AppInstallId app_id) {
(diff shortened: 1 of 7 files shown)


src/fw/apps/system_apps/settings/settings_quick_launch_app_menu.c
```diff
@@ -172,3 +181,39 @@ void quick_launch_app_menu_window_push(ButtonId button, bool is_tap) {
   const bool animated = true;
   app_window_stack_push(&option_menu->window, animated);
 }
+
+void quick_launch_app_menu_window_push_combo(void) {
+  QuickLaunchAppMenuData *data = app_zalloc_check(sizeof(*data));
+  data->is_tap = true;
```
jplexer: Setting is_tap = true for a hold-combo is confusing. Consider adding a comment explaining why, or restructuring the filter logic to be clearer.

src/fw/shell/normal/watchface.c
```diff
@@ -44,23 +51,93 @@ static void prv_launch_app_via_button(AppLaunchEventConfig *config,
   app_manager_put_launch_app_event(config);
 }
 
+static void prv_reset_button_state(ButtonId button_id) {
+  // Reset button state in bitset
+  s_buttons_pressed &= ~(BIT_SET << button_id);
+  // Reset recognizer state to prevent stale state
+  ClickRecognizer *recognizer = &s_click_manager.recognizers[button_id];
+  recognizer->is_button_down = false;
+  recognizer->number_of_clicks_counted = 0;
```
jplexer: Directly manipulating ClickRecognizer internal state is fragile. If button up events arrive after this reset, it could cause state corruption.

asyba: @jplexer I was having problems where, after successfully executing the combo, the quick launches would generally be messed up. This helped me reset them and fix them.

I just try instead call click_manager_reset(&s_click_manager);, I did a quick test and it worked fine.

I also tried removing both and it works fine... maybe none reset is necessary.
Although I do notice the animations are slower without any kind of reset, I don't know if that's just my imagination from so much testing, haha.

src/fw/apps/system_apps/settings/settings_quick_launch_app_menu.c
asyba: cleanup

src/fw/shell/normal/watchface.c
asyba: removed

src/fw/shell/normal/prefs.c
```diff
@@ -1318,6 +1273,28 @@ void quick_launch_single_click_set_enabled(ButtonId button, bool enabled) {
...
+  return s_quick_launch_combo_back_up.enabled;
+}
+
+AppInstallId quick_launch_combo_back_up_get_app(void) {
+  return app_install_get_id_for_uuid(&s_quick_launch_combo_back_up.uuid);
+}
+
+void quick_launch_combo_back_up_set_app(AppInstallId app_id) {
+  QuickLaunchPreference pref = (QuickLaunchPreference) {
+    .enabled = true,
```
jplexer, line 1343: This unconditionally sets .enabled = true, so calling set_app(INSTALL_ID_INVALID) stores an enabled pref with an invalid UUID. The settings UI handles this correctly by calling set_enabled(false) separately for the "None" case, but the asymmetry with the single-click variant (which just stores the UUID) could bite if set_app is ever called in isolation.

src/fw/shell/normal/watchface.c
```diff
@@ -44,23 +51,83 @@ static void prv_launch_app_via_button(AppLaunchEventConfig *config,
...
-  if (!quick_launch_single_click_is_enabled(click_recognizer_get_button_id(recognizer))) return;
+  ButtonId button = click_recognizer_get_button_id(recognizer);
+  
+  if ((s_buttons_pressed & COMBO_BACK_UP_BUTTONS) == COMBO_BACK_UP_BUTTONS) {
+    return;
+  }
+  
+  if (!quick_launch_single_click_is_enabled(button)) return;
   //check if quick launch app is not timeline
   if (quick_launch_single_click_get_app(click_recognizer_get_button_id(recognizer)) != APP_ID_TIMELINE) {
```
jplexer: nit: button is now extracted on line 124 but the function body still calls click_recognizer_get_button_id(recognizer) three more times instead of using it.

src/fw/shell/normal/watchface.c
```diff
@@ -44,23 +51,83 @@ static void prv_launch_app_via_button(AppLaunchEventConfig *config,
...
+  bool up_pressed = (s_buttons_pressed & (BIT_SET << BUTTON_ID_UP)) != BIT_CLEAR;
+  bool both_pressed = back_pressed && up_pressed;
+  
+  if (both_pressed) {
+    if (s_combo_back_up_timer == NULL) {
+      // Cancel individual button timers to prevent them from firing
+      // This ensures only the combo executes, not individual hold handlers
+      if (s_click_manager.recognizers[BUTTON_ID_BACK].hold_timer != NULL) {
+        app_timer_cancel(s_click_manager.recognizers[BUTTON_ID_BACK].hold_timer);
+        s_click_manager.recognizers[BUTTON_ID_BACK].hold_timer = NULL;
```
jplexer: Reaching directly into s_click_manager.recognizers[].hold_timer is fragile, it bypasses the recognizer's own state machine and could leave fields like is_held stale. There's no per-recognizer cancel API (only click_manager_reset() which resets all buttons), so this is the only option short of adding one. Might be worth adding a click_recognizer_reset() to click_internal.h and using it here.

src/fw/shell/normal/prefs.c
asyba: But this will not also happen with _quick_launch_single_click__?

src/fw/shell/normal/watchface.c
asyba: done

src/fw/shell/normal/watchface.c
asyba: done, created

src/fw/apps/system/settings/quick_launch.c
```diff
@@ -54,6 +55,8 @@ static const char *s_row_titles[NUM_ROWS] = {
   [ROW_HOLD_DOWN]    = i18n_noop("Hold Down"),
   /// Shown in Quick Launch Settings as the title of the hold back button quick launch option.
   [ROW_HOLD_BACK]    = i18n_noop("Hold Back"),
+  /// Shown in Quick Launch Settings as the title of the hold back+up button combination option.
+  [ROW_HOLD_BACK_UP] = i18n_noop("Hold Back + Up"),
```
jplexer: fyi, we'd only merge this as a phone-side setting (meaning this menu entry would be removed and the corresponding pref would be added to the settings whitelist at src/fw/services/normal/blob_db/settings_blob_db.c)

asyba: got it, Is there a set method for deciding this, or is it done on a case-by-case basis? 
For example, will each new configuration be automatically added to mobile-only, or will some other existing old configurations also be moved to mobile-only?

jplexer: currently its new configs will be mobile-only but it can be decided on a case by case basis :)

asyba:
@jplexer when this will get merge or waiting more approvals?
I don't want to keep fixing conflicts every week, if for the moment it will not be merge, better to resolve them all before merge it.

asyba:
Mobile setting side: https://github.com/coredevices/mobileapp/pull/125

PR #708, opened 2026-01-15
Timeline: Add the full Timeline app in launcher

### LINKED ISSUE #456: Add Timeline / Timeline Past to the menu when unmapped from UP/DOWN

### Proposed feature request

#448 made Health available from the menu when not mapped to any of the Quick Launch buttons.

But this problem still remains for the Timeline and the Timeline Past - if they are unmapped, they are not accessible. This poses an issue for users that would like to map UP/DOWN buttons to some other more frequently used application, but would still like to occasionally access the timelines.

### Alternatives

_No response_

### Anything else?

_No response_

MrXANA91:
This PR adds the full Timeline app in the launcher.

## Change
The application just starts as Timeline Future, but instead of going back to the watchface when pressing up, it switches to the Timeline Past app and instead of going back to the watchface when pressing back, it goes back to the launcher menu, as any other app would.
You can of course go back and forth between Past and Future.

A new "Timeline" is available in the quicklaunch menu, giving the ability to check both Past and Future from one press.

Behavior of Timeline Future and Timeline Past via the Quicklaunch remain unchanged, those changes are only applied if launching Timeline through the menu launcher or if launching "Timeline" through the Quicklaunch.

## Video

https://github.com/user-attachments/assets/768f0717-25a7-4845-a932-433213d6ddd9

## TODO
- [x] Proof of Concept
- [x] Add comments
- [x] Add Timeline launcher icon
- [x] Demo video
- [x] Solve known bug(s)

## Known bug(s)
- [x] If there is no future event for the current day, the day separator 'Today' won't be displayed when switching from Future to Past (illustrated in the video)
- [x] Same issue if no past event for the current day, the day separator 'Today' won't be displayed when switching from Past to Future
- [x] If Past or Future displays the No Event screen, it won't switch from Past to Future, or Future to Past

## Related issue(s)
Fix #456

### DIFF
--- a/src/fw/apps/system_apps/timeline/timeline.c
+++ b/src/fw/apps/system_apps/timeline/timeline.c
@@ -169,6 +169,40 @@ static void prv_launch_watchface(void *data) {
 #endif
 }
 
+// This function swaps from Timeline Future to Timeline Past when launching Timeline Full
+// To be precise: Timeline Full behaves like Timeline Future, but instead of going back to the
+// watchface when scrolling up, it will launch Timeline Past with a "force_full" argument.
+// By doing so, Timeline Past will launch the Timeline Full (being Timeline Future) when scrolling
+// down.
+static void prv_swap_timeline(void *data) {
+  const bool is_future = (s_app_data->timeline_model.direction == TimelineIterDirectionFuture);
+  const bool to_timeline = true;
+  const CompositorTransition *transition = PBL_IF_RECT_ELSE(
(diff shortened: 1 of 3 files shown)


ericmigi (approved):
Very cool feature! Nice addition to the OS

MrXANA91:
Thank you both!

I am currently working on a fix for the day separator issue (I will be able to test after the sunset in my timezone)
As for the menu app icon, I have honestly no idea what to do for the moment haha

PR #714, opened 2026-01-15
fw/drivers: recover AW2016 backlight from I2C clock stretching

jplexer:
NPM1300 ADC operations can cause I2C clock stretching >130ms, which resets the AW2016 LED controller on Obelix (shared I2C bus). Add recovery function that detects AW2016 reset and restores brightness state after NPM1300 ADC reads.


gmarull (commented):
Before adding workarounds, 

1. Can you confirm running this build backlight always works?
2. We can ask Nordic about this, maybe they can provide some ideas

PR #718, opened 2026-01-16
Fix I2C clock stretching issues on SF32LB5

jplexer:
Fix two related issues causing I2C delays and watchdog warnings during heavy App activity (e.g., launcher scrolling):

1. Optimize GPIO IRQ handlers
    - Only process pending interrupts instead of looping through all 78/36 pins
    - Prevents interrupt storms that blocked I2C from running

2. Fix display updates blocking I2C
    - Display driver does in-place pixel conversion in app task (higher priority than system task where I2C runs)
    - Now yields to system task every 32 rows during conversion
    - Gives I2C regular opportunities to run without slowing down display updates

### DIFF
--- a/src/fw/drivers/sf32lb52/display_jdi.c
+++ b/src/fw/drivers/sf32lb52/display_jdi.c
@@ -9,6 +9,7 @@
 #include "drivers/gpio.h"
 #include "kernel/events.h"
 #include "kernel/pbl_malloc.h"
+#include "kernel/pebble_tasks.h"
 #include "kernel/util/sleep.h"
 #include "kernel/util/stop.h"
 #include "os/mutex.h"
@@ -17,6 +18,7 @@
 
 #include "FreeRTOS.h"
 #include "semphr.h"
+#include "task.h"
 
 #include "bf0_hal.h"
 #include "bf0_hal_lcdc.h"
@@ -131,6 +133,10 @@ static void prv_display_update_start(void) {
 }
 
 static void prv_display_update_terminate(void *data) {
+  TaskHandle_t app_task = xTaskGetCurrentTaskHandle();
+  UBaseType_t original_priority = uxTaskPriorityGet(app_task);
+  uint16_t rows_since_yield = 0;
+
   // Convert the updated region back from 332 to 222 format
   for (uint16_t y = s_update_y0; y <= s_update_y1; y++) {
     uint8_t *row = &s_framebuffer[y * PBL_DISPLAY_WIDTH];
@@ -153,6 +159,16 @@ static void prv_display_update_terminate(void *data) {
       row32[x] = ((p >> 2) & 0x30303030) |  // R: bits 6-7 → 4-5
                  ((p >> 1) & 0x0C0C0C0C) |  // G: bits 3-4 → 2-3
                  (p & 0x03030303);          // B: bits 0-1 stay
+    }
+
+    // Periodically yield to system task to allow I2C to run
+    // Briefly lower priority, yield, then restore
+    rows_since_yield++;
+    if (rows_since_yield >= 32) {
+      vTaskPrioritySet(app_task, tskIDLE_PRIORITY + 1);
+      taskYIELD();
+      vTaskPrioritySet(app_task, original_priority);
+      rows_since_yield = 0;
     }
   }
 
@@ -270,11 +286,14 @@ bool display_update_in_progress(void) {
 }
 
 void display_update(NextRowCallback nrcb, UpdateCompleteCallback uccb) {
-  DisplayJDIState *state = DISPLAY->state;
   DisplayRow row;
   bool first_row = true;
+  uint16_t rows_since_yield = 0;
 
   PBL_ASSERTN(!s_updating);
+
+  TaskHandle_t app_task = xTaskGetCurrentTaskHandle();
+  UBaseType_t original_priority = uxTaskPriorityGet(app_task);
 
   // Convert rows in-place from 222 to 332 format
   // We use the compositor's framebuffer directly to save RAM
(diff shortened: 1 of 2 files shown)


src/fw/services/common/system_task.c
```diff
@@ -17,7 +17,7 @@
 #include "queue.h"
 #include "task.h"
 
-#define SYSTEM_TASK_PRIORITY (tskIDLE_PRIORITY + 1)
+#define SYSTEM_TASK_PRIORITY (tskIDLE_PRIORITY + 2)
```
gmarull: It feels a bit scary to do this change , TBH; I wonder why this was never needed before. What are we doing differently now? Why is the app using so much CPU for a long time without releasing it back?

src/fw/drivers/sf32lb52/exti.c
```diff
@@ -50,36 +50,56 @@ static GPIO_TypeDef *prv_gpio_get_instance(GPIO_TypeDef *hgpio, uint16_t gpio_pi
...
+  ExtiHandlerConfig_t *configs;
+  uint8_t max_num;
+
+  if (hgpio == hwp_gpio1) {
+    configs = s_exti_gpio1_handler_configs;
+    max_num = EXTI_MAX_GPIO1_PIN_NUM;
+  } else {
+    configs = s_exti_gpio2_handler_configs;
+    max_num = EXTI_MAX_GPIO2_PIN_NUM;
+  }
```
gmarull: we should never get anything else other than hwp_gpio1 on HCPU, why is this needed? on HCPU (main core), only a single GPIO controller is used for all pins.

jplexer: I wasn't aware that gpio2 wasn't being used (and as you can see the while loops before were checking if index is smaller than either gpio1 or 2 and then checked if the gpio1 handler config callback was not null)

src/fw/services/common/system_task.c
jplexer: display updates are really resource intensive in comparison to other platforms

gmarull: Can you measure where are we spending so much CPU cycles? There's an actual API that could help https://github.com/coredevices/PebbleOS/blob/main/src/fw/system/profiler.h. Sifli SoCs are used in products with much richer UIs, so we must be doing something wrong?

jplexer: Will do tomorrow 🫡

src/fw/drivers/sf32lb52/exti.c
gmarull: Gpio2 is for LCPU IIRC

src/fw/services/common/system_task.c
jplexer: profiling made it clear that the pixel conversion is what is blocking the system task

src/fw/drivers/sf32lb52/exti.c
```diff
@@ -105,30 +105,34 @@ void exti_configure_pin(ExtiConfig cfg, ExtiTrigger trigger, ExtiHandlerCallback
       gpiox->IPLSR = (1UL << offset);
       break;
   }
+
+  // Configure NVIC once during pin setup
```
gmarull, line 109: open a separate PR for GPIO changes, and drop any GPIO2 code as it's not needed/used

src/fw/drivers/sf32lb52/display_jdi.c
```diff
@@ -131,6 +133,10 @@ static void prv_display_update_start(void) {
 }
 
 static void prv_display_update_terminate(void *data) {
+  TaskHandle_t app_task = xTaskGetCurrentTaskHandle();
```
gmarull, line 136: It looks like drivers should not be dealing with system task settings (dependency inversion).

My proposal would be:

1. Best: How difficult is it to use RGB332? We're doing conversions unnecessarily just because of legacy.
2. Tweak priorities in the compositor layer as a temporary solution, but (1) is the right way

I still wonder why this is a problem for I2C transfers. UI may use a lot of CPU, but I2C IRQs will still be serviced, and so I2C transfers will finish. Under which condition/sequence of events is clock stretching happening?

jplexer: Good points! Agreed on the dependency inversion issue. The display driver shouldn't be manipulating task priorities. The problem is that the driver does the conversion in-place to save RAM, so it's the only component that knows it's blocking lower-priority tasks.

On RGB332: it's difficult and breaking. We'd lose the alpha channel going from ARGB2222 to RGB332, which breaks all existing SDK apps and watchfaces since we can't recompile them. It also requires rewriting graphics blending and converting all resources. Only worth it if we wouldn't want to keep backwards compatibility.

On why clock stretching happens: the critical detail is that I2C runs in system task context at priority idle+1, not at IRQ priority. The sequence is system task calls i2c_read and blocks waiting, then app task at priority idle+2 does 27ms of pixel conversion. FreeRTOS won't preempt the higher-priority app task, so the I2C hardware times out waiting for the CPU, causing clock stretching. The I2C interrupt does fire but it just wakes the system task which can't run due to app task priority. If I2C was pure IRQ-driven you'd be right, but the FreeRTOS driver uses task context for blocking operations.

gmarull: > Good points! Agreed on the dependency inversion issue. The display driver shouldn't be manipulating task priorities. The problem is that the driver does the conversion in-place to save RAM, so it's the only component that knows it's blocking lower-priority tasks.

Could compositor just swap priorities (with full update granularity)?

> 
> On RGB332: it's difficult and breaking. We'd lose the alpha channel going from ARGB2222 to RGB332, which breaks all existing SDK apps and watchfaces since we can't recompile them. It also requires rewriting graphics blending and converting all resources. Only worth it if we wouldn't want to keep backwards compatibility.
> 

We could use e.g. RGB232, making it compatible with RGB332 (upper bit would be for alpha), or, just use RGB332 with a color that means mask/alpha. I think it's ok to have inefficiencies to run legacy apps, but, on new apps, we should be free to do whatever we want (also because Emery is in practice a new platform, more Gabbro). I'd be in favor of using a color format that HW supports, or in general, making apps agnostic of the underlying format (!= available colors), except those manipulating the framebuffer.

> On why clock stretching happens: the critical detail is that I2C runs in system task context at priority idle+1, not at IRQ priority. The sequence is system task calls i2c_read and blocks waiting, then app task at priority idle+2 does 27ms of pixel conversion. FreeRTOS won't preempt the higher-priority app task, so the I2C hardware times out waiting for the CPU, causing clock stretching. The I2C interrupt does fire but it just wakes the system task which can't run due to app task priority. If I2C was pure IRQ-driven you'd be right, but the FreeRTOS driver uses task context for blocking operations.

As far as I read the code, I2C transfer is handled via IRQ. What IRQ does is just signal a ready flag so tasks waiting for the transfer to finish can be woken up and process the data. So, it should not cause clock stretching? Clock stretching happens if slave cannot keep up with master's clock speed.

PR #723, opened 2026-01-19
feat: add display orientation setting for obelix and getafix

rojas-claudio:
This PR introduces the 'Orientation' display setting (seen on asterix) to obelix and getafix devices. When enabled, the display will rotate 180º so that Up/Select/Down will be positioned on the left of the watch.

### DIFF
--- a/src/fw/apps/system_apps/settings/settings_display.c
+++ b/src/fw/apps/system_apps/settings/settings_display.c
@@ -88,7 +88,7 @@ static void prv_intensity_menu_push(SettingsDisplayData *data) {
       true /* icons_enabled */, s_intensity_labels, data);
 }
 
-#if PLATFORM_ASTERIX
+#if CAPABILITY_HAS_ORIENTATION_MANAGER
 // Orientation Settings
 /////////////////////////////
 static const char *s_display_orientation_labels[] = {
@@ -193,7 +193,7 @@ static void prv_legacy_app_mode_menu_push(SettingsDisplayData *data) {
 
 enum SettingsDisplayItem {
   SettingsDisplayLanguage,
-#if PLATFORM_ASTERIX
+#if CAPABILITY_HAS_ORIENTATION_MANAGER
   SettingsDisplayOrientation,
 #endif
   SettingsDisplayBacklightMode,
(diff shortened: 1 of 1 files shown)


src/fw/shell/normal/prefs.c
```diff
@@ -22,7 +22,7 @@
 #include "services/common/accel_manager.h"
 #include "services/common/hrm/hrm_manager.h"
 #include "services/common/i18n/i18n.h"
-#if PLATFORM_ASTERIX
+#if PLATFORM_ASTERIX || PLATFORM_OBELIX
```
jplexer: I think this would be better as a capability?
(i.e. CAPABILITY_CAN_ROTATE or CAPABILITY_HAS_ORIENTATION_MANAGER)

src/fw/drivers/sf32lb52/display_jdi.c
```diff
@@ -35,6 +35,8 @@ static bool s_updating;
 static UpdateCompleteCallback s_uccb;
 static SemaphoreHandle_t s_sem;
 
+static bool s_rotated_180 = false;
```
jplexer: getafix is rotated
maybe we should just replace DISPLAY_ORIENTATION_ROTATED_180 compile time checks with regular runtime checks and set the default for getafix to true

jplexer (changes requested):
over all this looks good to me but please note the comments :)

src/fw/drivers/sf32lb52/display_jdi.c
gmarull: yep, since rotation becomes a runtime setting, it doesn't make sense to keep it compile time.

src/fw/drivers/sf32lb52/display_jdi.c
```diff
@@ -35,6 +35,14 @@ static bool s_updating;
 static UpdateCompleteCallback s_uccb;
 static SemaphoreHandle_t s_sem;
 
+#if PLATFORM_OBELIX
+static bool s_rotated_180 = false;
+#elif PLATFORM_GETAFIX
+static bool s_rotated_180 = true;
+#else
+static bool s_rotated_180 = false;
+#endif
```
gmarull: just use `DISPLAY_ORIENTATION_ROTATED_180`

PR #728, opened 2026-01-19
Switch back to 30mA Backlight

jplexer:


gmarull (commented):
Please adjust PRF MFG backlight settings (color) that was recently adjusted as well

gmarull (approved):

PR #749, opened 2026-01-25
Docs (for Asterix B2Bv2 board): add first setup steps

MrXANA91:
After some troubleshooting with the Asterix B2Bv2 board (thanks @jplexer for the help), here is a PR that adds those steps into the PebbleOS docs.

### DIFF
--- a/docs/boards/asterix/index.md
+++ b/docs/boards/asterix/index.md
@@ -53,10 +53,69 @@ Do not use any adapter that does not operate at 1.8V, or there is a **risk of da
    - L: connected
    - R: disconnected
 
+### Core B2B v2 Setup
+
+Before plugging in the B2B board, make sure that your user is added to the `dialout` and `plugdev` groups.
+
+To check your groups, enter the following command:
+```bash
+groups ${USER}
+```
+
+If the `dialout` or `plugdev` group is missing, add your user to a group using this command (replace `{your_group}` by the group name, either `dialout` or `plugdev`):
+```bash
+sudo gpasswd --add ${USER} {your_group}
+```
+
+Logout then log back in to your user session.
+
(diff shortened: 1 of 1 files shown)


docs/boards/asterix/index.md
```diff
@@ -73,3 +73,74 @@ Correct orientation of the switches when using built-in programmer, USB-UART con
...
+```bash
+sudo gpasswd --add ${USER} plugdev
+```
+
+Logout then log back in and check that the console is now working properly by typing `help` once connected to the board.
+
+### I get a `OpenOCD: unable to find a matching CMSIS-DAP device` when trying to flash my watch. What should I do?
+First, make sure that you can connect to the console (see question above).
+
+Download [this file](https://raw.githubusercontent.com/raspberrypi/picotool/refs/heads/master/udev/60-picotool.rules)
```
jplexer: this doesn't seem correct. why tell a user to download a file to then just completely overwrite it?

jplexer (changes requested):
one small comment but other than that lgtm

docs/boards/asterix/index.md
MrXANA91: Edited to create the file manually

gmarull (changes requested):
I'd just add some default steps as many users will likely face that. Doesn't pico probe have a page where this is also described, so we can just point people there?

I definitely had to follow less steps (or at least simpler)

MrXANA91:
> Something that might be worth noting: I am using **Linux Mint 22.2**. It is based on Ubuntu 24.04 but it might have enough differences to complicate things compared to Ubuntu.

Someone on the Rebble discord had the same issues while being on Ubuntu 24.04.
I'll change this PR later today to add some default steps instead of having a FAQ, as @gmarull suggested earlier.

PR #751, opened 2026-01-26
fw/kernel: always reschedule wakeup timers on RTC time changes [FIRM-1224]

jplexer:
Wakeup timers use tick-based scheduling, so when RTC time changes they can fire early/late. Previously only substantial changes (>15s) triggered
rescheduling, missing small corrections.

Now unconditionally reschedule wakeup timers on any RTC time change.

### DIFF
--- a/src/fw/kernel/event_loop.c
+++ b/src/fw/kernel/event_loop.c
@@ -384,9 +384,14 @@ static NOINLINE void prv_extended_event_handler(PebbleEvent* e) {
           set_time_info->dst_changed ||
           ABS(set_time_info->utc_time_delta) > 15) {
         alarm_handle_clock_change();
-        wakeup_handle_clock_change();
         cron_service_handle_clock_change(set_time_info);
       }
+
+      // Always reschedule wakeup timers on any time change to prevent timers from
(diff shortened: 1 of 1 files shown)


src/fw/kernel/event_loop.c
```diff
@@ -384,10 +384,15 @@ static NOINLINE void prv_extended_event_handler(PebbleEvent* e) {
...
         alarm_handle_clock_change();
-        wakeup_handle_clock_change();
         cron_service_handle_clock_change(set_time_info);
       }
 
+      // Always reschedule wakeup timers on any time change to prevent timers from
+      // firing early/late. Wakeup timers use tick-based relative scheduling, so even
+      // small time changes (or time resets from crashes) can cause significant skew.
+      // This is critical for app wakeup events that must fire at precise times.
+      wakeup_handle_clock_change();
```
gmarull, line 394: I think we need to review PebbleOS time handling. Tick based timing should be simple, with events always firing >= the requested time, and then wall clock based events that should just happen based on RTC. In general, you either want:

1. Something in >= X secs/mins/etc. from now (so use a monotonic timer for that)
2. Something specifically at X hh:mm:ss (so use wall clock timer for that)

PR #763, opened 2026-01-28
Add a nix flake file for easy reproducible builds

matthewwardrop:
Greetings again!

I love all the progress you have been making (including many improvements to my original LSM6DSO and compass drivers). I returned to the project recently to try to build things from git head, and was having trouble getting my build environment set up properly again, and wanted to rectify that. I now have a really nice reproducible build environment that I can drop into using `nix develop` (especially nice since I also run on NixOS, but that is not necessary, and anyone on Linux and MacOS should be able to use this to get up to speed really quickly; note that I have tested this only on Linux). I was wondering if you would be interested in merging this in? Obviously I'd love that, since it would simplify my experience, but it could benefit others too. No pressure otherwise.

Usage instructions:

- install `nix` if you haven't already.
- run `nix develop`; and wait while the gcc cross-compiler and Python venv is set up for you.
- run `./waf` commands as normal.

This makes it super easy to reset your environment, or set it up on a different machine (`git clean -dxf; nix develop`). This is also "frozen" in time, so the build environment should never break (except the Python bits).

---

Note: While setting this up, it appeared that the assignment of AS <- CC configuration was happening too early in this environment and being overridden, so there is a one-line patch to the `wscript` file to fix that.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -36,10 +36,9 @@
           });
         in
         {
-          default = pkgs.mkShell {
+          default = pkgs.mkShellNoCC {
             hardeningDisable = [ "fortify" ]; # waf expects unoptimized builds
             buildInputs = with pkgs; [
-              clang_multi
               emscripten
               gcc
               gcc-arm-embedded-14_2r1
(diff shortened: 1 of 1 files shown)


jplexer:
Out of curiosity, why not? As a Nix fan myself this seems like a pretty low-cost addition. It's opt-in for contributors who use Nix and doesn't really require ongoing maintenance from the team. Would be nice to have!

gmarull:
> Out of curiosity, why not? As a Nix fan myself this seems like a pretty low-cost addition. It's opt-in for contributors who use Nix and doesn't really require ongoing maintenance from the team. Would be nice to have!

I prefer to keep a reduced official subset as supported platforms/dev methods. I'd only add new things if they come with a maintainership compromise. If, as a core developer, you use it, then I would be more favorable to adding this as an option since it's less probable it gets bit rot.

gmarull:
> oh yeah I use nix for other things already! I would be for merging it :)

Ok, then let's review. I'd also add a new "Advanced" section in docs, mentioning this is experimental, and not officially supported (so people expect for now 'best effort' quality)

matthewwardrop:
Thanks for taking the time to look at this @gmarull (and I definitely understand the reticence to increase the maintenance burden). To derisk this, did you want to take over this PR @jplexer ? I've cleaned it up slightly since the initial contribution, but happy to take your lead on this (including just maintaining this on my side for my purposes.... but didn't want to hoard the nice experience from others ;) ).

jplexer:
I wanted to push to my fork how did I manage to push to *your* repo???

jplexer:
gotcha lol

PR #766, opened 2026-01-29
Fix voice dictation frame drops by delaying recording start [FIRM-1254]

jplexer:
After commit 5d0999c increased BLE baseline to 600ms, voice dictation started dropping 40-60 frames because BLE negotiation takes 2-6s but the PDM buffer only holds 320ms.

A previous fix (f2634eed) was reverted because it used callbacks that caused a bunch of errors.

This fix uses a simple 3.5s delay before showing the microphone icon and reduces the baseline from 600ms to 300ms, giving BLE time to negotiate without blocking or callbacks.


jplexer:
I 100% agree with everything you've said here. The 3.5 second delay definitely still doesn't feel great even though this is already a way better experience than before.
My main goal with this PR was really just to get voice working in fast mode without adjusting the BLE rates too much. I honestly don't know why they were changed to 600ms in the first place, so I was hesitant to mess with them beyond what was necessary to make voice not suck as much.

PR #777, opened 2026-02-02
fw/drivers/accel: remove axis/direction from shake event

gmarull:
It's just useless. You just want to know if watch shake occurred, in any direction.


jplexer:
yeah thats better! We just gotta figure out how we do legacy apps that expect the axis at that callback

gmarull:
> yeah thats better! We just gotta figure out how we do legacy apps that expect the axis at that callback

The shake API is actually a new one; tap is kept the same, but we now need to feed the event from the sensor.

Issue #690, opened 2026-01-12
[Request] "Vibrate on the Hour" System Setting


saltedlolly:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

This feature often gets implemented by watchface developers, but this would be better suited as a system-wide option.

I gather there are plans to offload a number of watch settings to the mobile app, so perhaps this could be included in that, if not as an option in the watch "Settings > Vibrations" menu.

dkowis:
I don't suppose it could be extended to be an "hourly chime" either vibratory or auditory, or is that better suited as a different issue?

saltedlolly:
I agree, it does probably make sense to have an option for an hourly alert to:

- beep/chime and vibrate only
- beep/chime only
- vibrate only

Perhaps with an option for during Quiet Time?:

- vibrate only
- disabled

That said, the introduction of the speaker has complicated the meaning of “Quiet Time”. Does it mean do not disturb me, or do not disturb others? Overnight don’t disturb me, but what about in a meeting or movie theatre? Do you want vibrating alerts in this situation? This probably needs a separate discussion. We may need a “Theatre Mode”
Like Apple Watch has, which disables sounds and alerts for a finite amount of time.

GandalfDG:
The hourly vibration is implemented as "hourly notice" under sounds & haptics, but I would also like the option for an audible alert on speaker-equipped devices. I've gotten used to the beep on my Casio as an indication of time passing without tricking me into thinking I need to even look at the watch for a notification. Since the vibration is implemented, should this be closed and have the audible signal be a new issue?

PR #786, opened 2026-02-04
feat: add new vibe triggers for hourly and disconnection notice

rojas-claudio:
This PR adds two new vibration triggers to all watches, one for the top of the hour and the other for a disconnect notice (both of which are already features in a number of community-made watchfaces, but never a system-wide feature). 

That being said, I am seeking feedback, specifically in regards to triggering the vibe score(s) in `clock.c` and `debounced_connection_service.c`. PebbleOS currently expects a system app (ResAppNum = 0) to be in focus when creating the vibe score, however, this will not work when the user is running a third-party app or watchface. At the moment, there is a workaround for this, but I am not sure if this is the cleanest or will be an accepted solution. Any pointers would be much appreciated :)

Resolves issue #690.

### DIFF
--- a/include/pbl/services/vibes/vibe_client.h
+++ b/include/pbl/services/vibes/vibe_client.h
@@ -9,7 +9,9 @@ typedef enum VibeClient {
   VibeClient_Notifications = 0,
   VibeClient_PhoneCalls,
   VibeClient_Alarms,
-  VibeClient_AlarmsLPM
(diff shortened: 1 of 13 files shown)


ericmigi:
@jplexer want to review?

jplexer (approved):
looks good but @rojas-claudio please fix the tests 🙃

PR #811, opened 2026-02-10
sdk: add per-function updateRevision for minimum firmware detection [SDK-60]

jplexer:
Decouple symbol table ordering (addedRevision) from firmware compatibility by adding an updateRevision field to exported_symbols.json. After linking, the build system analyzes the app ELF to find which SDK functions are used, computes the max updateRevision among them, and patches the binary's SDK version field. This allows apps compiled with a newer SDK to target older firmware when they only use older APIs.

Revisions <= 89 map to minor 0x56 (historical cutoff); revisions >= 90 map directly to the minor byte (90 = 0x5A, 91 = 0x5B, etc.).

Basically a version of #577 that should not break existing behavior

### DIFF
--- a/sdk/tools/determine_min_sdk_version.py
+++ b/sdk/tools/determine_min_sdk_version.py
@@ -1,0 +1,67 @@
+# SPDX-FileCopyrightText: 2024 Google LLC
+# SPDX-License-Identifier: Apache-2.0
+
+import json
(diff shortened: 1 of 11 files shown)


jplexer:
Confirmed working by rebasing moddable repo onto this, compiling some apps with it and seeing if non-moddable api applications run on a older fw

PR #815, opened 2026-02-11
More cleanups

gmarull:


wscript
```diff
@@ -91,21 +82,12 @@ def options(opt):
...
     opt.add_option('--board', action='store',
-                   choices=[ 'bb2',
-                             'ev2_4',
-                             'v1_5',
-                             'v2_0',
-                             'snowy_bb2',  # alias for snowy_dvt, but with #define IS_BIGBOARD
-                             'snowy_evt2',
+                   choices=[ 'snowy_bb2',  # alias for snowy_dvt, but with #define IS_BIGBOARD
                              'snowy_dvt',
-                             'snowy_s3',
```
jplexer, line 101: I disagree with dropping snowy_s3 because that is the board name for Pebble Time Steel. I don't know exactly what the difference is between Pebble Time and Pebble Time Steel in that regard, but it's worth noting nonetheless.

gmarull: Note the objective is just to keep existing legacy boards for QEMU purposes only, so only single snowy is kept (snowy_dvt/snowy_bb2). Did snowy_s3 have any special use case from emulation perspective?

jplexer: nah

Issue #233, opened 2025-08-28
nPM1300 driver needs better configurability


gmarull:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

As of today, the nPM1300 driver contains a few hacks, as it initially contained hardcoded values that only applied to Asterix (like, e.g., voltages or LSW/LDO modes). When Obelix was added, some ifdeffery was added to "move quick," but it needs to be fixed.

PR #829, opened 2026-02-14
fw/drivers/pmic: replace npm1300 platform ifdefs with board-level config

leopoldch:
Issue #233

### Why

The npm1300 driver used hardcoded `#if PLATFORM_XXX`. Several `TODO` and `FIXME` comments in the code already acknowledged this needed to be cleaned up. Adding a new board meant editing the driver itself, which is fragile and hard to review.

### What

All platform-specific PMIC configuration is moved into new fields in `Npm1300Config` (voltages, enable flags, modes, SW control, erratum workaround). Each board file (`board_asterix.c`, `board_obelix.c`, `board_getafix.c`) now declares its own complete power rail config, and the driver in `npm1300.c` applies it generically without any `#if PLATFORM_*`.

### Note
The patch compiles cleanly but has not been tested on hardware. 
Also note that some magic values remain in this part of the codebase

### DIFF
--- a/src/fw/board/boards/board_asterix.c
+++ b/src/fw/board/boards/board_asterix.c
@@ -207,12 +207,32 @@ IRQ_MAP_NRFX(PWM0, nrfx_pwm_0_irq_handler);
 IRQ_MAP_NRFX(RTC1, rtc_irq_handler);
 
 const Npm1300Config NPM1300_CONFIG = {
-  // 128mA = ~1C (rapid charge)
-  .chg_current_ma = 128,
-  .dischg_limit_ma = 200,
-  .term_current_pct = 10,
-  .thermistor_beta = 3380,
-  .vterm_setting = NPM1300_VTERM_4V20,
+    .chg_current_ma = 128,        // 128mA = ~1C (rapid charge)
+    .dischg_limit_ma = 200,       // ~1.6C burst, hardware protection floor
+    .term_current_pct = 10,       // 12.8mA
+    .thermistor_beta = 3380,      // 10kΩ
+    .vterm_setting = NPM1300_VTERM_4V20,
+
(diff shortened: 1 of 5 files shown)


src/fw/board/boards/board_asterix.c
```diff
@@ -208,15 +207,36 @@ IRQ_MAP_NRFX(PWM0, nrfx_pwm_0_irq_handler);
...
-  .chg_current_ma = 128,
-  .dischg_limit_ma = 200,
-  .term_current_pct = 10,
-  .thermistor_beta = 3380,
+    // 128mA = ~1C (rapid charge)
+    .chg_current_ma = 128,
+    .dischg_limit_ma = 200,
+    .term_current_pct = 10,
+    .thermistor_beta = 3380,
+    .vterm_setting = NPM1300_VTERM_4V20,
```
gmarull: please, expose human readable settings at board level, no magic constants that are impl details

gmarull (changes requested):
follow commit guidelines

src/fw/board/boards/board_asterix.c
```diff
@@ -208,15 +207,36 @@ IRQ_MAP_NRFX(PWM0, nrfx_pwm_0_irq_handler);
...
+
+    // Buck1 (1.8V)
+    .buck1_enable = true,
+    .buck1_voltage_sel = 8,       // 1.8V
+
+    // Buck2 (3.0V)
+    .buck2_enable = true,
+    .buck2_voltage_sel = 20,      // 3.0V
+    .buck_sw_ctrl_sel = 3,        // both of them, load
```
gmarull: feedback not addressed

leopoldch: This is now addressed

PR #833, opened 2026-02-16
fw/applib/ui/vibes: add per-segment amplitude control to public SDK

mdrunk:
Expose vibration motor intensity control to apps via a new vibes_enqueue_custom_pattern_with_amplitudes() API. The internal syscall (sys_vibe_pattern_enqueue_step_raw) and hardware drivers already support per-step strength values; this bridges the gap in the public SDK.

Also fix a pre-existing issue where active vibration patterns were not explicitly stopped on app exit by adding
sys_vibe_pattern_clear() to the app cleanup path.

### DIFF
--- a/src/fw/applib/ui/vibes.h
+++ b/src/fw/applib/ui/vibes.h
@@ -49,6 +49,42 @@ typedef struct {
   uint32_t num_segments;
 } VibePattern;
 
+/** Data structure describing a vibration pattern with per-segment amplitude control.
+ Each segment has a duration and an amplitude. Amplitude 0 means no vibration
+ (off), and 100 means maximum strength. Values above 100 are clamped.
+
+ Example code:
+ \code{.c}
+ // Ramp-down pattern: 100% for 200ms, 50% for 200ms, 25% for 200ms:
+ static const uint32_t const segments[] = { 200, 200, 200 };
+ static const uint32_t const amplitudes[] = { 100, 50, 25 };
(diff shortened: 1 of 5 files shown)


gmarull:
> lgtm
> 
> is there a way to re-do ci? @gmarull

rebase

src/fw/applib/ui/vibes.h
```diff
@@ -49,6 +49,47 @@ typedef struct {
   uint32_t num_segments;
 } VibePattern;
 
+/** Data structure describing a vibration pattern with per-segment amplitude control.
+ A pattern consists of at least 1 vibe-on duration, optionally followed by
+ alternating vibe-off + vibe-on durations. Each "on" segment's vibration
```
kljohann: > followed by alternating vibe-off + vibe-on durations

Why enforce this alternation, though? If you just pass along the amplitudes one could, e.g., encode ramps of increasing/decreasing altitude: `100, 75, 50, 25`. (Whether that makes sense from a UX/HW perspective or not.) Alternating on/off could still be accomplished by alternating between zero and non-zero amplitudes. (In your proposed version zero would anyways be the most natural amplitude value to use for "off" segments.)

mdrunk: I actually agree with you.  I was just trying to match the existing VibePattern behavior.  if no one else has a comment on this, I will make that change tonight.

PR #834, opened 2026-02-16
waftools: Add many -Wno-error flags for newer GCC

jplexer:


gmarull (changes requested):
We should be careful and fix warnings instead; we're already relying on too many UB...

jplexer:
we cant fix these ourselves, these are all in user code

gmarull:
> nah its just that we used to ship gcc 6.1 and now with the new SDK i want to ship with gcc 14

so for me to understand:

- Warnings/issues caused by our SDK: we should fix as much as we can
- Caused by existing user apps/watch faces code: they should silence/fix warnings

In which case are we?

PR #837, opened 2026-02-16
waftools: Add helper to suppress newer GCC errors

jplexer:
Provide pbl_suppress_newer_gcc_warnings() that appends -Wno-error flags to each target platform's CFLAGS. Call it from wscript's configure() to avoid GCC 11+/14+ warnings being treated as errors.

To fix the build users can add one line to the app's wscript:
```
def configure(ctx):
    ctx.load('pebble_sdk')
    ctx.pbl_suppress_newer_gcc_warnings()  # opt-in workaround
```

### DIFF
--- a/waftools/pebble_sdk_gcc.py
+++ b/waftools/pebble_sdk_gcc.py
@@ -2,6 +2,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import os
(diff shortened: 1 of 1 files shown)


gmarull (approved):
thanks, that feels better :-)

PR #839, opened 2026-02-16
fw/sf32lb52: enable hardware stack overflow detection via PSPLIM/MSPLIM

spr4bhu:
This PR wires up the hardware stack limit registers on SF32LB52 (Cortex-M33) and routes STKOF violations into the existing stack overflow handler. The required FreeRTOS-side PSPLIM context save/restore changes have already been put up in https://github.com/coredevices/FreeRTOS-Kernel/pull/5 in the FreeRTOS-Kernel submodule repo.

Changes:
- Set MSPLIM to __isr_stack_start__ in Reset_Handler — replaces the FIXME that had it disabled (0); PSPLIM is left to FreeRTOS to set per-task via portSETUP_TCB
- Handle the STKOF bit (CFSR bit 20, ARMv8-M only) in usagefault_handler_c: receive lr from the fault entry stub to identify which stack was active, advance SP by 256 bytes to give the handler a landing zone, then call attempt_handle_stack_overflow(); define SCB_CFSR_STKOF_Msk fallback for toolchains that don't provide it
- Pass lr from the UsageFault_Handler assembly stub to the C handler (mov r1, lr)
- Add stack_offset_to_reg_index_v3 / thread_state_size_v3 to qemu_gdb_proxy.py for the CM33 stacking layout (which now includes the PSPLIM slot), enabling correct register display in the QEMU GDB proxy

### DIFF
--- a/src/fw/kernel/fault_handling.c
+++ b/src/fw/kernel/fault_handling.c
@@ -395,8 +395,30 @@ void BusFault_Handler(void) {
         "b %0\n" :: "i" (busfault_handler_c));
 }
 
-static void usagefault_handler_c(unsigned int* stacked_args) {
+// STKOF = bit 4 of UFSR = bit 20 of CFSR (ARMv8-M only, reads as 0 on CM3/CM4)
+#ifndef SCB_CFSR_STKOF_Msk
+#define SCB_CFSR_STKOF_Msk (1UL << 20)
+#endif
+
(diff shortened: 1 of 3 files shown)


gmarull:
sign-off requires real name

gmarull:
update submodule sha to `64a8ce153a3640d357793d6d85b45736a825460b`

gmarull (commented):
forgot, please, also fix commit author name (now set to your github id)

Issue #605, opened 2025-12-16
Quick launch shortcuts: a pinch more of those


obar:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

I'd love to have a few more quick launch shortcuts.

I press my Pebble Time buttons with a pinch between thumb (left side of watch, bottom) and forefinger (button). I can reliably press back+another button by putting my thumb on the back button instead. Is it wild to think that with back+[up,middle,down]  3 more quick launch options could be available, or even 6 if short and long presses are used?

### Anything else?

Aside from adding a tiny delay in button processing to determine if the user means to press a pair of buttons, of course there is additional UI complexity. But this is the power user smart watch and I would certainly put more shortcuts to work.

A point to keep in mind, there's already a pair of button presses being tested for in the firmware. It's the reset combo, select and back together, but that's using a much longer hold.

I initially posted this in a different repo but I think this is the correct place for it.

Somewhat related feature request: https://github.com/pebble-dev/pebble-firmware/issues/208

obar:
> for example, I wanted to try pressing the middle and next buttons simultaneously on my Duo2

I don't mean pressing two buttons on the same side of the watch. How is it when pressing the back button in combination with any other button?

saltedlolly:
I can see that a combo pressing top-left and top-right together using your thumb and index finger would be a good fifth quick-launch shortcut. That could work well.

And possibly also a combo that involves pressing top-right and bottom-right together by anchoring your thumb on the left side of the watch and pressing the two buttons with your index and middle fingers. That could also work.

I am less convinced by a combo involving the other buttons, such as top-left and bottom-right, which would result in a rotating motion of the watch as you try and pinch both together.

asyba:
@saltedlolly I was testing on my custom fw, to press BACK + UP for a quick launch 
but is not working as expected,  because its a simple tap and to sync two button at the same time press, I don't think it was designed for that.
because it triggers the backlight or the up shortcut.
sometimes it works other times it doesn't, and other it work but right away it trigger the UP shortcut also.

Maybe playing with how much to wait for pressing, etc but eventually can create false actions that the user does not want.
And adding more time to waits to detect the combo, will make other simple shortcuts to delay more when using them.

maybe doing a hold back+up can work better?

saltedlolly:
[...]

PR #846, opened 2026-02-17
Feat: 2-Clicks QuickLaunch

MrXANA91:
This PR adds a new feature to QuickLaunch: the 2-Clicks QuickLaunch. Instead of pressing only one button, you press two, one after the other.

2-Clicks adds a submenu with three choices: Tap Up, Tap Center, and Tap Down. Each of the 6 QuickLaunch buttons can be configured with 2-Clicks, effectively multiplying shortcuts slots by 3.
Not all three slots must be taken, you can only setup the Tap Up and/or Tap Down if you wish.

This new feature aims to give a different approach than #707 to solving #605

# Demo
pebble_screenshot_2026-02-17_11-49-22
pebble_screenshot_2026-02-17_11-49-26
pebble_screenshot_2026-02-17_11-49-48
pebble_screenshot_2026-02-17_11-50-02

In a QuickLaunch button menu, choosing 2-Clicks will open another submenu in which we can choose any app (except 2-Clicks of course).
Once every app are setup, you can go back to the Watchface.

pebble_screenshot_2026-02-17_11-50-09

Pressing the 2-Clicks button will open a small action menu and will display for each app the icon, the app name and in the ActionBar the button to press.

When opening the app, pressing the Back button will bring the screen to the watchface, like any other app launched by QuickLaunch.

# Known bug(s)
pebble_screenshot_2026-02-17_11-51-42
pebble_screenshot_2026-02-17_11-51-50

- The default app icon generic_watchapp_icon and some user app icons are incorrectly displayed
- When the app name is long, the text will not be centered vertically, rendering the menu esthetically awkward
- Will not properly work for rounded display yet


MrXANA91:
@jplexer Not marking it as ready for review because of the known bugs, but a code and an overall review are still welcome. A code review might help me solve the icon display bug I am getting.

asyba:
Interesting idea, I also considered a submenu.

But by then it would involve more clicks and layer, similar to pressing the middle button that displays the main menu.
However, you have to press more to scroll...

Perhaps when we can rearrange the main menu as we want, it will be closer, but there might still be an extra click required to scroll.

I wonder if this could be done directly with a watch app without editing the firmware.
For example, creating a watch app with three options and configuring this app for quick launch.
However, we would need to know the UUIDs of each app beforehand.
Of course, this assumes the SDK allows calling other apps from within an app...

PR #854, opened 2026-02-18
fix: Improved appearance for health screen

tilden:
This change fixes the poor rendering of the rotated 4-point square in the HRM UI, and improves the contrast by adding a background color to the heart icon.

**Background**
In https://github.com/coredevices/PebbleOS/pull/850, the health screen contrast was improved by changing the background color to white, and by replacing the glitched square with a circle. However, this causes the heart rate screen to no longer visually mesh with the rest of the health UI (which is dark, and uses angular shapes.)

After discussion, @hellcp determined that the root cause for the glitched rectangle was incorrect metrics. By correcting these, it is now possible to display a rectangle that does not glitch:
image

Furthermore, the low-contrast heart graphic appears to be caused by the heart having a transparent, as opposed to white, background color. This commit should also correct this:
[Screencast From 2026-02-18 08-57-23.webm](https://github.com/user-attachments/assets/130ba5e6-176d-4408-ac67-ebb1e6c9df69)

Thanks to @hellcp for figuring all this out and implementing the changes over in https://github.com/pebble-dev/pebble-firmware/pull/319


gmarull:
Since this seems to be derived work from https://github.com/pebble-dev/pebble-firmware/pull/319, can the original author confirm it's Apache-2.0? pebble-firmware messed up licenses a while ago, so I want to be sure.

PR #866, opened 2026-02-19
treewide: revert ruff lint for sdk files

jplexer:
some things that ruff sees as "unused" are actually quite important for the SDK


gmarull (changes requested):
Some of the fixes are valid; I guess some of these scripts are used by apps using import *? Add a `# noqa: FXXX` to unused imports we want to preserve for 3rd party

PR #871, opened 2026-02-20
fw/apps/system/settings: fix text and background colors after coloring removal

jplexer:
Commit 06926f11 removed the custom normal colors but left the highlight foreground as GColorWhite and the window background as GColorBlack, inconsistent with the new white background theme.


gmarull (approved):
thx

PR #874, opened 2026-02-21
feat: updated contrast and theme for health 

tilden:
In #854, it was noted that there was a potential plan to make health menus lighter and more consistent.

> Note we're thinking to make all the health menus lighter to make things consistent and overall more readable

_Originally posted by @gmarull in https://github.com/coredevices/PebbleOS/issues/854#issuecomment-3922899957_

My goal here was to achieve this in a way that was consistent with the Pebble visual language established in PebbleOS 3.0. 

1. Used higher-contrast color palette consistent with PebbleOS 3.x fitness visualizations
2. Enabled the black stroke for color watches (previously it was only visible on BW watches)
3. Used a light grey background matching the PebbleOS 3.x Health app and built-in Music app)
4. Re-enabled deep sleep visualization on BW watches

There are several known issues with this branch:

- On round watches, the bottom left stroke line draws slightly too far to the left
- On round watches, the progress bar lacks a bottom outline
- On color watches, the sleep icon should have a black outline; _while I replaced the icon in /resources, I was unable to find a way to rebuild the snowy pbpack resource file that is used for the test cases._
- On color watches, the steps hexagon outline has some points that are slightly misaligned.

I can definitely fix all those issues, but before moving forward I wanted to get confirmation if the Core team would be open to merging changes along these lines. Thoughts?

Example images (generated via test run):

| Before| After |
| ------------- | ------------- |
| image| image |
|  image| image|
| image | image|


ericmigi:
CleanShot 2026-02-21 at 10 57 09@2x

> Re-enabled deep sleep visualization on BW watches
 
 Ouch, did these get disabled somewhere? Was there any comment about why they were disabled?
 
 We used different background colours to more clearly differentiate between these views. It makes it feel a bit more vibrant! But I agree that contrast feels a bit worse on colour here. Any other ideas that keep the background colours distinct but increase contrast?

ericmigi:
yeah that's a good catch! great to enable that. let's see how it feels.

Re colours, tough thing about changing these is that everyone has different opinions/thoughts. So inertia always prefers the status quo. We do have theming now, maybe add these as theme options?

tilden:
> yeah that's a good catch! great to enable that. let's see how it feels.

Great! I can make a separate MR that solely contains the deep sleep color change (should just be 1 line)

> Re colours, tough thing about changing these is that everyone has different opinions/thoughts. So inertia always prefers the status quo.

I agree! That's why I was a bit surprised to hear that all the fitness screens were changing to a light theme. (The HRM screen was [changed to a new lighter color scheme in a recent commit](https://github.com/coredevices/PebbleOS/commit/1abd5ceced4a2f141d5f30d57f18aa538eb7d0c9)).

> We do have theming now, maybe add these as theme options?

Would be neat if it could tie into PebbleOS theming overall!

PR #875, opened 2026-02-21
fw/popups/phone_ui: fix call duration timer showing on unanswered calls

### LINKED ISSUE #808: Duration of not answered call is showing

### Current Behavior

Sometimes I'm getting call duration, an abnormal one. Without answering at all.

Image

### Expected Behavior

Not get a call duration, until call is answered.

### Steps To Reproduce

1. Get an incoming call
2. Do not answer it
3. See abnormal call duration anyway

### Version

4.9.122

### Host OS

N/A

### Watch

Pebble 2 Duo (Asterix)

### Anything else?

I was about to leave a bug report through mobile app, but bug report wasn't uploading...

spr4bhu:
This PR complements the mobileapp fix ([coredevices/mobileapp#124](https://github.com/coredevices/mobileapp/pull/124)) by adding guards against stale call_start_time values and explicit status text handling for all call end scenarios.

Changes:

- Add call_start_time == 0 guard in prv_update_call_time() — if the timer callback fires after being stopped, it won't compute a duration from a stale timestamp
- Reset call_start_time = 0 in prv_stop_call_duration_timer() — ensures no stale timestamp can persist after stopping the timer
- Add CallStatus.ENDED enum and update prv_set_status_text() to explicitly handle ENDED — replaces the heuristic that used call_start_time to infer "Call Ended" vs "Call Declined"
- Update phone_ui_handle_call_end() to use ENDED for unanswered calls (caller hangs up before answer) instead of DECLINED
- Update phone_ui_handle_call_hide() for iOS unanswered calls to show "Call Ended" with proper end-of-call UI (icon, color, delay) instead of immediately popping the window

Closes #808

### DIFF
--- a/src/fw/popups/phone_ui.c
+++ b/src/fw/popups/phone_ui.c
@@ -118,7 +118,8 @@ typedef struct {
 typedef enum {
   ACCEPTED,
   DECLINED,
-  DISCONNECTED
+  DISCONNECTED,
+  ENDED
 } CallStatus;
 
 typedef struct {
@@ -515,6 +516,9 @@ static void prv_update_call_time(void *unused) {
   if (s_phone_ui_data == NULL) {
     return;
   }
+  if (s_phone_ui_data->call_start_time == 0) {
+    return;
+  }
 
   if (layer_get_hidden(&s_phone_ui_data->call_status_text_layer.layer)) {
     prv_show_call_status();
@@ -550,6 +554,7 @@ static void prv_start_call_duration_timer(void) {
 static void prv_stop_call_duration_timer(void) {
   evented_timer_cancel(s_phone_ui_data->call_duration_timer);
   s_phone_ui_data->call_duration_timer = EVENTED_TIMER_INVALID_ID;
+  s_phone_ui_data->call_start_time = 0;
 }
 
 static void prv_set_status_text(CallStatus status) {
@@ -559,14 +564,12 @@ static void prv_set_status_text(CallStatus status) {
   } else if (status == DISCONNECTED) {
     i18n_get_with_buffer("Disconnected", s_phone_ui_data->call_status_text_buf,
                          CALL_STATUS_BUFFER_LENGTH);
+  } else if (status == ENDED) {
+    i18n_get_with_buffer("Call Ended", s_phone_ui_data->call_status_text_buf,
+                         CALL_STATUS_BUFFER_LENGTH);
   } else {
-    if (s_phone_ui_data->call_start_time) {
-      i18n_get_with_buffer("Call Ended", s_phone_ui_data->call_status_text_buf,
-                           CALL_STATUS_BUFFER_LENGTH);
-    } else {
-      i18n_get_with_buffer("Call Declined", s_phone_ui_data->call_status_text_buf,
-                           CALL_STATUS_BUFFER_LENGTH);
-    }
+    i18n_get_with_buffer("Call Declined", s_phone_ui_data->call_status_text_buf,
+                         CALL_STATUS_BUFFER_LENGTH);
   }
 
   text_layer_set_text(&s_phone_ui_data->call_status_text_layer,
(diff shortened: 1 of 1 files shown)


src/fw/popups/phone_ui.c
```diff
@@ -1174,23 +1177,31 @@ void phone_ui_handle_call_end(bool call_accepted, bool disconnected) {
...
+  if (!s_phone_ui_data) {
+    return;
+  }
+  prv_stop_ringing();
+  prv_stop_call_duration_timer();
+  prv_action_bar_setup(PhoneCallActions_None);
+  prv_set_icon_resource(TIMELINE_RESOURCE_DISMISSED_PHONE_CALL);
+  prv_set_window_color(DECLINE_COLOR, true);
+  prv_set_status_text(ENDED);
+  prv_window_pop_with_delay(CALL_END_DELAY_MS);
```
gmarull: this looks overcomplicated compared to the previous impl

src/fw/popups/phone_ui.c
```diff
@@ -1174,23 +1177,31 @@ void phone_ui_handle_call_end(bool call_accepted, bool disconnected) {
...
+    prv_set_status_text(ENDED);
   } else {
     prv_set_icon_resource(TIMELINE_RESOURCE_DISMISSED_PHONE_CALL);
     prv_set_window_color(DECLINE_COLOR, true);
     if (disconnected) {
       prv_set_status_text(DISCONNECTED);
     } else {
-      prv_set_status_text(DECLINED);
+      prv_set_status_text(ENDED);
     }
```
gmarull, line 1188: both branches now set 'ENDED', code is hard to follow

gmarull (changes requested):
please, follow commit guidelines

src/fw/popups/phone_ui.c
spr4bhu: The previous phone_ui_handle_call_hide() just called prv_window_pop() which silently dismissed the UI with no status text or delay

I will simplify it to reuse the phone_ui_handle_call_end(bool call_accepted, bool disconnected) function

src/fw/popups/phone_ui.c
spr4bhu: Both branches set ENDED because the status text is always 'Call Ended' regardless of whether the call was answered (The status appears on screen only when the call ends so ACCEPTED is the wrong thing to display)

the if else only controls the icon and colour

src/fw/popups/phone_ui.c
```diff
@@ -1174,23 +1177,22 @@ void phone_ui_handle_call_end(bool call_accepted, bool disconnected) {
   if (call_accepted) {
     prv_set_icon_resource(TIMELINE_RESOURCE_DURING_PHONE_CALL);
     prv_set_window_color(ACCEPT_COLOR, true);
-    prv_set_status_text(ACCEPTED);
+    prv_set_status_text(ENDED);
```
gmarull, line 1180: While call has ended, it was ended but initially accepted, so should be kept as is unless it was previously wrong.

src/fw/popups/phone_ui.c
```diff
@@ -1174,23 +1177,22 @@ void phone_ui_handle_call_end(bool call_accepted, bool disconnected) {
...
   } else {
     prv_set_icon_resource(TIMELINE_RESOURCE_DISMISSED_PHONE_CALL);
     prv_set_window_color(DECLINE_COLOR, true);
     if (disconnected) {
       prv_set_status_text(DISCONNECTED);
     } else {
-      prv_set_status_text(DECLINED);
+      prv_set_status_text(ENDED);
```
jplexer: Should still say call declined

src/fw/popups/phone_ui.c
```diff
@@ -1174,23 +1177,22 @@ void phone_ui_handle_call_end(bool call_accepted, bool disconnected) {
...
     }
   }
 
   prv_window_pop_with_delay(CALL_END_DELAY_MS);
 }
 
 void phone_ui_handle_call_hide(void) {
-  // Just pop the window - it'll handle all the cleanup
-  prv_window_pop();
+  phone_ui_handle_call_end(false /* call_accepted */, false /* disconnected */);
```
jplexer: call hide should hide the ui, not display something before hiding

PR #890, opened 2026-02-23
Raise 4K storage

### LINKED ISSUE #416: Raise 4K storage limit

### Proposed feature request

Currently, every app is limited to only storing maximum of 4KB of data. Even on the Pebble Steel with 8MB storage, that means that every app can only use 0.05% of the device's storage. And newer watches have 16MB storage, presumably P2D and PT2 have even more.

Could we get that limit raised up, or maybe be dynamic depending on the platform / available storage?

### Alternatives

_No response_

### Anything else?

_No response_

gmarull:
Fixes #416


jplexer:
Superseded by #1197

PR #899, opened 2026-02-24
treewide: remove RockyJS support

gmarull:


tools/generate_native_sdk/exported_symbols.json
```diff
@@ -152,20 +151,6 @@
...
       "exports" : [
-        {
-          "type": "group",
-          "name": "Rocky",
-          "appOnly": true,
-          "exports": [
-            {
-              "_comment": "Marked as deprecated so that the function it self won't appear in pebble.h",
-              "type": "function",
-              "name": "rocky_event_loop_with_resource",
```
jplexer, line 163: Dont remove from SDK, replace with stub instead, since there are existing apps that use rocky

PR #914, opened 2026-02-27
tools/wscript: fix image_recovery on SF32LB52 (PBLBOOT header mismatch)

### LINKED ISSUE #913: `./waf image_recovery` writes wrong header format on SF32LB52 (PBLBOOT)

### Current Behavior

Running `./waf image_recovery --tty /dev/ttyACM0` on an SF32LB52 board (Obelix, Getafix) completes with "Success!", but when the device boots into recovery (boot bit 17 set), the bootloader rejects the image:

```
PBLBOOT:0> I: PebbleOS bootloader 0.9.16
PBLBOOT:0> I: Forced PRF load requested
PBLBOOT:0> E: PRF not found or invalid (err -22)
PBLBOOT:0> E: Failed to load PRF (err -22)
PBLBOOT:0> E: System panic (0x00020016), press any button to reset
```

The device enters an infinite panic loop if the bootloader does not clear bit 17 on failure.

Directly writing the same PRF binary via sftool works fine:
```bash
sftool -c SF32LB52 -p /dev/ttyACM0 write_flash build/prf/src/fw/tintin_fw.bin@0x12A20000
```

### Expected Behavior

`./waf image_recovery --tty /dev/ttyACM0` should write a valid recovery image to SAFE_FIRMWARE so that the PBLBOOT bootloader can load PRF without errors.

### Steps To Reproduce

1. Configure for an SF32LB52 board:
   ```bash
   ./waf configure --board obelix_pvt
   ```
2. Build PRF:
   ```bash
   ./waf build_prf
   ```
3. Flash main firmware and boot the device normally:
   ```bash
   ./waf flash_fw --tty /dev/ttyACM0
   ```
4. Flash recovery via `image_recovery`:
   ```bash
   . .venv/bin/activate
   ./waf image_recovery --tty /dev/ttyACM0
   ```
5. Trigger PRF boot from serial console:
   ```
   boot bit set 17 1
   reset
   ```
6. Observe the bootloader output — it panics with "PRF not found or invalid (err -22)".

### Version

19d969f7e92aacebc4ac0225063ed24ec11c63b9

### Host OS

Ubuntu 24.04 LTS

### Watch

Pebble Time 2 (Obelix)

### Root cause

`image_recovery` uses the PULSE flash imaging tools (`pulse_flash_imaging.py` / `pulse_legacy_flash_imaging.py`), which prepend an **old 12-byte FirmwareDescription** header (`description_length`, `firmware_length`, `checksum`) to the PRF binary before writing it to SAFE_FIRMWARE (0x12A20000).

The PBLBOOT bootloader (`boot/src/firmware.c`) expects a **28-byte FirmwareHeader** starting with magic `0x96f3b83d`. It reads the first 4 bytes at the partition address and checks for this magic. Instead it finds `0x1c000000` (the `description_length` field of the old header), fails the magic check, and returns `-EINVAL` (-22).

This only affects SF32LB52 boards (Obelix, Getafix) which use pblboot. Older platforms (STM32F4, NRF52840) use a legacy bootloader that expects the old FirmwareDescription — the PULSE tools produce the correct format for them.

The SF32LB52 build already outputs `tintin_fw.bin` with the correct PBLBOOT header (magic at offset 0). The problem is entirely on the host side: the PULSE tool wraps the already-correct binary in an additional old-format header.

### Header format comparison

**Old (FirmwareDescription, 12 bytes)** — used by STM32F4/NRF52840 bootloaders:
```c
struct FirmwareDescription {
    uint32_t description_length;  // 12
    uint32_t firmware_length;
    uint32_t checksum;
};
```

**New (FirmwareHeader, 28 bytes)** — used by pblboot on SF32LB52:
```c
struct firmware_header {
    uint32_t magic;          // 0x96f3b83d
    uint32_t header_length;  // 28
    uint64_t timestamp;
    uint32_t start_offset;   // offset to firmware payload
    uint32_t length;
    uint32_t crc;            // CRC32-IEEE
};
```

ClusterM:
Fixes #913

image_recovery used PULSE flash tools that prepend the old 12-byte FirmwareDescription header. The PBLBOOT bootloader expects a 28-byte FirmwareHeader (magic 0x96f3b83d), causing "PRF not found or invalid (err -22)" on Obelix/Getafix.

Add sftool_flash_imaging.py that writes images directly via sftool without prepending any header. Select it in _get_pulse_flash_tool() for MICRO_FAMILY == 'SF32LB52'.

### DIFF
--- a/wscript
+++ b/wscript
@@ -187,6 +187,9 @@ def options(opt):
     opt.add_option('--no-pulse-everywhere',
                    action='store_true',
                    help='Disables PULSE everywhere, uses legacy logs and prompt')
+    opt.add_option('--force-pulse',
+                   action='store_true',
+                   help='Force PULSE-based flashing even on SF32LB52 (default: sftool)')
 
 def handle_configure_options(conf):
     if conf.options.noprompt:
@@ -1397,7 +1400,7 @@ def _is_pulse_everywhere(ctx):
 
 
 def _get_pulse_flash_tool(ctx):
-    if ctx.env.MICRO_FAMILY == 'SF32LB52':
+    if ctx.env.MICRO_FAMILY == 'SF32LB52' and not ctx.options.force_pulse:
         return "sftool_flash_imaging"
     if _is_pulse_everywhere(ctx):
         return "pulse_flash_imaging"


gmarull (commented):
Note that pulse-based flashing is still available on SF32LB52 (unless you build with `--release`), but it's pretty useless in these SoCs as you can just flash directly into QSPI. So it feels like the right change to do (on nRF we could also do that as they support flashing external QSPI). However, I'd suggest to:

- Leave the pulse option available for now
- On SF32 SoCs, default to using sftool
- Add an option to force pulse if users want to

PR #941, opened 2026-03-08
graphics/fonts: add Hebrew (RTL) text rendering support

alonmln:
## Summary

Adds Hebrew text rendering support. Hebrew notifications and UI text render right-to-left with correct glyph shapes on all Pebble targets.

**BiDi renderer fix** (`text_layout.c`, `rtl_support.c`): The segment-based BiDi path in `walk_line()` iterates UTF-8 bytes directly, bypassing `char_iter`. It was not filtering out invisible codepoints (ZWJ U+200D, variation selector U+FE0F, skin tone modifiers, control characters) that `char_iter` silently drops. These caused wildcard glyph rendering and corrupted line layout. A new `prv_codepoint_is_invisible()` helper is applied at all three scan sites in the BiDi path. `MAX_RTL_CODEPOINTS` is also raised from 16 to 32 for longer word forms.

**Hebrew font glyphs** (`resources/`): 34 Hebrew codepoints (U+05D0–U+05F4, punctuation, shekel sign U+20AA) from the Alef typeface by HaGilda & Mushon Zer-Aviv (© 2012, SIL Open Font License 1.1) are merged into all 11 base Gothic PBF sizes across both `normal/` and `common/` resource directories. The generator script and source TTFs are included for reproducibility.

## Testing

Verified on QEMU (`snowy_bb2`) by injecting Hebrew phone notifications. Title and body render right-aligned and right-to-left. Mixed Hebrew/Latin text and Hebrew text containing emoji render without wildcard glyphs.


ericmigi:
Yes, please create a pblpack and we would be happy to add it to the list of Language Packs in the mobile app!

alonmln:
Hi @ericmigi, thanks for the pointer! 
I have the he_IL.pbl built and ready. Before I open the PR, where should I host the .pbl file? 
Should I upload it to binaries.rebble.io/lp/ (and if so, how do I get access?), or is a raw GitHub URL acceptable like in #121?

PR #959, opened 2026-03-10
fw/comm/ble/ancs: fix resubscription failing with BTErrnoInvalidState

jplexer:
When iOS silently drops GATT notification subscriptions (e.g. during PPoGATT resets), the alive-check timer detects ANCS is unresponsive and attempts to resubscribe. However, the local GATTClientSubscriptionNode still records the subscription as active, causing
gatt_client_subscriptions_subscribe() to return BTErrnoInvalidState without re-writing the CCCD to the remote device.

Fix by unsubscribing first to clear the local state before resubscribing, ensuring the CCCD write is sent to iOS.

Fixes FIRM-1412


gmarull (approved):
hard to tell if this is an adequate solution without more context, but let's merge and see

PR #976, opened 2026-03-12
fw/comm_session: coalesce receiver callbacks to prevent BT deadlock

jplexer:
A deadlock occurred between KernelMain (holding bt_lock, blocked on full system task queue) and KernelBG (sole queue consumer, blocked on bt_lock) during BT reconnect floods.

Implement the callback coalescing suggested by the existing TODO: instead of scheduling a new system_task_add_callback per completed Pebble Protocol message, maintain pending lists and batch-drain them in a single callback per task type. This reduces queue usage from N entries to at most 2, preventing the overflow that caused the deadlock.

Fixes FIRM-1405

### DIFF
--- a/src/libutil/includes/util/slist.h
+++ b/src/libutil/includes/util/slist.h
@@ -1,0 +1,94 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+#include "order.h"
+
+typedef struct SingleListNode {
+  struct SingleListNode *next;
+} SingleListNode;
+
+typedef bool (*SingleListFilterCallback)(SingleListNode *found_node, void *data);
+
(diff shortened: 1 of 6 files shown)


gmarull (commented):
doesn't pebbleos have a linked list primitive?

src/libutil/includes/util/slist.h
```diff
@@ -0,0 +1,94 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+#include <stdbool.h>
+#include <stdint.h>
+#include "order.h"
+
+typedef struct __attribute__((packed)) SingleListNode {
```
gmarull: why does this need to be packed?

jplexer: copied from ListNode, probably not needed

gmarull: fix this, otherwise lgtm

gmarull (approved):

PR #980, opened 2026-03-16
set sensor_bpm_stable once received offwrist event

Jiyun-Yang:
Treat sensor as stable status when received the offwrist event, otherwise hrm_manager would delay ~20s to report the offwrist status to subscribers when GH3x2x recognized the offwrist event.

### DIFF
--- a/src/fw/services/common/hrm/hrm_manager.c
+++ b/src/fw/services/common/hrm/hrm_manager.c
@@ -224,6 +224,10 @@ static bool prv_is_sensor_bpm_stable(const HRMData *data) {
     // Once we receive at least one good reading, we are stable
     s_manager_state.sensor_bpm_stable = true;
     s_manager_state.off_wrist_when_stable = false;
+  } else if (quality == HRMQuality_OffWrist) {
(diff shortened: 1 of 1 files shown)


src/fw/services/common/hrm/hrm_manager.c
```diff
@@ -224,6 +224,10 @@ static bool prv_is_sensor_bpm_stable(const HRMData *data) {
     // Once we receive at least one good reading, we are stable
     s_manager_state.sensor_bpm_stable = true;
     s_manager_state.off_wrist_when_stable = false;
+  } else if (quality == HRMQuality_OffWrist) {
+    // Once received off wrist event, we are stable
+    s_manager_state.sensor_bpm_stable = true;
+    s_manager_state.off_wrist_when_stable = true;
```
gmarull, line 230: I'll actually review the whole logic here, because the `else` branch will handle the off-wrist case as well. We may just ignore the whole "stable" concept...

gmarull: ALso because the current impl never reports certain values like no signal

PR #985, opened 2026-03-16
Misc fixes

jplexer:


src/fw/services/common/clock.c
```diff
@@ -414,7 +414,16 @@ static void prv_watch_dst(void* user) {
   }
 }
 
+// Minimum valid time: January 1, 2010 00:00:00 UTC
```
gmarull, line 417: maybe 2020?

src/fw/services/normal/services_normal.c
```diff
@@ -43,9 +43,9 @@
 
 #include "util/size.h"
 
-// Minimum valid time: January 1, 2010 00:00:00 UTC (timestamp: 1262304000)
+// Minimum valid time: January 1, 2020 00:00:00 UTC (timestamp: 1577836800)
 // This represents the minimum time we consider valid for activity initialization
-#define MIN_VALID_TIME_TIMESTAMP 1262304000
+#define MIN_VALID_TIME_TIMESTAMP 1577836800
```
gmarull, line 48: squash

src/fw/services/common/clock.c
jplexer: overall valid system time should be further in the past than the valid service time

src/fw/services/normal/services_normal.c
jplexer: I would consider this a different change than clock

gmarull (approved):

PR #995, opened 2026-03-19
analyze_coredump: handle GDB errors without aborting script

jplexer:
GDB --batch mode aborts the entire sourced command file on the first error. When analyzing corrupted coredumps, commands like 'thread apply all info frame' can fail with inaccessible memory, which silently prevents all subsequent commands (pbl heap, pbl lockstats, pbl metadata) from ever running.

Wrap each GDB command in a python try/except so errors are printed but execution continues. Also add a backtrace limit to prevent infinite frame chasing on corrupted stacks, and a 5 minute timeout as a safety net.

### DIFF
--- a/tools/analyze_coredump.py
+++ b/tools/analyze_coredump.py
@@ -42,10 +42,19 @@ class CoredumpAnalyzer:
 
     def create_gdb_script(self):
         """Create a GDB script with commands to extract relevant information."""
-        commands = [
+        # GDB --batch aborts on errors in sourced command files (e.g. if a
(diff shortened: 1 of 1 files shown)


tools/analyze_coredump.py
```diff
@@ -135,6 +165,7 @@ def analyze(self):
                 capture_output=True,
                 text=True,
                 check=False,
+                timeout=300,
```
gmarull: probably too much? It usually takes a few seconds, maybe 60?

PR #997, opened 2026-03-19
activity: stop HRM sampling early when watch is flat or off-wrist

jplexer:
Previously, the flat-on-table check only ran at the toggle boundary (every ~10 minutes), so if the watch was placed flat during an active 60-second sampling window, the green HRM LED would keep shining until the window expired. Similarly, when the GH3X2X sensor reported HRMQuality_OffWrist, sampling continued for the full window.

Move the flat check earlier so it is also evaluated during active sampling, and track off-wrist events from the sensor to stop sampling immediately when either condition is detected. This saves battery and avoids a bright green LED at night when the watch is on a table.

Fixes FIRM-1523

### DIFF
--- a/src/fw/services/normal/activity/activity.c
+++ b/src/fw/services/normal/activity/activity.c
@@ -83,7 +83,11 @@ static void prv_heart_rate_subscription_update(uint32_t now_ts) {
   const uint32_t last_toggled_ts = s_activity_state.hr.toggled_sampling_at_ts;
 
   // Check if the watch is face up or face down (flat = likely off wrist)
-  const uint8_t z_axis = s_activity_state.last_orientation >> 4;
+  // Use the real-time orientation from the algorithm when available, otherwise fall back to the
+  // cached value from the last minute handler call.
+  uint8_t orientation = s_activity_state.last_orientation;
+  activity_algorithm_get_orientation(&orientation);
+  const uint8_t z_axis = orientation >> 4;
   const bool watch_is_flat = z_axis == 0 || z_axis == 8;
 
   bool should_toggle = false;
--- a/src/fw/services/normal/activity/activity.c
+++ b/src/fw/services/normal/activity/activity.c
@@ -82,19 +82,28 @@ static void prv_heart_rate_subscription_update(uint32_t now_ts) {
 
   const uint32_t last_toggled_ts = s_activity_state.hr.toggled_sampling_at_ts;
 
+  // Check if the watch is face up or face down (flat = likely off wrist)
+  const uint8_t z_axis = s_activity_state.last_orientation >> 4;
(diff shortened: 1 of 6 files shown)


src/fw/services/normal/activity/activity.c
```diff
@@ -146,9 +162,13 @@ T_STATIC void prv_hrm_subscription_cb(PebbleHRMEvent *hrm_event, void *context)
...
     // Perform a basic validity check so we only proceed with reasonable data
-    // TODO: Use quality to filter out some readings,
-    // TODO PBL-40784: Use HRMQuality_OffWrist as a special case to slow down the HRM subscription
+    // TODO: Use quality to filter out some readings
     bool valid_hr_reading = true;
+
+    // Track off-wrist state so we can stop sampling early
+    if (hrm_event->bpm.quality == HRMQuality_OffWrist) {
+      s_activity_state.hr.off_wrist_detected = true;
+    }
```
gmarull: sensor usually starts reporting offwrist, until it has certaininity it's onwrist, is this handled well?

gmarull: I'd probably not use off-wrist to stop the sensor but movement/position instead. There can be momentary off-wrist events, and that's useful to discard bogus samples, but from my (limited) experience, we could be turning off the sensor when we should not... But I may be wrong!

gmarull: To save power, we could add some heuristics like, if we have N periods where reported quality is consistently off-wrist, increase the sampling period (from 10 min to maybe 60)

src/fw/services/normal/activity/activity.c
```diff
@@ -82,19 +82,28 @@ static void prv_heart_rate_subscription_update(uint32_t now_ts) {
...
+  const uint8_t z_axis = s_activity_state.last_orientation >> 4;
+  const bool watch_is_flat = z_axis == 0 || z_axis == 8;
+
   bool should_toggle = false;
   if (s_activity_state.hr.currently_sampling) {
     // If we are currently sampling, turn off when:
     // - We reach the end of our maximum time on, ACTIVITY_DEFAULT_HR_ON_TIME_SEC
     // - We get ACTIVITY_MIN_NUM_GOOD_SAMPLES_SHORT_CIRCUIT good quality samples
     // - We get ACTIVITY_MIN_NUM_EXCELLENT_SAMPLES_SHORT_CIRCUIT excellent quality samples
+    // - The watch is flat (likely off wrist, no point wasting battery)
```
gmarull, line 95: I think we should just make sure HRM does not start if the watch is flat; the fact it goes flat during a sampling is not going to make much difference, TBH, and adds yet more logic to care about.

PR #999, opened 2026-03-20
do hw reset when touch IC cannot communicate with MCU

Jiyun-Yang:
CST816 may hung up by ESD, so reset it when it cannot communicate with MCU.

### DIFF
--- a/src/fw/drivers/touch/cst816/cst816.c
+++ b/src/fw/drivers/touch/cst816/cst816.c
@@ -51,6 +51,7 @@ static PebbleMutex *s_i2c_lock;
 
 static void prv_exti_cb(bool *should_context_switch);
 static void cst816_hw_reset(void);
+void touch_sensor_set_enabled(bool enabled);
 
 static bool prv_read_data(uint16_t register_address, uint8_t *result, uint16_t size, bool is_work_mode) {
   mutex_lock(s_i2c_lock);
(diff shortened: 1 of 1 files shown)


src/fw/drivers/touch/cst816/cst816.c
```diff
@@ -51,6 +51,7 @@ static PebbleMutex *s_i2c_lock;
 
 static void prv_exti_cb(bool *should_context_switch);
 static void cst816_hw_reset(void);
+void touch_sensor_set_enabled(bool enabled);
```
gmarull, line 54: should not be needed

src/fw/drivers/touch/cst816/cst816.c
```diff
@@ -251,14 +252,17 @@ void touch_sensor_init(void) {
 static void prv_process_pending_messages(void* context) {
   s_callback_scheduled = false;
 
+  const uint64_t current_time_ms = ticks_to_milliseconds(rtc_get_ticks());
   uint8_t data[CST816_TOUCH_DATA_SIZE] = {0};
   bool rv = prv_read_data(CST816_TOUCH_DATA_REG, data, CST816_TOUCH_DATA_SIZE, 1);
   if (!rv) {
-    PBL_LOG_ERR("Failed to read touch data, dropping event");
+    PBL_LOG_ERR("get touch data error, reset the hw");
+    touch_handle_update(0, TouchState_FingerUp, NULL, 0, current_time_ms);
```
gmarull, line 260: what if there was no down event before?

src/fw/drivers/touch/cst816/cst816.c
```diff
@@ -251,14 +252,17 @@ void touch_sensor_init(void) {
 static void prv_process_pending_messages(void* context) {
   s_callback_scheduled = false;
 
+  const uint64_t current_time_ms = ticks_to_milliseconds(rtc_get_ticks());
   uint8_t data[CST816_TOUCH_DATA_SIZE] = {0};
   bool rv = prv_read_data(CST816_TOUCH_DATA_REG, data, CST816_TOUCH_DATA_SIZE, 1);
   if (!rv) {
-    PBL_LOG_ERR("Failed to read touch data, dropping event");
+    PBL_LOG_ERR("get touch data error, reset the hw");
```
gmarull:
```suggestion
    PBL_LOG_ERR("Failed to read touch data, trying to recover");
```

src/fw/drivers/touch/cst816/cst816.c
```diff
@@ -251,14 +252,17 @@ void touch_sensor_init(void) {
...
+    touch_handle_update(0, TouchState_FingerUp, NULL, 0, current_time_ms);
+    exti_disable(CST816->int_exti);
```
gmarull, line 261: is this needed?

src/fw/drivers/touch/cst816/cst816.c
Jiyun-Yang: when it happens on touch down status, the upper layer may not know the hardware been reset with default FingerUp status, so force report the FingerUp event to upper layer.

src/fw/drivers/touch/cst816/cst816.c
Jiyun-Yang: When do hw reset, don't know if the INT pin is stable or not, so force disable interrupt during reset cycle.

src/fw/drivers/touch/cst816/cst816.c
Jiyun-Yang: this function is called before implementation, so declare it here.

gmarull:
#1407

PR #1006, opened 2026-03-20
fw/services/common/accel_manager: implement accel_manager_enable

jplexer:
accel_manager_enable() was a no-op stub, meaning the accelerometer hardware was never powered down when the runlevel dropped to BareMinimum (e.g. during factory reset). This caused an event storm as the driver kept firing sample, shake, and double-tap callbacks into queues that were being torn down, potentially freezing the device on reboot.

Implement accel_manager_enable() to call accel_power_down/up and add early-return guards in the driver callbacks to ignore events when disabled.

Fixes FIRM-1304

### DIFF
--- a/src/fw/services/common/accel_manager.c
+++ b/src/fw/services/common/accel_manager.c
@@ -85,6 +85,10 @@ static uint8_t s_buffer_storage[200 * sizeof(AccelManagerBufferData)];
 
 static uint64_t s_last_empty_timestamp_ms = 0;
 
+//! Whether the accel manager is enabled. When disabled (e.g. during factory reset via
+//! RunLevel_BareMinimum), the accelerometer hardware is powered down and callbacks are ignored.
+static bool s_enabled = true;
+
 // Accel Idle
 #define ACCEL_MAX_IDLE_DELTA 100
 static bool s_is_idle = false;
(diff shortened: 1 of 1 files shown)


src/fw/services/common/accel_manager.c
```diff
@@ -608,7 +612,25 @@ DEFINE_SYSCALL(bool, sys_accel_manager_consume_samples,
...
+  if (on && !prev) {
+    prv_update_driver_config();
+    if (s_shake_subscribers_count > 0) {
+      accel_enable_shake_detection(true);
+    }
+    if (s_double_tap_subscribers_count > 0) {
+      accel_enable_double_tap_detection(true);
+    }
+  } else if (!on && prev) {
+    accel_set_num_samples(0);
```
gmarull, line 628: accel_set_sampling_interval(0) ?

jplexer: accel_set_sampling_interval(0) means "give me the shortest interval possible" which would be wrong

accel_set_num_samples(0) explicitly means "the accelerometer driver must not call accel_cb_new_sample()"

gmarull: Hmm, interface does not claim this:

```
//! Sets the accelerometer sampling interval.
//!
//! Not all sampling intervals are supported by all drivers. The driver must
//! select a sampling interval which is equal to or shorter than the requested
//! interval, saturating at the shortest interval supported by the hardware.
//!
//! The new sampling rate takes effect immediately. The driver may flush any
//! queued samples before changing the sampling rate to ensure that timestamps
//! remain accurate.
//!
//! @param interval_us The requested sampling interval in microseconds.
//!
//! @return The actual sampling interval set by the driver.
uint32_t accel_set_sampling_interval(uint32_t interval_us);
```

Both active drivers (LIS2DW12 and LSM6DSO will enter power down if given 0 here...)

gmarull: `accel_set_num_samples(0)` will effectively disable FIFO, but not ser ODR to "0"

gmarull: If apps rely on this behavior, drivers need to be fixed, and interface properly specify this.

jplexer: ok I gathered that from the context wrong, amended commit

gmarull (approved):

PR #1008, opened 2026-03-22
fw/settings: add HR monitoring interval setting

gmarull:
Add a configurable HRM measurement frequency to Health settings, allowing users to choose between 10min, 30min, 1h, or disabled. The setting is only shown on platforms with a built-in HRM (CAPABILITY_HAS_BUILTIN_HRM). When disabled, periodic HR sampling is stopped entirely.


jplexer (approved):
heh sorry!

PR #1023, opened 2026-03-24
Improve frozen SDK generation and add standalone build script

jplexer:
- Add tools/build_sdk.py, a standalone script to generate SDK files for one or more platforms without requiring a full waf configure/build cycle
- Improve frozen SDK support: platforms with FROZEN_AT_REVISION now generate proper stub macros for APIs added after their freeze point, and skip libpebble.a generation (since toolchain differences make reproducible builds impractical, we just copy from an older sdk)
- Fix PBL_IF_SMARTSTRAP_ELSE to be conditional on PBL_SMARTSTRAP since older fw does actually still support it


tools/build_sdk.py
```diff
@@ -0,0 +1,187 @@
+#!/usr/bin/env python
+# SPDX-FileCopyrightText: 2024 Google LLC
```
gmarull: ooops

jplexer: thanks claude

gmarull (approved):

PR #1027, opened 2026-03-25
fw/memfault: right-size coredump storage allocation [FIRM-1537]

jplexer:
Instead of always allocating a fixed 32KB buffer for the RAM-backed coredump storage, compute the exact serialized size using memfault_coredump_get_save_size() and allocate only what's needed.

### DIFF
--- a/third_party/memfault/port/src/memfault_platform_coredump_storage.c
+++ b/third_party/memfault/port/src/memfault_platform_coredump_storage.c
@@ -21,21 +21,26 @@
 #include "memfault_pebble_coredump.h"
 
 #include "kernel/pbl_malloc.h"
+#include "util/math.h"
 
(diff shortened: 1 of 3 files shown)


third_party/memfault/port/src/memfault_platform_coredump_storage.c
```diff
@@ -22,20 +22,24 @@
...
 static size_t s_coredump_storage_size;
 
-void memfault_coredump_storage_alloc(void) {
+void memfault_coredump_storage_alloc(size_t size) {
   if (s_coredump_storage != NULL) {
     return;  // Already allocated
   }
-  s_coredump_storage = kernel_zalloc(COREDUMP_STORAGE_SIZE);
+  // Round up to sector size for the SDK's erase operation
+  size = (size + COREDUMP_SECTOR_SIZE - 1) & ~(COREDUMP_SECTOR_SIZE - 1);
```
gmarull: ROUND_TO_MOD_CEIL?

gmarull (approved):

PR #1030, opened 2026-03-26
fw: fix build errors when test apps are enabled

jplexer:
Fix several -Werror warnings that prevented building with --test_apps:

- timeline_pins_demo: move fallthrough comment outside braces so GCC recognizes it
- event_service: increase snprintf buffer size for int range
- menu: increase snprintf buffer size for format string output
- movable_line: increase snprintf buffer size for int16_t range
- temperature_demo: use memmove instead of memcpy for overlapping buffers
- panic_window: guard unused code with CAPABILITY_HAS_HARDWARE_PANIC_SCREEN

### DIFF
--- a/src/fw/apps/core/panic_window.c
+++ b/src/fw/apps/core/panic_window.c
@@ -20,6 +20,7 @@
 
 #include <stdio.h>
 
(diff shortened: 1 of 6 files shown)


gmarull (approved):
Can we enable these in CI?

PR #1033, opened 2026-03-26
fw/activity: add setting to toggle HR tracking during activities

gmarull:
The Kraepelin activity detection algorithm subscribes to continuous HRM when a walk/run is detected (after 3 minutes of activity). Add a new "HR During Activity" toggle in Health settings to allow disabling this behavior, saving battery during walks/runs.


jplexer (approved):
im gonna disregard that license check since kraepelin is mit ;)

PR #1050, opened 2026-03-30
fw/applib/animation: reduce render frame rate on sf32lb52

jplexer:
Add a separate ANIMATION_FRAME_RENDER_INTERVAL_MS that controls how often frames are rendered, while keeping ANIMATION_TARGET_FRAME_INTERVAL_MS for animation duration calculations.

On SF32LB52 at 144 MHz, target 20 Hz instead of 30 Hz to give each frame more CPU time and reduce jitter.


gmarull:
this may be made dynamic, and can now be tested with the powermode service

PR #1056, opened 2026-04-01
Add logs to Memfault coredumps via compact logs [FIRM-1569]

jplexer:
- Enable Memfault compact log system alongside existing PebbleOS log hashing, so each PBL_LOG call emits both a hashed log entry (for serial/flash) and a ~5-10 byte CBOR compact log entry into a 2KB RAM circular buffer

- On crash, the NMI handler captures all RAM (including the log buffer) to SPI flash; after reboot, coredump reconstruction includes the log buffer regions so the Memfault cloud can decode and display pre-crash logs

- Add log_fmt linker section at 0xF0000000 to hold compact log format strings in the ELF (non-loaded, like .log_strings)

- Switch firmware C standard from c11 to gnu11 to support the ##__VA_ARGS__ comma-elision required by Memfault's compact log macros

### DIFF
--- a/waftools/pebble_arm_gcc.py
+++ b/waftools/pebble_arm_gcc.py
@@ -150,10 +150,17 @@ def configure(conf):
     conf.load("gcc")
     conf.load("asm")
 
+    # Memfault compact logs require GNU ##__VA_ARGS__ comma-elision;
+    # use -std=gnu11 when Memfault is enabled, otherwise -std=c11.
+    c_std = (
+        "-std=gnu11"
+        if getattr(conf.env, "memfault_needs_gnu_extensions", False)
+        else "-std=c11"
+    )
     conf.env.append_value(
         "CFLAGS",
         [
-            "-std=c11",
+            c_std,
         ],
     )
 
--- a/src/fw/popups/bluetooth_pairing_ui.c
+++ b/src/fw/popups/bluetooth_pairing_ui.c
@@ -609,7 +609,7 @@ void bluetooth_pairing_ui_handle_event(PebbleBluetoothPairEvent *event) {
         prv_handle_pairing_complete(event->success);
       } else {
         PBL_LOG_ERR("Got complete event for unknown process %p vs %p",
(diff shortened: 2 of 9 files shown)


waftools/pebble_arm_gcc.py
```diff
@@ -153,7 +153,7 @@ def configure(conf):
     conf.env.append_value(
         "CFLAGS",
         [
-            "-std=c11",
+            "-std=gnu11",
```
gmarull: that's scary, why is this really needed? Can we confirm Memfault really requires this?

jplexer: Memfault's compact log feature has a hard compile-time assertion that ##\_\_VA_ARGS\_\_ comma elision works. GCC only supports this in -std=gnu* modes, not -std=c* modes. This is explicitly documented by Memfault in their own source (lines 36-37 of preprocessor.h: "When using GCC this means you should be compiling with -std=gnu11"). There's no way around it without patching the SDK.

gmarull: can GNU11 only be enabled on their lib, not globally? or at least only if memfault is enabled?

jplexer: Unfortunately we can't scope it to just the Memfault lib.

That said, I've scoped it so gnu11 is only enabled when Memfault is enabled (non-Memfault builds stay on c11)

src/fw/fw_common.ld
```diff
@@ -202,6 +202,14 @@ SECTIONS {
         KEEP(*(.log_strings))
     } >LOG_STRINGS
 
+    /* Memfault compact log format strings. Not loaded to memory. */
+    log_fmt 0xF0000000 (INFO) :
```
gmarull: we should avoid modifying core linker file for a third_party, optional module, can this be done in a separate file?

jplexer: done

gmarull (approved):

PR #1071, opened 2026-04-04
fw/apps/system: Add music settings for volume controls and progress bar

asyba:
Add "Show Volume Controls" and "Show Progress Bar" options with the mobile app watch settings.

When "Show Volume Controls" is hidden:
UP/DOWN buttons perform next/previous track.
The SELECT button long-press is disabled to prevent entering volume mode.

When "Show Progress Bar" is hidden:
The progress bar and time labels are hidden.
The app unsubscribes from the tick timer to save power.

core app PR: https://github.com/coredevices/mobileapp/pull/148

refactor of previous work: https://github.com/coredevices/PebbleOS/pull/525

### DIFF
--- a/src/fw/apps/system/music.c
+++ b/src/fw/apps/system/music.c
@@ -18,6 +18,7 @@
 #include "services/common/i18n/i18n.h"
 #include "services/normal/music.h"
 #include "services/normal/vibes/vibe_score.h"
+#include "shell/prefs.h"
 #include "shell/system_theme.h"
 #include "process_management/app_manager.h"
 #include "process_state/app_state/app_state.h"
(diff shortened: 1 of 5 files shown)


gmarull (commented):
feels like it would be nice to introduce at some point per-app settings, that can be controller in a menu (pretty much like in iOS), but for now lgtm

jplexer (approved):

ericmigi:
Cool, just to confirm, this is default off and the setting only shows up on the phone side, right?

asyba:
@ericmigi yes for default its always showing and can only be change from mobile app.
_static bool s_music_show_volume_controls = true;
static bool s_music_show_progress_bar = true;_

ericmigi (approved):

PR #1073, opened 2026-04-06
fw/comm/ble: auto-recover from wedged BLE controller

ericmigi:
## Summary

- Detect when BLE advertising repeatedly fails (10 consecutive failures = 10 seconds) and automatically trigger a full BLE stack reset including LCPU power cycle
- Add `ble_transport_ll_deinit()`/`reinit()` to power-cycle the BLE controller hardware, wired into `bt_driver_stop()`/`start()` so `bt_ctl_reset_bluetooth()` now does a full hardware reset
- Add test infrastructure for advertising enable failure injection

## Problem

When the BLE controller (LCPU) becomes unresponsive, the advertising scheduler enters a tight retry loop that pegs the CPU at 100%, draining the battery from full to dead in ~8 hours. The existing `bt_ctl_reset_bluetooth()` only did a NimBLE host-level reset without power-cycling the LCPU, so the wedged controller stayed wedged.

## Fix

1. **`hci_sf32lb52.c`**: Add transport teardown/reinit that closes the IPC queue and power-cycles the LCPU via `lcpu_power_off()`/`lcpu_power_on()`
2. **`init.c`**: Wire transport deinit/reinit into `bt_driver_stop()`/`start()` 
3. **`gap_le_advert.c`**: Add consecutive failure counter → after 10 failures, stop timer + call `bt_ctl_reset_bluetooth()`

## Test plan

- [ ] Verify new `test_gap_le_advert__enable_failure_triggers_reset` test passes
- [ ] Verify existing advertising tests still pass
- [ ] Test on getafix hardware: toggle airplane mode to verify BLE stop/start with LCPU power cycle works correctly
- [ ] Soak test: verify no regressions in normal BLE operation

Fixes FIRM-1602

### DIFF
--- a/src/bluetooth-fw/nimble/init.c
+++ b/src/bluetooth-fw/nimble/init.c
@@ -27,6 +27,10 @@ static const uint32_t s_bt_stack_start_stop_timeout_ms = 3000;
 
 extern void pebble_pairing_service_init(void);
 extern void nimble_discover_init(void);
+#ifdef BT_CONTROLLER_SF32LB52
+extern void ble_transport_ll_deinit(void);
+extern void ble_transport_ll_reinit(void);
+#endif
 
 #if NIMBLE_CFG_CONTROLLER
 static TaskHandle_t s_ll_task_handle;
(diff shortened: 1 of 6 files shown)


gmarull (changes requested):
we need to review some memory pool sizing, the workaround was merged as an attempt, seems it is not enough, the real fix is to redimension some of the pools (or add analytics to track usage and find a good number)

gmarull:
ref https://github.com/coredevices/PebbleOS/pull/1129

PR #1074, opened 2026-04-06
fw/applib/tick_timer_service: fix spurious handler calls on timer skew

ericmigi:
## Summary

- Fix spurious tick handler calls caused by FreeRTOS timer double-fires
- The `(units_changed == 0)` condition in `do_handle()` was intended for the first tick after subscribe, but also fires on any subsequent tick where no time component changed (e.g. timer skew, RTC adjustment)
- For `MINUTE_UNIT` subscribers (watchfaces), this causes visible bugs: animations firing multiple times per minute instead of once
- Fix separates the first-tick path from the normal path: first tick always fires, subsequent ticks only fire when a subscribed unit actually changed

Related: FIRM-1377 — the `regular_timer.c` fix in PR #838 addressed missed minute callbacks, but didn't fix spurious extra calls at the `tick_timer_service` layer.

## Test plan

- [x] Verified no handler in the codebase checks `units_changed == 0` (all 24+ callers ignore the parameter)
- [ ] Test MINUTE_UNIT watchface — should animate exactly once per minute, not on timer double-fires
- [ ] Test SECOND_UNIT app (e.g. music timer) — should still fire every second
- [ ] Test first-tick behavior — handler should fire immediately on `tick_timer_service_subscribe()`

### DIFF
--- a/src/fw/applib/tick_timer_service.c
+++ b/src/fw/applib/tick_timer_service.c
@@ -44,7 +44,9 @@ static void do_handle(PebbleEvent *e, void *context) {
   struct tm currtime;
   sys_localtime_r(&e->clock_tick.tick_time, &currtime);
 
-  if (!state->first_tick) {
+  bool is_first = state->first_tick;
+
+  if (!is_first) {
     if (state->last_time.tm_sec != currtime.tm_sec) {
       units_changed |= SECOND_UNIT;
     }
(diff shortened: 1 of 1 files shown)


ericmigi:
@gmarull ran into this while building my star wars watchface - I noticed that the minute handler was firing multiple times per minute.

gmarull (commented):
Not sure how this fixes the problem... The real problem is that we're using two different time sources that are not in sync (system timer to fire timers, and RTC time to compare and decide what to do here).

PR #1083, opened 2026-04-09
Use RTC to generate tick events

gmarull:
Note that RTC is not used as an RTOS "wake-up alarm" because some issues are observed like longer than expected wake-up (in discussion with SiFli). Because of that, RTC/RTOS are not in sync, causing issues. However, it is a good idea to have a clear split between RTC-based timers (like, e.g., CLOCK_REALTIME in POSIX) and others that just follow OS tick rate (regular timer API does not offer any guarantees about RTC time...)

Fixes FIRM-1377

### DIFF
--- a/src/fw/drivers/rtc.h
+++ b/src/fw/drivers/rtc.h
@@ -100,6 +100,22 @@ RtcTicks rtc_alarm_get_elapsed_ticks(void);
 //!     us from going into stop mode before we're ready to wake up from it.
 bool rtc_alarm_is_initialized(void);
 
+// RTC Wake-up
+///////////////////////////////////////////////////////////////////////////////
+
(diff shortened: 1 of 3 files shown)


jplexer (approved):
peak

ericmigi:
ready to merge?

gmarull:
> ready to merge?

The main issue was actually due to https://github.com/coredevices/PebbleOS/pull/1084. With this PR, what you get is _perfectly_ aligned tick events and an actual RTC clock, but in reality we do not need that. That also comes at a cost of extra RTC wake-ups (because we're no longer relying on OS ticks). I'm waiting for SiFli feedback as well because it seems the need for other PR is not well understood/documented.

PR #1090, opened 2026-04-13
Make analytics multi-backend, add native solution

gmarull:
This transforms the analytics service into multi-backend capable. It also adds a native backend option that does not rely on third-party services. Compared to the old native implementation, here we should not rely on any fixed blob layout. Instead, the ELF file can be used to obtain the actual blob layout/fields/sizes, etc. (see the last commit with a tool that shows how to do that).

Needs some testing...!


ericmigi:
just updated eng-dash to match this format. @sjp4 can you add this to mobile pls?

src/fw/services/common/analytics/native.c
```diff
@@ -0,0 +1,350 @@
...
+
+static PebbleMutex *s_mutex;
+static DataLoggingSession *s_dls_session;
+
+void pbl_analytics__native_init(void) {
+  Uuid system_uuid = UUID_SYSTEM;
+
+  PBL_ASSERTN(dls_initialized());
+
+  s_dls_session = dls_create(DlsSystemTagAnalyticsDeviceHeartbeat, DATA_LOGGING_BYTE_ARRAY,
```
sjp4: Should we use a different/new tag for this? The existing tag will be reporting from legacy firmwares with a different format, so maybe we should separate this format

gmarull: Yes, I'm gonna fix this

gmarull:
> just updated eng-dash to match this format. @sjp4 can you add this to mobile pls?

The new mechanism should work as follows:

1. Add a new eng-dash endpoint where we can start uploading ELF files; this way, you can parse (see example tool): (1) the build ID and (2) the blob layout
2. Mobile is the _dumb_ part: it should just get the blob and post it to the server
3. The server can now match the blob ID with any previously uploaded ELF ID. If there's a match, simply take data following blob layout.

This makes sure the backend will always store new metrics without any changes, and that non-released builds will be ignored (no ELF match).

sjp4:
> just updated eng-dash to match this format. @sjp4 can you add this to mobile pls?

see mobile branch `steve/analytics` if you want to try it - i tried testing with this fw branch on asterix but couldn't get it to boot (not sure if it's my setup or... but time for bed right now)

gmarull:
> > just updated eng-dash to match this format. @sjp4 can you add this to mobile pls?
> 
> see mobile branch `steve/analytics` if you want to try it - i tried testing with this fw branch on asterix but couldn't get it to boot (not sure if it's my setup or... but time for bed right now)

It should boot now, force pushed a few mins ago

ericmigi:
Plan is to have GitHub CI upload the elf?

ericmigi:
Problem is I don't want to have it auto add new columns in bigquery if we can help it...so we'll still need a manual step somewhere (the new data will be stored in the raw field so we don't lose it. )

gmarull:
> Problem is I don't want to have it auto add new columns in bigquery if we can help it...so we'll still need a manual step somewhere (the new data will be stored in the raw field so we don't lose it. )

We can enable fields we want, but we should process all we get and store it (same as memfault really). We could have a place where to select which detected metrics we want to track, give them a name, etc. It's a more robust approach IMHO, things are decoupled which is great. The blob also contains scaling factors.

gmarull:
> Plan is to have GitHub CI upload the elf?

Yes, same as we do for Memfault

ericmigi:
> We can enable fields we want, but we should process all we get and store it (same as memfault really). We could have a place where to select which detected metrics we want to track, give them a name, etc. It's a more robust approach IMHO, things are decoupled which is great. The blob also contains scaling factors.

it's all tracked in a raw field so we don't lose any data but we'll need to add column/to the metrics viewer - it won't do it automatically. I think we're saying the same things.

gmarull:
> 
> 
> > We can enable fields we want, but we should process all we get and store it (same as memfault really). We could have a place where to select which detected metrics we want to track, give them a name, etc. It's a more robust approach IMHO, things are decoupled which is great. The blob also contains scaling factors.
> 
> 
> 
> it's all tracked in a raw field so we don't lose any data but we'll need to add column/to the metrics viewer - it won't do it automatically. I think we're saying the same things. 

That works as soon as old raw can be post processed when new column is added. I think Memfault just inserts columns automatically and adds data, so you can immediately plot/track whatever you want. But you can also archive a metric, and customize a few aspects of it. IMHO Memfault does the right thing here, it's just not great in their inspection capabilities (plots, stats, etc.)...

ericmigi:
ok it's merged and deployed to eng-dash, can you give me a sample file or something that I can test the ingester with?

gmarull:
> ok it's merged and deployed to eng-dash, can you give me a sample file or something that I can test the ingester with? 

Just build this branch and take tintin_fw.elf

src/fw/console/prompt_commands.h
```diff
@@ -654,6 +658,10 @@ static const Command s_prompt_commands[] = {
   { "mflt metrics_dump", command_mflt_metrics_dump, 0 },
   { "mflt device_info", command_mflt_device_info, 0 },
 #endif  // MEMFAULT
+
+#if ANALYTICS_NATIVE
+  { "analytics native metrics_dump", command_analytics_native_metrics_dump, 0 },
```
sjp4, line 664: can you also add a command to send a heartbeat to the phone, to make testing easier?

third_party/memfault/port/include/memfault_platform_config.h
```diff
@@ -36,7 +36,4 @@
 // slightly shorter length here
 #define MEMFAULT_DATA_EXPORT_CHUNK_MAX_LEN 50
 
-#define MEMFAULT_REBOOT_REASON_CUSTOM_ENABLE 1
-
-#define MEMFAULT_METRICS_BATTERY_ENABLE 1
```
sjp4, line 41: Does this mean we're not setting the "special" memfault metrics for battery/connectivity now?

sjp4 (approved):

third_party/memfault/port/include/memfault_platform_config.h
gmarull: we do, but the logic is within PebbleOS, so all backends benefit from it.

src/fw/console/prompt_commands.h
gmarull: good point, let me add that

PR #1092, opened 2026-04-14
fw/applib: redraw watchface immediately on 24h format change

ericmigi:
## Summary
- When the 24h clock format setting changes, watchfaces now re-render immediately instead of waiting up to 60s for the next minute tick
- `prv_set_s_clock_24h` fires a `PEBBLE_TICK_EVENT` when the value actually changes
- `tick_timer_service` tracks the 24h format state and forces the tick handler to fire when it detects a change, regardless of what time units changed

## Root cause
`prv_set_s_clock_24h()` in `prefs.c` only updated the static variable without notifying anyone. Watchfaces like TicToc (BW) call `clock_is_24h_style()` in their MINUTE_UNIT tick handler, so they wouldn't pick up the change until the next minute boundary — up to 60 seconds later.

## Test plan
- [x] Change 24h format via phone app Settings — watchface should update within 1 second
- [x] Change via on-watch Settings > Time > 24h toggle — same immediate update
- [x] Verify watchfaces subscribing to MINUTE_UNIT still get the update
- [x] Verify no extra redraws when setting doesn't actually change

Fixes https://linear.app/core-dev/issue/FIRM-1624

### DIFF
--- a/src/fw/applib/tick_timer_service.c
+++ b/src/fw/applib/tick_timer_service.c
@@ -7,6 +7,7 @@
 #include "event_service_client.h"
 #include "process_management/app_manager.h"
 
+#include "pbl/services/common/clock.h"
 #include "pbl/services/common/event_service.h"
 #include "pbl/services/common/tick_timer.h"
 #include "kernel/events.h"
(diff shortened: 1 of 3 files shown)


ericmigi:
## QEMU test: immediate watchface redraw on 24h toggle

Built for `spalding_gabbro --qemu`, installed "Real Weather" digital watchface, toggled 12h/24h via `pebble emu-time-format` while the watchface was showing.

**Before this fix:** watchface would stay in the old format until the user navigated away and back.
**After this fix:** watchface updates immediately on each toggle.

24h toggle fix demo

jplexer (approved):

gmarull:
@ericmigi needs rebase

PR #1098, opened 2026-04-14
Update Moddable to 8.0.0

jplexer:
+ Fix Stub for Touch until pulled upstream


third_party/moddable/wscript
```diff
@@ -185,6 +185,7 @@ def build(bld):
         'moddable/build/devices/pebble/modules/accelerometer/pebble-accelerometer.c',
         'moddable/build/devices/pebble/modules/battery/pebble-battery.c',
         'moddable/build/devices/pebble/modules/compass/pebble-compass.c',
+        'moddable/build/devices/pebble/modules/touch/pebble-touch.c',
```
gmarull: maybe this should be added only if we have touch available?

jplexer: sadly doesnt work since the main moddable file depends on that file
i know I know

gmarull (approved):

PR #1102, opened 2026-04-15
fw/drivers/mic/sf32lb52: statically allocate PDM buffers

jplexer:
The PDM circular buffer (20 KB) and DMA rx buffer (1152 B) were being kernel_malloc'd on every mic_start. On obelix the ~93 KB kernel heap is prone to fragmentation, and the 20 KB contiguous allocation started failing reliably in the field once commit 486b2860 bumped the circular buffer from 128 ms to 320 ms. Every voice session then aborted at pdm.c:97 ("Failed to allocate circular buffer storage") before any audio was captured.

Both buffers are fixed-size, exist once per device, and are only used while the mic is running. Move them to .bss, drop the alloc/free helpers, and remove the now-dead NULL guards. This also takes ~21 KB of pressure off the kernel heap for everything else.

Fixes FIRM-1662

### DIFF
--- a/src/fw/drivers/mic/sf32lb52/pdm.c
+++ b/src/fw/drivers/mic/sf32lb52/pdm.c
@@ -4,7 +4,6 @@
 #include "drivers/mic.h"
 #include "drivers/pmic/npm1300.h"
 #include "board/board.h"
-#include "kernel/pbl_malloc.h"
 #include "system/logging.h"
 #include "os/mutex.h"
 #include "system/passert.h"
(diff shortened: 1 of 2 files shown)


src/fw/drivers/mic/sf32lb52/pdm.c
```diff
@@ -36,6 +35,15 @@
...
 static MicDeviceState* s_state;
 
+// Static backing storage for the PDM buffers. Previously these were allocated
+// from the kernel heap on every mic_start, which could fail once the heap
+// became fragmented. The buffers are a fixed size, exist once per device, and
+// are only live while the mic is running, so .bss is fine.
+static uint8_t s_circ_buffer_storage[PDM_CIRCULAR_BUF_SIZE_BYTES]
+    __attribute__((aligned(4)));
+static int16_t s_pdm_rx_buffer[PDM_CH_COUNT * PDM_AUDIO_RECORD_PIPE_SIZE]
+    __attribute__((aligned(4)));
```
gmarull, line 45: we should likely move buffers to upper layers, that can use whatever they want (e.g. static if always on, heap to have it allocated for a period, etc.) But we have quite a few other problems to solve :-) ...

gmarull (approved):

PR #1106, opened 2026-04-15
fw/services/normal: instrument storage mutex and settings compaction

jplexer:
Add temporary timing instrumentation to help diagnose multi-second KernelMain UI stalls reported in FIRM-1649. Two known triggers (a QuickView event arriving from the phone, saving an Alarm) both reduce to "something wrote a settings/BlobDB file while a flash-heavy operation was running", and the suspected root cause is settings_file_compact() running under TimelineItemStorage->mutex while ftl_erase_sector_blocking() ties up the flash bus for several seconds. flash_erase_sector_blocking() feeds the task watchdog internally, which explains why the user sees a recoverable freeze rather than a reboot.

This patch wires up two pieces of telemetry to confirm or kill the hypothesis:

- timeline_item_storage.c: wrap every mutex_lock/unlock pair with prv_storage_lock/prv_storage_unlock helpers that measure both the wait-for-lock time and the hold time, and PBL_LOG_WARNING when either crosses 100 ms. The log line includes __func__ and the storage name so we know which operation stalled and on which database (pin_db, reminder_db, ...).

- settings_file.c: log entry and exit of settings_file_rewrite_filtered with elapsed ms and final status. This covers both the auto-trigger via settings_file_compact() from settings_file_set() and the explicit rewrite path used by timeline_item_storage_flush().

If the hypothesis is right we expect to see triples like:

  FIRM-1649: settings_file_rewrite_filtered start file=pin_db
  FIRM-1649: settings_file_rewrite_filtered end file=pin_db
              elapsed=4823ms status=0
  FIRM-1649: timeline_item_storage_read waited 4810ms for pin_db mutex

Refs FIRM-1649

### DIFF
--- a/src/fw/services/normal/blob_db/timeline_item_storage.c
+++ b/src/fw/services/normal/blob_db/timeline_item_storage.c
@@ -3,12 +3,42 @@
 
 #include "pbl/services/normal/blob_db/timeline_item_storage.h"
 
+#include "drivers/rtc.h"
 #include "kernel/pbl_malloc.h"
 #include "pbl/services/normal/filesystem/pfs.h"
 #include "pbl/services/normal/settings/settings_raw_iter.h"
 #include "system/logging.h"
 
 #define MAX_CHILDREN_PER_PIN 3
+
+// FIRM-1649: temporary instrumentation to catch long mutex hold/wait on the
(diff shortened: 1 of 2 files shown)


gmarull (approved):
we should likely add something like analytics events for these kind of problems

PR #1114, opened 2026-04-17
Tweak motion sensitivity defaults

gmarull:


mattnovelli:
This would be great. Find myself flicking my wrist pretty hard on my PT2.

jplexer (approved):

ericmigi:
> This would be great. Find myself flicking my wrist pretty hard on my PT2.

@mattnovelli Please report a bug through the app if you're having any issues that send us additional information. In this case, we're actually making it less sensitive, so it would help if you could report the bug.

mattnovelli:
Oh I see. Might be worth a setting then. Regardless, loving the PT2 experience so far.

ericmigi:
already is! Pebble app ->  enable debug mode -> Watch Settings -> Display -> Motion sensitivity

PR #1118, opened 2026-04-17
Adjust watchfaces for Gabbro

hellcp:
Adds gabbro variant of the tictoc spalding:
tictoc-gabbro
The faces are meant to pay tribute to PTR default faces and their bezels

Adds resources and temporary config for possible timezone variants (they will require further work to add configuration to configure timezones):
tictoc-gabbro-timezones
And general tictoc spalding refactor to accomodate those new features

Additionally a fix to the kickstart values, to display the face correctly


jplexer (approved):
Love it! But please split kickstart change into it's own commit

gmarull (approved):

PR #1127, opened 2026-04-20
RGB backlight: qemu_emery driver + SDK color API

jplexer:


src/fw/board/boards/board_obelix.c
```diff
@@ -570,6 +571,7 @@ const BoardConfig BOARD_CONFIG = {
   .ambient_light_dark_threshold = 150,
   .ambient_k_delta_threshold = 25,
   .dynamic_backlight_min_threshold = 5,
+  .backlight_default_color = LED_WARM_WHITE,
```
gmarull, line 574: I think we can drop this from board config and assume LED_WARM_WHITE is the default

include/pbl/services/common/light.h
```diff
@@ -46,6 +46,25 @@ void light_enable_interaction(void);
...
+
+//! Push a modal color preempt. While the refcount is non-zero, the LED is
+//! forced to the user default color even if an app has set an override.
+//! Used by notifications and other modals so they display in neutral white
+//! without permanently clearing the underlying app's color.
+void light_push_modal_color_preempt(void);
+
+//! Pop a modal color preempt. When the refcount returns to zero, any app
+//! override is re-applied.
+void light_pop_modal_color_preempt(void);
```
gmarull: I'd name `light_system_color_request|release`

include/pbl/services/common/light.h
```diff
@@ -46,6 +46,25 @@ void light_enable_interaction(void);
...
 void light_reset_user_controlled(void);
 
+//! @copydoc app_light_set_color
+//! Argb is a GColor8-packed byte (2 bits per channel, alpha ignored).
+//! No-op on platforms without a color backlight.
+void light_set_color(uint8_t argb);
+
+//! @copydoc app_light_reset_color
+//! No-op on platforms without a color backlight.
+void light_reset_color(void);
```
gmarull: `light_system_color_set` (or `force`)

src/fw/applib/app_light.c
```diff
@@ -10,3 +12,11 @@ void app_light_enable_interaction(void) {
 void app_light_enable(bool enable) {
   sys_light_enable(enable);
 }
+
+void app_light_set_color(GColor color) {
+  sys_light_set_color(color.argb);
+}
+
+void app_light_reset_color(void) {
```
gmarull: `set_system_color`?

src/fw/board/boards/board_obelix.c
jplexer: for QEMU, warm white looks like this, which is weird imo, which is why I made the default color for qemu just LED_WHITE
image

gmarull: ok, makes sense then

gmarull (approved):

PR #1133, opened 2026-04-22
fw/kernel/event_loop: require double-tap for touch backlight

ericmigi:
## Summary

The touch-to-backlight setting (added in c51621e9) activated the backlight on any `Touchdown`, which made accidental screen contact light the display. Require two `Touchdown` events within 400 ticks (~400 ms) to trigger, following the same elapsed-ticks pattern as the back-button quickpress coredump trigger a few lines above.

Fixes [FIRM-1705](https://linear.app/core-dev/issue/FIRM-1705/touch-feature-activates-backlight-on-single-tap-should-be-double-tap).

## Test plan

- [x] Builds cleanly for `qemu_emery` with `CONFIG_TOUCH=y`
- [ ] On device: single tap no longer lights the backlight
- [ ] On device: two taps within ~400 ms lights the backlight, respecting DND
- [ ] On device: two taps spaced > 400 ms apart do not trigger

### DIFF
--- a/src/fw/kernel/event_loop.c
+++ b/src/fw/kernel/event_loop.c
@@ -95,6 +95,10 @@ static RtcTicks s_back_quickpress_last = 0;
 static int s_back_quickpress_count = 0;
 #endif
 
+static const uint32_t TOUCH_DOUBLE_TAP_INTERVAL_TICKS = 400;
+static RtcTicks s_touch_last = 0;
+static int s_touch_count = 0;
+
 void launcher_task_add_callback(void (*callback)(void *data), void *data) {
   PebbleEvent event = {
     .type = PEBBLE_CALLBACK_EVENT,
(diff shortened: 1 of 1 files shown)


ericmigi:
@gmarull feel free to use if ok or delete

gmarull (changes requested):
hw supports double tap among other gestures (in theory, not tested). In any case, we need a centralized source of "gestures".

ericmigi:
ok I'd just like to have double tap integrated before we ship this feature so that people don't get used to single tapping to turn on backlight

gmarull:
> ok I'd just like to have double tap integrated before we ship this feature so that people don't get used to single tapping to turn on backlight

let me try implementing gestures, should not be too much work

ericmigi:
sounds good!

machinaeZER0:
Could single vs double be an option, if this moves forward in the future? I quite like single-tap, and I don't think I'll accidentally trigger it nearly as often as I was triggering motion activation. That said, I've only been using it for a few hours, and maybe it's more sensitive to accidental presses from non-skin items than I realize?

PR #1135, opened 2026-04-22
fw/asterix: add DA7212 audio driver for speaker playback + raise PMIC discharge limit during audio

jplexer:
nRF5 I2S driver implementing drivers/audio.h for P2D. Codec runs as I2S master (nRF as slave) so the DA7212 PLL can synthesize an exact 16 kHz WCLK from our 4 MHz MCK
nRF-as-master can't divide 32 MHz to exactly 16 kHz.

also adds callbacks to the audio driver to raise pmic limit at start + reset pmic limit at end

### DIFF
--- a/src/fw/drivers/audio.h
+++ b/src/fw/drivers/audio.h
@@ -1,8 +1,20 @@
 /* SPDX-FileCopyrightText: 2025 Core Devices LLC */
 /* SPDX-License-Identifier: Apache-2.0 */
 
+#pragma once
+
+#include <stdint.h>
+
 typedef const struct AudioDevice AudioDevice;
 typedef void (*AudioTransCB)(uint32_t *free_size);
+
+//! Optional board-level power hooks. Either callback may be NULL.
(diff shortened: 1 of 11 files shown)


src/fw/drivers/audio.h
```diff
@@ -1,9 +1,22 @@
...
 typedef void (*AudioTransCB)(uint32_t *free_size);
 
+//! Optional board-level power callbacks for audio playback.
+//! Platform AudioDevice structs should include these fields.
+//! power_up is called before the PA is enabled (e.g. raise PMIC discharge limit).
+//! power_down is called after the PA is disabled (e.g. restore PMIC discharge limit).
+//! Either may be NULL if no board-level power management is needed.
+#define AUDIO_DEVICE_POWER_FIELDS \
+  void (*power_up)(void);         \
+  void (*power_down)(void);
```
gmarull: avoid using macros for this purpose, maybe define a struct with ops that can be used in other places

gmarull (approved):

PR #1137, opened 2026-04-22
third_party: update moddable to 8.0.1

jplexer:
mostly a formality


gmarull (commented):
Let's keep using our fork, because we have cyclic dependencies, and every time we make a breaking change, we need to patch moddable in sync. I'd do the following:

git tag `rebase-DDMMYYYY` current `pebbleos` HEAD in our fork (so it's not garbage collected), then, force-push `pebbleos` to our new desired HEAD from upstream and just pull it here.

That was a fairly common practice I used in Zephyr/NCS.

PR #1138, opened 2026-04-22
shell/battery_ui_fsm: skip first warning when next one is imminent

jplexer:
If the battery crosses a warning threshold already close to the next one (e.g. from a fast drop or a fuel-gauge correction), we can show a gray "Powered 'til tomorrow" and then immediately a red "Powered 'til tonight" — individually correct, but contradictory to the user because the daypart message is computed from different reference times.

Require a minimum of 3h of headroom above the next warning threshold before firing the current warning. If the headroom isn't there, skip the earlier warning and let the next one fire normally with an accurate daypart message.

Fixes FIRM-1703


gmarull (approved):
we need a good rework in the battery FSM :-) ...

PR #1141, opened 2026-04-22
applib/light: add light_set_color_rgb888() for full 8-bit-per-channel tint

jplexer:
I was not realizing people would want full RGB control until I started writing the docs ._.


gmarull (approved):
ok, even though our controller doesn't have such resolution IIRC

jplexer:
i mean hey, at least its better than just 64 colors!

PR #1143, opened 2026-04-23
Speaker API

jplexer:
Exposes the speaker to apps. The API supports:

  - `speaker_play_notes` — play a melody as a sequence of notes (sine/square/triangle/saw)
  - `speaker_play_tracks` — mix up to 4 monophonic tracks for polyphony
  - `speaker_play_tone` — one-shot tone convenience wrapper
  - `speaker_stream_{open,write,close}` — raw PCM streaming for app-generated audio
  - `speaker_stop`, `speaker_set_volume`, `speaker_get_status`, `speaker_set_finish_callback`

One drive-by: `evented_timer` now compensates for QEMU TCG callback drift, which was making note timing audibly wrong under emulation.


gmarull (commented):
my only concern is the last commit, it feels a bit dirty, and pollutes service code with target dependent stuff. Isn't there any better solution?

jplexer:
> my only concern is the last commit, it feels a bit dirty, and pollutes service code with target dependent stuff. Isn't there any better solution?

happy to drop that one, its only really needed for an edgecase that wont be that common probably anyway

gmarull (approved):

PR #1148, opened 2026-04-23
fw/asterix: reduce DA7212 line gain to 0dB

jplexer:
The speaker service was configuring DA7212_LINE_GAIN to 0x3a (~+10dB), while the manufacturing speaker test uses 0x30 (0dB). Running the line amp that hot overdrives the watch speaker: the board noticeably warms up during playback

Drop the line gain to 0x30 to match the mfg test's 0dB setting.


jplexer:
merging, as this is a pretty important *hot*fix

PR #1152, opened 2026-04-24
fw/kernel/event_loop: force backlight on for app-subscribed touches

jplexer:
When an app subscribes to touch events, ignore the global wake-on-touch preference and tie the backlight to the touch itself: forced on while a finger is down, then timed out after liftoff. Skip the gesture path in this case so double-tap no longer triggers a redundant wake.


gmarull (approved):
good catch

PR #1156, opened 2026-04-25
fw/services/battery: add optional 80% charge limit to extend battery lifespan

### LINKED ISSUE #722: Ability to limit battery charge to X%

### Proposed feature request

Lithium batteries degrade much faster, the closer they are to the 100% or 0%. With these watches, this is made even worse, because, after being fully charged, the watch will potentially spend days being close to the 100%.

To make this problem better, there could be an option to limit the charge to a smaller level (usually 80%, but a configurable level would be even better). When the watch would reach that level, it would stop charging and just hold it.

Alternatively, if the charging circuit does not allow holding the level, it could just turn charging off and on again, so it would oscillate between 79% - 81%.

### Alternatives

_No response_

### Anything else?

_No response_

spr4bhu:
This PR adds a software charge limit feature that stops charging at 80% and resumes at 77% (3% hysteresis) to reduce lithium battery degradation from sustained high charge levels. The feature is exposed as a user-facing toggle in System settings, off by default.

Changes:

-  Add `battery_charge_limit` module that calls `battery_set_charge_enable()` to disable/re-enable charging based on battery percentage thresholds; includes a 60s rate limit to prevent rapid charger cycling and a 60s periodic timer to cover gaps in fuel gauge event delivery when charging is disabled
-  Add `charge_limit_enabled` pref (persistent on/off setting via `SettingsFile`, default false) with getter/setter following the existing `stationary_mode_enabled` pattern
- Add "Charge Limit (80%)" toggle row in System settings menu, mirroring the existing "Stand-By Mode" toggle
- Hook `battery_charge_limit_evaluate()` into `battery_monitor_handle_state_change_event()` and `battery_charge_limit_init()` into `battery_monitor_init()`
- Register `battery_charge_limit.c` in `services/battery/wscript_build`
  
  Resolves #722

### DIFF
--- a/include/pbl/services/battery/battery_charge_limit.h
+++ b/include/pbl/services/battery/battery_charge_limit.h
@@ -1,0 +1,15 @@
+/* SPDX-FileCopyrightText: 2025 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include "pbl/services/battery/battery_state.h"
+
+// The battery charge limit service optionally stops charging at 80% and resumes at 77%
+// to reduce battery degradation from sustained high charge levels.
+
+void battery_charge_limit_init(void);
+
+void battery_charge_limit_evaluate(PreciseBatteryChargeState state);
+
+bool battery_charge_limit_is_active(void);
--- a/include/pbl/services/battery/battery_charge_limit.h
+++ b/include/pbl/services/battery/battery_charge_limit.h
@@ -1,0 +1,16 @@
+/* SPDX-FileCopyrightText: 2026 Shashvat Prabhu */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include "pbl/services/battery/battery_state.h"
+
+// The battery charge limit service optionally stops charging at 80% to reduce battery
+// degradation from sustained high charge levels. Charging resumes if the level drops back
+// below 80%.
+
+void battery_charge_limit_init(void);
+
+void battery_charge_limit_evaluate(PreciseBatteryChargeState state);
+
+bool battery_charge_limit_is_active(void);
--- a/src/fw/services/battery/battery_charge_limit.c
+++ b/src/fw/services/battery/battery_charge_limit.c
@@ -4,20 +4,16 @@
 #include "pbl/services/battery/battery_charge_limit.h"
 
 #include "drivers/battery.h"
-#include "drivers/rtc.h"
 #include "pbl/services/regular_timer.h"
 #include "shell/prefs.h"
 #include "system/logging.h"
 
 #define CHARGE_LIMIT_PCT 80
-#define CHARGE_RESUME_PCT 77
-#define MIN_TOGGLE_INTERVAL_S 60
 #define PERIODIC_CHECK_INTERVAL_S 60
 
 ////////////////////////
 // State
 T_STATIC bool s_limit_active;
-T_STATIC RtcTicks s_last_toggle_ticks;
 static RegularTimerInfo s_periodic_timer;
 
 static void prv_periodic_timer_cb(void *data) {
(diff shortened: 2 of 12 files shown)


EspressoTini:
Amazing, i have been meaning to circle back on this.  Thank you so much!

ericmigi:
Awesome. Thanks for the PR! Need to add this behind one of the features that only enables the setting to be shown on the phone side. We're reducing how many settings are shown on the Watch UI and encouraging people to add the setting on the phone side. 

@jplexer do you have an example that we can show to folks on how to add this on both sides?

jplexer:
fw side it needs to be added to the settings blob db whitelist (like: https://github.com/coredevices/PebbleOS/commit/4d8abd99f1a244190689abae82a6ba2db10686ad)

On the mobile side I do not know, @sjp4 definitely knows more there

src/fw/services/battery/battery_charge_limit.c
```diff
@@ -0,0 +1,77 @@
...
+#include "pbl/services/battery/battery_charge_limit.h"
+
+#include "drivers/battery.h"
+#include "drivers/rtc.h"
+#include "pbl/services/regular_timer.h"
+#include "shell/prefs.h"
+#include "system/logging.h"
+
+#define CHARGE_LIMIT_PCT 80
+#define CHARGE_RESUME_PCT 77
```
gmarull: While plugged in, if not charging, the watch will run from USB power, so the battery will not drain. I guess you can simply disable the charger and forget about any hysteresis.

spr4bhu:
@sjp4 can I have the format on how the features are to be added on phone side

src/fw/services/battery/battery_charge_limit.c
spr4bhu: Made the changes, thanks.

Flexximilian:
> While plugged in, if not charging, the watch will run from USB power, so the battery will not drain. I guess you can simply disable the charger and forget about any hysteresis.

So then if the watch runs on USB power directly when plugged in, what's the use of a charge limit in the first place? The only reason to do that, then is to avoid topping up if the watch comes on and off the charger repeatedly, correct? That is something a user can simply avoid.

But even if someone would put the watch on the charger say, every night, out of habit. Would it not be smarter then, to have a setting that goes like this: Do not charge to 100% (or some max level) unless you've been discharged to 80% (or some min level) or lower since being at or over the max level. 

That way you do allow a full charge, but not a repeated trickle charge at a high level (which is unnecessary if 80% gives multi-day operation).

I am not a battery expert, but charging a battery to 100% is not so much the problem if I understand it correctly, it's keeping it there constantly via trickle charging, or discharging fully and leaving it there for any amount of time. I also heard that the best would be to keep Li-Ion between 40%-ish and 60%-ish at all times for maximum lifespan. So, I'd love a function where the watch simple alerts when it wants to be charged for maximum lifespan (via a notification) and beeps every 5 minutes for 5 - 10 seconds when it wants to come off the charger (or it simply ends charging and runs on the USB power) at 60%, again optimizing for battery health instead of charge frequency. I'd rather charge once a week for 10 years, than once every three weeks for 5 years...

So I'd a good default and recommendation from a vendor and the freedom to override it. And while I bought two Time 2s to have a spare, I certainly would love the battery to last as long as possible.

include/pbl/services/battery/battery_charge_limit.h
```diff
@@ -0,0 +1,15 @@
+/* SPDX-FileCopyrightText: 2025 Core Devices LLC */
```
jplexer: wrong copyright text (you wrote it)

jplexer (commented):
squash and got that one nit

caco3:
> We're reducing how many settings are shown on the Watch UI and encouraging people to add the setting on the phone side.

But battery management / extension IMO should be a configuration directly accessible on the watch itself.

adipascu:
Thanks @spr4bhu for the original patch here, it was a great starting point. I picked it up and tried to address the review feedback in #1841, with the companion app side in coredevices/mobileapp#318. Happy to adjust anything, or to step back if you would rather continue this one.

PR #1164, opened 2026-04-26
docs: update action bar width for platforms

jamsinclair:
The action bar width is no longer always 30px. Learned from experience 😅 . 

Would be awesome to get this updated in the docs.

See the current values in the SDK.:

https://github.com/coredevices/PebbleOS/blob/d7ef6199a70462cc0154103b965006c60d8eddf1/src/fw/applib/ui/action_bar_layer.h#L88-L95


jamsinclair:
Is it possible to squash merge this commit so it uses the correct commit structure? (should inherit from the PR title)

Otherwise happy to do this myself, I made this via the GH web ui (hence the number of commits).

jplexer (approved):

PR #1174, opened 2026-04-27
fw/apps/system: add timeout other places

asyba:
* Added the same main timeout of 30 seconds from main menu to other stock apps: Notifications, Settings, Watchfaces and Weather. Its enable by default.

If we need to make them optional, I can add a mobile settings side.


jplexer:
I dont think we should do this

asyba:
> I dont think we should do this

I wanted to have it, for example,if we leave it open in those pre-installed apps and forget to close it, it will be there forever, until I realize it and when I check the time I notice that watchface isn't showing.

PR #1185, opened 2026-04-28
src/fw/kernel: Add HRM struct version capability bit

sjp4:
Needed so that the mobile app can safely send the ActivityHRMSettings
 struct to the watch on different firmware versions.


sjp4:
Actually - because we didn't add the capability at the same time as the struct change this would break enabling/disabling health until users are all on the version with the capability, so I'll hard-code the FW versions of the changes into the app instead (in the future we need to consider this when changing a struct which the phone can update)

PR #1189, opened 2026-04-29
fw/drivers/display/sf32lb: add update-timeout watchdog

jplexer:
The JDI display driver gates all compositor rendering on a single `s_updating` flag that is only ever cleared in the LCDC transfer-complete callback. If the LCDC fails to deliver that callback — e.g. on the SiFli HAL's ICB-overflow path which sets HAL_LCDC_ERROR_OVERFLOW without invoking XferCpltCallback — the flag stays true forever and the compositor silently stops rendering, even though the rest of the system keeps working.

Add a 200 ms NewTimer-based watchdog: armed in `display_update`, cancelled in `prv_display_update_terminate`, and on timeout it logs the latched HAL ErrorCode, clears it, and posts a PEBBLE_CALLBACK_EVENT so the finalize path runs on KernelMain (matching the IRQ-driven dispatch and keeping compositor state single-threaded). A test-and-clear of `s_updating` ensures only one of the IRQ and watchdog paths runs the recovery body when they race.

Fixes FIRM-1740

### DIFF
--- a/src/fw/console/prompt_commands.c
+++ b/src/fw/console/prompt_commands.c
@@ -1129,6 +1129,18 @@ void command_audit_delay_us(void) {
   __enable_irq();
 }
 
+#if !defined(RELEASE) && (PLATFORM_OBELIX || PLATFORM_GETAFIX)
+#include "drivers/display/sf32lb/display_jdi.h"
+
+// Arms the JDI display driver to drop the next LCDC transfer-complete callback,
+// simulating the silent-DMA failure mode (e.g. ICB overflow). The display
+// update watchdog should fire ~200ms later and recover.
+void command_display_drop_complete(void) {
+  display_jdi_test_drop_next_complete();
+  prompt_send_response("display: armed drop of next LCDC complete; watch for watchdog log");
+}
+#endif
+
 #if !MICRO_FAMILY_SF32LB52
 // Simply parks the chip permanently in stop mode in whatever state it's currently in. This can be
 // pretty handy when trying to profile power of the chip under certains states
--- a/src/fw/console/prompt_commands.h
+++ b/src/fw/console/prompt_commands.h
@@ -293,6 +293,10 @@ extern void command_console_disable_rx(const char *seconds_str);
 extern void command_force_deepwfi(const char *arg);
 #endif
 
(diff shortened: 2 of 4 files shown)


gmarull (commented):
that sounds hacky, has it been verified?

jplexer:
Temporarily added a prompt command that simulates the issue
Without this fix we experience the same issue as the bug (i.e. input works but no screen updates), while with this we recover in a split second

gmarull:
> Temporarily added a prompt command that simulates the issue Without this fix we experience the same issue as the bug (i.e. input works but no screen updates), while with this we recover in a split second

Can you add reproduction code, so we can ask Sifli to take a look before merging?

jplexer:
I tried but couldn't get reliable repro code working from our side. `__HAL_LCDC_LOCK` does while (1) on a stuck lock and WaitBusy is a non-yielding busy-spin, so once the LCDC is in a bad state any HAL call from a task wedges that task and trips a different watchdog.

Best pointer for SiFli (and seems like a bug itself?): `HAL_LCDC_IRQHandler` (bf0_hal_lcdc.c ~5354) sets `HAL_LCDC_ERROR_OVERFLOW` on `LCD_IF_IRQ_ICB_OF_RAW_STAT` but never calls `XferCpltCallback` / `XferErrorCallback`, so the application silently loses the completion. That matches the symptoms in FIRM-1740 / FIRM-1676 / MOB-6633.

To at least prove the watchdog recovers, I added `display drop_complete` which skips one `event_put_isr` from `HAL_LCDC_SendLayerDataCpltCbk` (same lost-completion symptom). Watchdog fires within ~200ms, logs the timeout, next compositor frame redraws. Works on hardware.

gmarull:
@rabbitsaviola can you please take a look? Note, we recently pulled newer HAL, could that be related?

rabbitsaviola:
@jplexer please dump LCDC register and all ram data when it's reproduced, so we can analyze it offline. 
LCDC register address:  0x50008000, size: 0x8128 bytes. 
RAM start address: 0x20000000, size: 512KByte
It's prefered that lcdc register and ram data are saved in separate binary, e.g. lcdc_reg.bin and hcpu_ram.bin. And please attach fw.elf and fw.map also, thanks.

PR #1190, opened 2026-04-29
kernel/util: fix UX and PFS race in phone-triggered factory reset

jplexer:
Fixes FIRM-1743


gmarull (commented):
is it feasible to maybe go to runlevel 0 to execute this op?

jplexer:
> is it feasible to maybe go to runlevel 0 to execute this op?

we already do

gmarull:
> > is it feasible to maybe go to runlevel 0 to execute this op?
> 
> we already do

is commit message wrong then?

```
When a factory reset is triggered via the phone protocol, the watch
becomes a zombie for ~47s while filesystem_regions_erase_all() runs:
BLE is torn down (looking like an "unpair") but the App task keeps
running, so the user can navigate menus and interact with apps with no
indication that a reset is in progress.
```

jplexer:
this is the symptom that claudio apparently experienced both on asterix and obelix! when I tried it with an app that app still ran but I couldnt interact with it (so wasnt able to repro fully)

gmarull:
What I miss e.g. is, why if we have:

```
  {
    .set_enable_fn = bt_ctl_set_enabled,
    .enable_mask = R_FirmwareUpdate | R_Normal,
  },
```

but in `prv_factory_reset_non_pfs_data`, we run `services_set_runlevel(RunLevel_BareMinimum);`, phone gets the Bluetooth prompts :/

gmarull:
```
This only addresses the UX half of the issue. The App task still runs
during the erase, so writes to PFS during the reset window can still
race (e.g. airplane mode persisting across the reset, as reported).
```

so we still miss a proper fix?

jplexer:
> so we still miss a proper fix?

done by 4686429342ebafb08c8b9c5b673fe4d7d7c8e55c, but that was done seperately from claude doing the first one so isnt reflected in its commit message

src/fw/kernel/util/factory_reset.c
```diff
@@ -73,17 +73,20 @@ void factory_reset(bool should_shutdown) {
 
   prv_factory_reset_non_pfs_data();
 
-  // TODO: wipe the registry on tintin?
-  filesystem_regions_erase_all();
-
 #if !defined(RECOVERY_FW)
+  // pfs_format() holds the PFS mutex across the erase, blocking concurrent
+  // writes from the App task that would otherwise survive into a freshly-erased region.
+  pfs_format(false /* write_erase_headers */);
```
gmarull, line 79: still, if app is running, after this there could still be a chance of a new PFS op?

src/fw/kernel/util/fw_reset.c
```diff
@@ -35,6 +41,23 @@ typedef enum {
...
+  Dialog *base = simple_dialog_get_dialog(dialog);
+  dialog_set_text(base, i18n_get("Resetting...", dialog));
+  dialog_set_icon(base, RESOURCE_ID_GENERIC_WARNING_LARGE);
+  dialog_set_timeout(base, DIALOG_TIMEOUT_INFINITE);
+  simple_dialog_set_buttons_enabled(dialog, false);
+  simple_dialog_set_icon_animated(dialog, false);
+  i18n_free_all(dialog);
+
+  simple_dialog_push(dialog, modal_manager_get_window_stack(ModalPriorityCritical));
+}
```
gmarull: wondering if a cleaner approach, when triggered from mobile, would be to follow the updates model (like progress_ui app)

jplexer: good call!

src/fw/kernel/util/factory_reset.c
jplexer: we immediately reset afterwards so no

gmarull (approved):

PR #1197, opened 2026-04-30
Raise 4K storage

jplexer:
Took over the branch and folded in some review-driven follow-ups. Three commits now, in this order:

  1. **`fw/services/normal/settings: add growable settings files`** _(was `cf6e09f46`, expanded)_
  2. **`fw/services/normal/persist: raise per-app storage limit to 1 MiB`** _(unchanged content; reordered to land second)_
  3. **`fw/services/normal/settings: speed up file rewrite, pause watchdog`** _(new)_
  
  Reordered for bisectability (need growable files before raising the limits), made growable files shirnk on compact (before it would stay at what it grew to)
  Big grows hit the task watchdog, thats fixed with the third commit
  There are some files that currently allocate for the worst case (appglancedb ~100KB, notifstr ~30KB, weatherdh ~40KB, etc), we can switch those to growable in a later PR.

Fixes FIRM-1655

### DIFF
--- a/include/pbl/services/settings/settings_file.h
+++ b/include/pbl/services/settings/settings_file.h
@@ -41,6 +41,16 @@ typedef struct SettingsFile {
   //! fail.
   int max_used_space;
 
+  //! The current allocation budget for physical file size. For growable files,
+  //! this starts at a small initial value and grows toward max_used_space on
+  //! demand. For non-growable files, this equals max_used_space.
+  int alloc_used_space;
+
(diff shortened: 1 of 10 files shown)


gmarull:
services no longer user 'normal' prefix, adjust title maybe?

jplexer:
> services no longer user 'normal' prefix, adjust title maybe?

done

gmarull (approved):

PR #1199, opened 2026-04-30
sdk: expose PERSIST_STORAGE_MAX_SIZE to apps

jplexer:
Apps had no way to know the per-app persist storage cap at compile time; the existing limit lived only in firmware as an internal define. Expose it as PERSIST_STORAGE_MAX_SIZE in applib/persist.h so apps can size buffers and allocations against it. The persist service now uses the same public define as its single source of truth.

Frozen-platform SDKs (aplite, basalt, chalk, diorite) ship pre-1 MiB firmware, so build_sdk.py rewrites the value to 4096 in their generated pebble.h / pebble_worker.h. Apps targeting older SDKs that don't define PERSIST_STORAGE_MAX_SIZE at all should fall back to 4096 themselves.

### DIFF
--- a/src/fw/applib/persist.h
+++ b/src/fw/applib/persist.h
@@ -35,7 +35,8 @@
 //! retrieve values from the phone, it provides you with a much faster way to restore state.
 //! In addition, it draws less power from the battery.
 //!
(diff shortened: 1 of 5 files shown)


jplexer:
actually no sorry this should be a runtime check

PR #1203, opened 2026-04-30
fw/drivers/mic/nrf5: shrink PDM circular buffer on OOM

jplexer:
Mirror the sf32lb52 fallback: when the kernel heap is fragmented, retry the 10 KB circular buffer allocation in 32 ms steps down to a 128 ms floor instead of failing dictation outright. 

Fixes FIRM-1737


gmarull (approved):
feels like we should manage one level above if we have same pattern on both platforms?

jplexer:
Yes but let's merge for now

PR #1224, opened 2026-05-05
Introduce PebbleOS SDK

gmarull:


docs/development/qemu.md
```diff
@@ -2,88 +2,12 @@
...
-   export PATH=$PWD:$PATH
-   ```
-
-   or, if you'd rather not shadow another `qemu-system-arm` already on your
-   `PATH`, set `PEBBLE_QEMU_BIN` to the binary's absolute path:
-
-   ```shell
-   export PEBBLE_QEMU_BIN=$PWD/qemu-system-arm
-   ```
+The QEMU binary ships with the [PebbleOS SDK](https://github.com/coredevices/PebbleOS-SDK).
```
jplexer, line 10: I would still include how to build from src

jplexer (approved):
small nitpick that I'd prefer but it's not necessary so approv d anyway

docs/development/qemu.md
gmarull: Can we include this in the QEMU repo maybe?

jplexer: sure

PR #1229, opened 2026-05-05
fw/touch: mirror touch coordinates in left-hand mode

jplexer:
In left-hand (display rotated 180°) mode, the framebuffer is mirrored but the touch sensor still reports physical coordinates, so taps land in the diagonally opposite location from where the user sees them.

Add a touch_sensor_set_rotated() driver hook that mirrors reported coordinates around max_x/max_y, and have the orientation manager toggle it alongside the other _set_rotated drivers. Populate max_x/max_y for obelix's CST816 panel from the display dimensions.

Fixes FIRM-1786

### DIFF
--- a/src/fw/services/orientation_manager/service.c
+++ b/src/fw/services/orientation_manager/service.c
@@ -11,6 +11,9 @@
 #include "drivers/imu/mmc5603nj/mmc5603nj.h"
 #include "kernel/events.h"
 #include "process_management/process_manager.h"
+#ifdef CONFIG_SERVICE_TOUCH
+#include "pbl/services/touch/touch.h"
+#endif
 
 
 void prv_change_orientation(bool rotated) {
@@ -18,6 +21,9 @@ void prv_change_orientation(bool rotated) {
   button_set_rotated(rotated);
   accel_set_rotated(rotated);
   mag_set_rotated(rotated);
+#ifdef CONFIG_SERVICE_TOUCH
+  touch_set_rotated(rotated);
+#endif
 }
 
 void orientation_handle_prefs_changed(void) {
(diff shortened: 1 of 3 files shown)


gmarull (commented):
we should not rotate at driver level, it's a generic operation, driver agnostic

src/fw/services/orientation_manager/service.c
```diff
@@ -11,13 +11,19 @@
 #include "drivers/imu/mmc5603nj/mmc5603nj.h"
 #include "kernel/events.h"
 #include "process_management/process_manager.h"
+#ifdef CONFIG_TOUCH
```
gmarull:
```suggestion
#ifdef CONFIG_SERVICE_TOUCH
```

src/fw/services/orientation_manager/service.c
```diff
@@ -11,13 +11,19 @@
...
+#include "pbl/services/touch/touch.h"
+#endif
 
 
 void prv_change_orientation(bool rotated) {
   display_set_rotated(rotated);
   button_set_rotated(rotated);
   accel_set_rotated(rotated);
   mag_set_rotated(rotated);
+#ifdef CONFIG_TOUCH
```
gmarull:
```suggestion
#ifdef CONFIG_SERVICE_TOUCH
```

src/fw/services/orientation_manager/service.c
```diff
@@ -11,13 +11,19 @@
...
+  touch_set_rotated(rotated, DISP_COLS, DISP_ROWS);
```
gmarull: DISP_COLS/ROWS are "globally visible", can be obviated?

gmarull: it's ok as is, at some point we'll cleanup all this anyway

jplexer: doesnt matter lol, claude already did its thing :P

gmarull (approved):

PR #1241, opened 2026-05-06
fw/kernel: don't trap divide-by-zero in user code

jplexer:
Fixes FIRM-1830


gmarull (commented):
hmm not sure if this is a good idea tbh...

jplexer:
> hmm not sure if this is a good idea tbh...

legacy watches didnt fault on divbyzero, asterix doesnt, legacy qemu doesnt, new qemu also doesnt
only outlier is obelix and getafix

gmarull:
> > hmm not sure if this is a good idea tbh...
> 
> legacy watches didnt fault on divbyzero, asterix doesnt, legacy qemu doesnt, new qemu also doesnt only outlier is obelix and getafix

ok, I didn't know, if that's the case, ok

gmarull (approved):

PR #1253, opened 2026-05-07
QEMU screenshot & skill

gmarull:


wscript
```diff
@@ -986,9 +984,8 @@ def qemu_launch(ctx):
         "-monitor unix:{mon_sock},server=on,wait=off "
         "-s "
         "-serial file:uart1.log "
-        "-serial tcp::12344,{serial} "   # Used for bluetooth data
```
jplexer, line 981: is not unused, used in conjunction with pebble-tool (i.e. pebble install --qemu)

jplexer (changes requested):

wscript
gmarull: so not for bluetooth data?

jplexer: yeah not really, just pebbleprotocol ig

jplexer (approved):

PR #1254, opened 2026-05-07
Add and expose backlight events

jplexer:


src/fw/applib/backlight_service.h
```diff
@@ -0,0 +1,41 @@
...
+//! @addtogroup Foundation
+//! @{
+//!   @addtogroup EventService
+//!   @{
+//!     @addtogroup BacklightService
+//!
+//! \brief Notifies your app when the backlight turns on or off.
+//!
+//! The BacklightService lets your app react to the screen waking and
+//! sleeping. The handler is invoked whenever the backlight transitions
```
gmarull: screen waking/sleeping may not be accurate terminology, outdoors (if ALS is used) this event will never be triggered.

gmarull (approved):

gmarull:
merging as I want to tag new release

PR #1256, opened 2026-05-07
nrf5/audio: fix phaser distortion from rapid play/pause [FIRM-1742]

jplexer:


src/fw/drivers/nrf5/audio/audio.c
```diff
@@ -297,19 +322,50 @@ void audio_init(AudioDevice *audio_device) {
...
+  if (s_pwr_state == AudioPwrWarm) {
+    // Codec, DAI, and I2S have been running continuously, streaming zeros
+    // from the ISR's auto-fill. Just resume writes and the next refill will
+    // land real audio.
+    state->trans_cb = cb;
+    state->callback_pending = false;
+    state->is_running = true;
+    s_pwr_state = AudioPwrActive;
+    mutex_unlock(s_audio_mutex);
+    PBL_LOG_INFO("Audio started (warm)");
```
gmarull: Let's avoid using info level to not spam logs

jplexer: claude keeps doing that!! maybe we should put something in the agents.md about this?

gmarull: if that's the case yes :D ...

src/fw/drivers/nrf5/audio/audio.c
```diff
@@ -422,26 +481,87 @@ void audio_stop(AudioDevice *audio_device) {
...
+    audio_device->power_ops->power_down();
+  }
+
+  s_active_device = NULL;
+  prv_free_buffers(state);
+
+  s_pwr_state = AudioPwrCold;
+  mutex_unlock(s_audio_mutex);
+
+  PBL_LOG_INFO("Audio fully stopped");
```
gmarull: oops

gmarull (approved):

src/fw/drivers/nrf5/audio/audio.c
jplexer: pahahahah i told it to not do that!

PR #1272, opened 2026-05-12
bluetooth-fw/nimble: track driver state and recover lost stop signal

jplexer:
Follow-up to FIRM-1837. The May 7 fix (#1247) covered `BLE_HS_EALREADY`, but `BLE_HS_EBUSY` still trips `ble_hs_event_start_stage2`'s assert.

**Root cause:** `bt_driver_start()`'s `ble_hs_is_enabled()` guard only catches vendor state `ON`. If a prior `ble_hs_stop()` times out, vendor state stays at `STOPPING`, the guard returns false, and the next start hits `BLE_HS_EBUSY` → assert → reboot.

**Fix:**
- Track local driver state (`Stopped` / `Starting` / `Started` / `Stopping`) so we don't start mid-stop.
- Drain stale `s_host_started` / `s_host_stopped` signals around each wait.
- `bt_driver_stop` mirrors the transitions, keeps `PBL_ASSERT` on timeout.
- On entry-guard timeout, probe with `ble_hs_stop(NULL, NULL, NULL)`: `BLE_HS_EALREADY` → vendor is `OFF`, recover; anything else → wedged, refuse. Probe is side-effect-free in `OFF` and `STOPPING`, unreachable while `ON`.

### DIFF
--- a/src/bluetooth-fw/nimble/init.c
+++ b/src/bluetooth-fw/nimble/init.c
@@ -41,6 +41,19 @@ static SemaphoreHandle_t s_host_stopped;
 static DisInfo s_dis_info;
 static struct ble_hs_stop_listener s_listener;
 
+typedef enum {
+  DriverStateStopped,
+  DriverStateStarting,
+  DriverStateStarted,
+  DriverStateStopping,
+} DriverState;
+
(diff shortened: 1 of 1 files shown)


gmarull:
squash?

gmarull (approved):

PR #1274, opened 2026-05-12
bluetooth-fw/nimble: fix discovery stop race causing KernelBG hang

gmarull:
`bt_driver_gatt_stop_discovery()` could block forever on `xSemaphoreTake(s_discovery_stopped, portMAX_DELAY)` when a discovery completed naturally between the in-progress check and the flag set. KernelBG would then miss its watchdog and the watch reset into PRF.

Race sequence:
 1. KernelBG enters stop, reads `s_discovery_in_progress = true`.
 2. NimBLE host task fires the last discovery callback. Its top check sees `s_stop_discovery_requested = false`, so it takes the natural completion path, sets `s_discovery_in_progress = false`, and returns without giving the semaphore.
 3. KernelBG resumes, sets `s_stop_discovery_requested = true`, and blocks on the semaphore forever.

Fix by setting the stop flag before checking in-progress and by giving the semaphore from every discovery termination point (new `prv_signal_discovery_done()` helper) whenever a stop is pending. Drain any stale signal at the start of stop. Also reset the stop flag and set in-progress before issuing `ble_gattc_disc_all_svcs()` in `bt_driver_gatt_start_discovery_range()` so callbacks on the NimBLE host task cannot observe stale flags and silently abort the new discovery, and free the context on failure.

Fixes FIRM-1895

### DIFF
--- a/src/bluetooth-fw/nimble/gatt_client_discovery.c
+++ b/src/bluetooth-fw/nimble/gatt_client_discovery.c
@@ -46,6 +46,17 @@ typedef struct {
   ListNode *current_characteristic;
 } GATTServiceDiscoveryContext;
 
+// Sets s_discovery_in_progress to false and gives the stop semaphore if a stop has been
+// requested. Called from every callback termination path so that bt_driver_gatt_stop_discovery
+// is always unblocked, even when discovery completes (naturally or via error) after the stop
+// request flag was set but before the next stop-check at the top of a callback runs.
+static void prv_signal_discovery_done(void) {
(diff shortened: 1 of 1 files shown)


jplexer (approved):

gmarull:
@sjp4 pls try

sjp4:
> @sjp4 pls try

I tried - got some more crashes (see https://linear.app/core-dev/issue/MOB-6961/crashlooped-to-prf) - I think it was on that version but not 100% sure

PR #1275, opened 2026-05-12
fw/services/activity: gate sleep tracking on HRM off-wrist

jplexer:
The kraepelin sleep detector already treats charging as a definite not-worn signal, but its accel-only off-wrist heuristic takes 30+ minutes to fire and is easily defeated by ambient vibration. Users report multi-hour false sleep sessions when the watch sits on a desk.

Cache the worn-status from each HRM BPM event in the activity service and OR a recent HRMQuality_OffWrist reading into the not-worn signal passed to kalg_activities_update. The logged AlgMinuteDLSSample.plugged_in field keeps its original "is charging" meaning; only kraepelin's internal definite-not-worn decision sees the combined signal.

The cached reading goes stale after 15 minutes, which comfortably covers the default HRMonitoringInterval_10Min (~11-min) cycle. At the 30/60-min intervals the cache will expire between bursts and sleep detection falls back to the existing accel heuristic, rather than leaning on a possibly-stale on-wrist reading.

Fixes FIRM-1901

### DIFF
--- a/include/pbl/services/activity/activity_private.h
+++ b/include/pbl/services/activity/activity_private.h
@@ -67,6 +67,13 @@ typedef uint16_t ActivityScalarStore;
 #define ACTIVITY_HRM_SUBSCRIPTION_ON_PERIOD_SEC  (1)
 #define ACTIVITY_HRM_SUBSCRIPTION_OFF_PERIOD_SEC (SECONDS_PER_DAY)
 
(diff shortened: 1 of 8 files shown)


gmarull (approved):

jplexer:
(forcepush only changed param from plugged_in to definitely_not_worn to make it more clearer)

PR #1277, opened 2026-05-13
third_party/nimble/transport/hci_sf32lb52: handle ACL pool exhaustion

gmarull:
The H4 RX task allocates ACL mbufs from mpool_acl on every inbound packet and forwards them to ble_hs_rx_q. The host event loop drains the queue, but it can be parked for long stretches (e.g. blocked in gatt_client_subscriptions_handle_server_notification() waiting on a full per-client subscription buffer). While the host is parked, ACL packets keep arriving via IPC and accumulate in ble_hs_rx_q, each pinning one mbuf. With BLE_TRANSPORT_ACL_FROM_LL_COUNT = 60, sustained notification backlogs can drain the pool within ~100ms; the next alloc returns NULL and trips the assert at hci_h4.c:289.

Now that the matching submodule change lets hci_h4_sm_rx() propagate -1 instead of asserting, give the transport a clean back-off path:

  * Add s_acl_pool_avail, a binary semaphore signaled from a put callback registered on the LL-direction slot of the transport pool. The callback fires whenever an LL-tagged ACL mbuf is returned to mpool_acl, i.e. once the host has finished processing one packet and ble_l2cap_remove_rx() frees the chain.

  * In prv_hci_task_main(), when hci_h4_sm_rx() returns < 0 (pool exhausted) block on s_acl_pool_avail with a 100ms safety timeout, then retry with the same pbuf/len. The state machine has preserved partial state, so the retry resumes exactly where the failed allocation left off; this matters because ipc_queue_read() is destructive and dropping bytes would desync the H4 stream.

  * Distinguish OOM (rc < 0) from "consumed nothing" (rc == 0): the latter still logs and breaks out of the inner loop.

ble_hs_rx_q is now implicitly bounded by POOL_ACL_COUNT: every queue entry corresponds to one live mbuf, so the queue can never exceed the pool size. The pool itself acts as the back-pressure channel; once it fills, the transport blocks and BLE LE-credit flow control propagates the pressure to the controller via the IPC FIFO.

Also bump the mynewt-nimble submodule to pull in the matching hci_h4.c change that removes the two assert(rc >= 0) lines.

### DIFF
--- a/src/fw/comm/ble/gatt_client_subscriptions.c
+++ b/src/fw/comm/ble/gatt_client_subscriptions.c
@@ -29,10 +29,6 @@
 #include "FreeRTOS.h"
 #include "semphr.h"
 
-//! Time to wait/block for when the buffer is full and needs to be drained by the client.
-//! Note that bt_lock() is held while waiting, so this has to be rather small.
-#define GATT_CLIENT_SUBSCRIPTIONS_WRITE_TIMEOUT_MS (100)
-
 // TODO:
 // - Intercept "manual" CCCD writes from the app, error for now? or translate to
 //   ble_client_subscribe calls?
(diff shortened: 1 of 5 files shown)


jplexer (approved):
memfault thanks you

gmarull:
Will test on some heavily affected users first and also wait for NimBLE guys' feedback.

ericmigi:
how's this going?

PR #1284, opened 2026-05-13
Fix a few syscall problems

gmarull:


src/fw/syscall/syscall_speaker.c
```diff
@@ -59,6 +59,19 @@ DEFINE_SYSCALL(bool, sys_speaker_play_tone, uint16_t freq_hz,
...
+  // each track's num_notes and sample->num_bytes from user memory (e.g. as the
+  // size argument to kernel_malloc/memcpy that copy the user buffers into
+  // kernel buffers). A racing app could resize them after our validation and
+  // overflow either the kernel allocation or the validated source range.
+  //
+  // Take a kernel-stack copy of the track table (and embedded sample structs)
+  // up front and pass the kernel copy downstream. The notes/data pointers
+  // themselves still aim into user memory, but the sizes are now stable.
+  SpeakerTrack kernel_tracks[SPEAKER_MAX_TRACKS];
+  SpeakerSample kernel_samples[SPEAKER_MAX_TRACKS];
```
jplexer, line 72: is kernel stack large enough? because a speaker track can get large

gmarull: haven't checked tbh, this has all been automated still reviewing myself as well :-)

gmarull: Kernel-side numbers: `SPEAKER_MAX_TRACKS == 4` (include/pbl/services/speaker/speaker_service.h:76), `sizeof(SpeakerTrack) == 12` (notes ptr + num_notes + sample ptr), `sizeof(SpeakerSample) == 16`. So the two on-stack arrays add up to 4·12 + 4·16 ≈ 112 bytes — well within the kernel stack budget. The user-facing sample/notes payload itself never lives on the stack; only the small descriptor table does.

jplexer (approved):

PR #1290, opened 2026-05-14
toggle: add shutdown as quick launch shortcut

### LINKED ISSUE #1211: add 'Shutdown' to Quick Launch

### Proposed feature request

should add 'Shutdown' to Quick Launch, I'd love to press and hold the back button for easy way to shutdown the watch completely.

I shut it down each night (I don't wear mine to bed, and I like preserving battery with it off).  I press and hold the back button to turn it on the morning, would love to do the same at night when I go to bed.

I understand there's already 'Airplane Mode' in Quick Launch already.  I also know there's 'Standby Mode' to go into/out-of Airplane Mode automatically.  I want full power off every night, right now every night I navigate the menu system to do it, little clunky:  Settings; scroll all the way down to System; scroll all the way down to Shutdown (but not one too far, you don't want Factory Reset!)

I think if you implement this still having the 'Do you want to shut down?' dialog makes sense, good balance between ease of use and intentionality 

### Alternatives

_No response_

### Anything else?

_No response_

s00se:
Closes #1211

### DIFF
--- a/src/fw/apps/system/toggle/shutdown.c
+++ b/src/fw/apps/system/toggle/shutdown.c
@@ -1,0 +1,63 @@
+/* SPDX-FileCopyrightText: 2025 The PebbleOS Contributors */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "shutdown.h"
+
+#include "applib/app.h"
+#include "applib/ui/dialogs/actionable_dialog.h"
+#include "applib/ui/dialogs/dialog.h"
+#include "kernel/event_loop.h"
+#include "process_management/app_manager.h"
+#include "pbl/services/i18n/i18n.h"
+#include "resource/resource_ids.auto.h"
+#include "shell/normal/battery_ui.h"
+
+static void prv_do_shutdown(void *data) {
+  battery_ui_handle_shut_down();
+}
+
+static void prv_shutdown_confirm_cb(ClickRecognizerRef recognizer, void *context) {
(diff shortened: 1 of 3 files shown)


jplexer (commented):
Can you clean up commits? otherwise lgtm

asyba:
Would this new feature also appear when you go to settings-system shut down?
because if not, could be nice to have it also there?

s00se:
Cleaned up as requested

gmarull (changes requested):
shutdown is currently not ideal, we do a full PMIC shutdown (aka shipping mode). I'm fine merging this once we use MCU shutdown, so we preserve e.g. time.

calexander3:
So what's the story with this pr? I'd love to have this functionality. What can I do to help?

caco3:
I did a re-implementation, but it needs also changes in a submodule, see https://github.com/coredevices/PebbleOS/pull/1824

calexander3:
Cool! Just curious, what's the difference between hibernate and shutdown? Is it why my watch always starts at 7:00 when I turn it on?

caco3:
@calexander3 The shutdown mode switches the power and clocks completely off. This is also called "Shipping Mode". The RTC (Real Time Clock) then also looses power and the watch will use 1 January 2010 when it powers on (default value). In the hibernate mode, the RTC will continue to get powered and clocked, thus the time information will not get lost.

PR #1299, opened 2026-05-15
fw/drivers/display/sf32lb: silent-loss watchdog with CDR telemetry [FIRM-1928]

jplexer:
Based on the work in #1189 (closed pending vendor analysis), rebased
  onto current main and extended with a telemetry layer so we keep
  collecting diagnostic data in the field once this lands.

  ## What's here

  - **Watchdog**: 200ms NewTimer-based timeout that catches the case
    where the LCDC kicks off but never delivers an EOF callback,
    leaving `s_updating` stuck true and the compositor silently
    quiesced. Force-completes on KernelMain so the next frame renders.
  - **Drop-complete test hook**: `display drop_complete` prompt
    command, gated `!RELEASE && (OBELIX || GETAFIX)`, arms a one-shot
    drop of the next IRQ callback. Exercises the watchdog path
    without needing a real reproduction.
  - **CDR telemetry** (new in this PR): when the watchdog fires,
    snapshot the LCDC handle (State, ErrorCode, debug counters),
    IRQ raw status, JDI FSM-running bit, and the first 2KB of the
    LCDC register file into a Memfault Custom Data Recording.
    Throttled to 1 upload per device per 24h to match Memfault's
    server-side limit, with freshest-of-window semantics so a burst
    of freezes uploads the most recent snapshot, not the first.
    Plus a `lcdc_silent_loss_count` heartbeat metric for fleet-wide
    prevalence.
  - **Decoder**: `tools/decode_lcdc_snapshot.py` parses the binary
    CDR and prints decoded HAL state, error bits, kickoff/completion
    delta (the smoking gun for the silent-loss path), and named
    register values from `lcd_if.h`.
    
## Why land this without a root cause from Sifli yet

  #1189 was held while waiting for a register dump from a real
  reproduction; that hasn't materialized because the freeze is rare
  on dev hardware. Meanwhile field reports keep accumulating —
  FIRM-1740, FIRM-1747, FIRM-1793, FIRM-1805, FIRM-1864, FIRM-1886,
  FIRM-1899, FIRM-1909, FIRM-1924, all on Obelix v4.9.163 through
  v4.9.175, all matching the same signature (screen frozen, buttons
  still react, no coredump, screenshots stuck in the un-converted
  332 framebuffer format).

  The telemetry layer turns deployed watches into our reproduction
  fleet: each watchdog fire auto-uploads exactly the register state
  Sifli originally asked for, without the user having to do anything.
  We get the fix in users' hands AND keep collecting root-cause data
  in parallel.

  ## Test plan

  - [x] Builds clean for `obelix_pvt --memfault=1` and `getafix_dvt2`
        (MEMFAULT=0 stubs the diagnostics module to no-ops).
  - [x] CDR upload verified end-to-end: synthetic capture via a
        temporary `display force_cdr` prompt command (not included
        in this PR), shipped to Memfault, downloaded, decoded with
        `tools/decode_lcdc_snapshot.py`, sentinel `0xDEADBEEF`
        `watchdog_fire_count` identifiable.
  - [ ] Watch `lcdc_silent_loss_count` metric in Memfault over the
        next 1–2 weeks of field rollout to confirm fleet prevalence
        and start collecting real CDRs.


gmarull (commented):
could we just dump LCDC into coredump, and when it happens croak?

jplexer:
i guess we could, yeah
so remove watchdog and *just* do that?

gmarull:
> i guess we could, yeah so remove watchdog and _just_ do that?

yes, so we can capture problem and send the dump to sifli, and users will just see a crash (better than freeze IMHO...)

PR #1302, opened 2026-05-15
fw/drivers/display/sf32lb: crash on LCDC silent loss [FIRM-1928]

jplexer:
Supersedes #1299 per @gmarull's review feedback: instead of recovering
  silently, detect the failure and crash so we get a Memfault coredump for
  Sifli + the user sees a reboot (better than a permanent freeze).

  ## What's here

  - **Detect-and-crash path** in `display_jdi.c`. A 500ms NewTimer arms
    inside `display_update()` alongside the LCDC kickoff. If the EOF
    callback doesn't fire by the timeout, snapshot the first 2KB of the
    LCDC peripheral register file into a static volatile BSS buffer
    (`s_lcdc_pre_crash_regs`) and `PBL_CROAK`. The user sees a clean
    reboot; Memfault gets a coredump with reboot reason `Assert` and the
    croak message embedded in the log.
  - **Race protection**. A volatile `s_eof_observed` flag is set from the
    IRQ before the terminate event is posted. If EOF fires between us
    arming the timer and the timeout (e.g. scheduler latency on the
    KernelMain queue), the silent-loss handler sees the flag and bails
    out before crashing — keeps false-positives off the dashboard.
  - **Coredump extra-regions registry**
  (`src/fw/kernel/coredump_extra_regions.{h,c}`).
    Generic kernel-side API for drivers to register BSS buffers that
    should ride in Memfault coredumps in addition to the default thread
    stacks / TCBs / log buffers. The first user is the LCDC snapshot,
    but the registry is general — any driver that snapshots fault-time
    state can call `coredump_extra_regions_register()`. Wired through
    `main.c` immediately before `pbl_analytics_init()` so the registry
    is populated before Memfault reconstruction reads it.
  - **`display drop_complete` prompt hook** (debug-only, gated on
    `!RELEASE && (PLATFORM_OBELIX || PLATFORM_GETAFIX)`). Arms a one-shot
    drop of the next LCDC transfer-complete callback, simulating the
    silent-loss failure mode. Used to validate the detect-and-crash path
    end-to-end on hardware without needing a real reproduction.

### DIFF
--- a/src/fw/console/prompt_commands.c
+++ b/src/fw/console/prompt_commands.c
@@ -1105,6 +1105,19 @@ void command_audit_delay_us(void) {
   __enable_irq();
 }
 
+#if !defined(RELEASE) && defined(CONFIG_DISPLAY_JDI_SF32LB)
+#include "drivers/display/sf32lb/display_jdi.h"
+
+// Arms the JDI display driver to drop the next LCDC transfer-complete
+// callback, simulating the silent-loss failure mode (e.g. SiFli HAL ICB
+// overflow). The silent-loss timer should fire ~500ms later and PBL_CROAK,
+// producing a Memfault coredump and a watch reboot.
+void command_display_drop_complete(void) {
+  display_jdi_test_drop_next_complete();
+  prompt_send_response("display: armed drop of next LCDC complete; PBL_CROAK in ~500ms");
+}
+#endif
+
 #if !MICRO_FAMILY_SF32LB52
 // Simply parks the chip permanently in stop mode in whatever state it's currently in. This can be
 // pretty handy when trying to profile power of the chip under certains states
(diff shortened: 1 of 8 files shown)


src/fw/console/prompt_commands.c
```diff
@@ -1105,6 +1105,19 @@ void command_audit_delay_us(void) {
   __enable_irq();
 }
 
+#if !defined(RELEASE) && (PLATFORM_OBELIX || PLATFORM_GETAFIX)
```
gmarull:
```suggestion
#if !defined(RELEASE) && defined(CONFIG_DISPLAY_JDI_SF32LB)
```

gmarull (approved):
nitpick

PR #1303, opened 2026-05-16
Scale SimpleDialog font on >=200px displays

aveao:
Makes the SimpleDialog (quiet time, alarm removed, charge notifications etc) scale properly for the screen size for obelix and gabbro.

~~lmk if you prefer a different approach to scaling the icon.~~

~~this excludes scaling animated icons (Alarm Clock, Result Shredded, Dismissed, Sent, Generic Confirmation, etc.) for simplicity, and as those already appear to look alright.~~ (dropped icon scaling after discussion about cramped views on pebble discord, lmk if you disagree and I'll revert)

---

before & after:


aveao:

gmarull (approved):

ericmigi:
I missed this one before it got merged - I think moving to a bigger font is OK, but we shouldn't bold all these strings. Do we have the same bigger font but not bolded? cc @jplexer

PR #1306, opened 2026-05-17
Health: Add current bin in typical step text

aveao:
Turns out the "typical <day>" text represents the typical amount of steps on that day up to the 15 minute minute block you're on :D

Caught this while taking a shot at https://github.com/coredevices/mobileapp/pull/192

This changes that text to "Typical <day> <timestamp>" so that it's clear what it's representing (and ideally should not require translation changes), plus it widens the pill for typical and moves it down slightly on larger screens (emery and gabbro) so that the whole view appears more spacious. Rest of health view needs re-scaling too, but this is hopefully a good starting point.

---

Before & After:


aveao:

gmarull:
shouldn't 'typical thursday' mean, typical step count for 'thursday' (whole day)? if it's not like that, we should fix?

gmarull:
> shouldn't 'typical thursday' mean, typical step count for 'thursday' (whole day)? if it's not like that, we should fix? I guess the objective of the metric is to tell how far we are from the typical total, I find time confusing here.

aveao:
Currently, the number represents how much you walked on avg until this 15 minute block. Agreed that it's confusing, which is why I did this PR :p I was trying to figure out why it differed from the debug data on the phone.

I find the idea of "here's how much you've walked until this time on thursdays" to be kind of useful, because unless you lead a particular life you're not walking consistent amounts but a bit on morning, midday and evening, and so comparing to the full day's 8k when you're doing 5k of that in the evening usually would reveal a different data point. But I think it's a different datapoint, and I think displaying both would be best.

An alternative I was thinking of was showing both on the lower text, then it fits into all screens better. maybe something like

> **TYPICAL THU**
> 8000 (🕖 2000)

gmarull:
> An alternative I was thinking of was showing both on the lower text, then it fits into all screens better. maybe something like
> 
> > **TYPICAL THU**
> > 8000 (🕖 2000)

that seems better

aveao:
ehhhh it's a tough one

first tried some cool clock svgs, it didn't convey super clearly that it's the number until now

then tried this really cool homage to tictoc (actually dynamically generated on open)

screenshot_emery_red  screenshot_flint_red

but it's both not very clear to the user and kinda bad screen real estate

then I tried this but it's way too busy:

screenshot_emery_split_v2-1


---

current state of code is showing the all-day-typical on small screen watches, and this on emery and gabbro:

screenshot_emery_shifted

It's a bit busy but I'm just not sure what's the best balance of eye candy and functionality is here.

fwiw the yellow bar in the bar around the shoe is based on the "typical until this time on this weekday" number.

ericmigi:
I'm not sure it entirely makes it more understandable than the original UI? I think the bar that fills up through the day makes sense and explains that it's a progress-based number.

aveao:
I think that's pretty cryptic, it took me until two hours ago to figure that out and you know how long I've been using one of these :p

IMO even if not in this shape, some form of clarification would be useful, as sleep has the exact same text but shows the total of the day, and the step count is showing typical until the time of day. I don't think very many realized this feature works this way.

ericmigi:
image
 seems like the strongest candidate, but agreed it is still busy

tilden:
Randomly jumping in here -- I like the idea of clarifying that "TYPICAL TUE" actually means "On a typical Tuesday you have walked N steps by the current time", since it _is_ kind of ambiguous, as @aveao noted.

I was thinking @aveao's second layout could potentially work with some refinement; maybe something like this:

image

gmarull:
> Randomly jumping in here -- I like the idea of clarifying that "TYPICAL TUE" actually means "On a typical Tuesday you have walked N steps by the current time", since it _is_ kind of ambiguous, as @aveao noted.
> 
> 
> 
> I was thinking @aveao's second layout could potentially work with some refinement; maybe something like this:
> 
> 
> 
> image
> 
> 

I like this last mockup!

jplexer (commented):
looks good but could you maybe squash (at least 1 + 3) since the first design is essentially dead?

gmarull:
image

time/total should be gray?

tilden:
> time/total should be gray?

I was actually using `GColorArmyGreen` in the mockup, though there were some concerns about contrast on the real watch display. I took a couple of real watch photos for comparison:

`GColorBlack` subheading text + divider line:
steps-gcolorblack

`GColorArmyGreen` subheading text + divider line:
steps-gcolorarmygreen

I don't feel *exceedingly* strongly about it either way. Thoughts?

gmarull:
> > time/total should be gray?
> 
> I was actually using `GColorArmyGreen` in the mockup, though there were some concerns about contrast on the real watch display. I took a couple of real watch photos for comparison:
> 
> `GColorBlack` subheading text + divider line: steps-gcolorblack
> 
> `GColorArmyGreen` subheading text + divider line: steps-gcolorarmygreen
> 
> I don't feel _exceedingly_ strongly about it either way. Thoughts?

I personally vote for the grey to differentiate

aveao:
Apologies for shaky photos,

I tested both dark gray and army green yesterday, alongside a few other colors, and they all result in really poor contrast on the PT2. I would not suggest to use them in a low light situation.

PXL_20260520_190111522 PXL_20260520_190119667

I did some capturing of the screen in real life light conditions with color calibration so I can emulate more realistic colors on screenshots, and here's some simulated cases, specifically outdoors (hamburg sunlight, so not very sunlight), outdoors (PTS) vs indoors (low cri, mostly direct), neither with backlight:

Untitled-on-device
Untitled-on-device(1) Untitled-on-device(3)

(I know especially the middle one looks ridiculous at first glimpse, but it's pretty similar to the right image of the actual watch, even though it was with a different camera and different light source)

fwiw, same profiles on armygreen:

Untitled-armygreen-on-device Untitled-armygreen-on-device(1) Untitled-armygreen-on-device(2)

tilden:
Regarding readability, another thing to consider is that on the Time 2 and Round 2, the fonts should all probably be bumped up by 1 size to match other font size increases throughout the OS on those platforms (Presumably this would be in a future PR)

gmarull:
what about a "dark yellow"?

tilden:
> what about a "dark yellow"?

The color name is a bit confusing, but "GColorArmyGreen" is actually the darkest yellow on the platform IMO:

image

aveao:
Yeah, there's no dark yellow, and closest is army green, which I tried both on device and simulated colors and it looked almost identical to dark gray.

jplexer:
first of all, sorry ave you gotta regenerate those fixtures 😆 
secondly I personally think that while army green is less readable than black text in darker environments, it is worth the differenciation
so i'd go with army green

tilden:
I'm also okay with the black text if it would make it easier on @aveao!

aveao:
I'll do army green, but a bit later than today ;)

aveao:
Okay, rebased, moved to armygreen, and regenerated images.

jplexer (approved):
Looks good to me! CC: @gmarull

PR #1307, opened 2026-05-17
health/detail_card: drop HEALTH_HEIGHT_OFFSET which causes incorrect heights

aveao:
Introduced here: https://github.com/coredevices/PebbleOS/commit/ef740cb5d9f5da66e5859c3f2a4f6d8da62f9eea

Dropping it leads to text being centered properly, and better utilizes screen real estate.

---

before & after:


aveao:

gmarull (approved):

PR #1311, opened 2026-05-18
fw/services/light: linearly ramp dynamic backlight intensity

aveao:
This one is meant more as an example impl and a space for discussions.

Right now backlight intensity is one of 3:
- 10%, if below user-set dynamic backlight min threshold (default 5, imo should be 2)
- User-set "Max Intensity", if above the user-set dynamic backlight min threshold
- Nothing, if above the hardcoded dark threshold (150 for obelix)

So it's not very much a dynamic backlight. I believe having the value be dynamic based on measured light levels would be more appropriate. If in a dimly lit room (~8 counts on my watch for bedside lamps, 2-3 for ceiling light if very indirect), it can very much benefit from being above 10%, but below user-set max. In a rather bright space (e.g. same room on a brighter day with sunlight, but no direct light to remove need for backlight), it can benefit reaching user-set max (in my case, "blinding").

This is one such demo. This already improves my experience, but I am not maximally happy with it. In a dim room it makes the screen readable, but not properly match ambient levels, while not being blinding, so it's still a partial win.

Several implementation details to figure out would be:
- linear vs not
- whether user's max intensity is used for a range, or as a clamp. this is for a range, but I think clamp may make more sense.

Open to ideas! :D

### DIFF
--- a/src/fw/services/light/service.c
+++ b/src/fw/services/light/service.c
@@ -148,15 +148,33 @@ static uint8_t prv_backlight_get_intensity(void) {
   }
   
 #if CAPABILITY_HAS_DYNAMIC_BACKLIGHT && !defined(RECOVERY_FW)
-  // Dynamic backlight: dim in utter darkness, otherwise user max. The
-  // bright-outdoor case is already filtered upstream by prv_light_allowed()
-  // for ambient-sensor-enabled wakes; the few paths that bypass that gate
-  // (app-driven force-on, ambient-sensor pref off) sensibly land at max here.
+  // Dynamic backlight: linear ramp from dim_intensity at dynamic_min_threshold
+  // up to 100% at ambient_light_dark_threshold, then clamped to user_max. This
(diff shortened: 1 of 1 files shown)


ericmigi:
cool idea! @jplexer can you test it out?

aveao:

jplexer:
we had something similar before but my impl was not great :D
i'll try this out, thanks!

aveao:
I'd strongly suggest to try it with blinding, as I honestly think the clamping vs linearly scaling approach is better (but haven't added that yet, also my lunch break is about to be over so can't add that until later).

aveao:
(and I think it works better with the ranges on #1313 due to less noise in the data, making the backlight more consistent)

aveao:
Moved this to be clamping instead. Might make sense to make it non-linear with some further, but in current state it's still an improvement over before. Feel free to merge if you like it.

jplexer (approved):

PR #1313, opened 2026-05-18
Utilize continuous sampling on W1160 + increase integration time to 99ms

aveao:
(building on #1311, has part of that code here too)

This is very much intended as a demo, and a discussion starter. ~~The code as sent here won't work right as dark-room-threshold (150) is still set for the lower integration time. Trivial fix but not worth it without talking it out first, plus I'd want to measure a bright room which is not so easy at 3:39AM.~~ (4:15 update: I found good thresholds)

This barely changes the overall power usage (prob like 10x from 7uA to 70uA, amounting to almost nothing over a day) per datasheet, makes the sensor read non-blocking, while allowing increasing integration time (how long to capture photons for) by quite a bit. What was 7 samples is ~50 samples, and overall values are a lot less noisy, staying within ~10%. IMO, a solid improvement to data quality of the ambient light sensor.

### DIFF
--- a/src/fw/board/boards/board_obelix.c
+++ b/src/fw/board/boards/board_obelix.c
@@ -569,8 +569,8 @@ const BoardConfigPower BOARD_CONFIG_POWER = {
 
 const BoardConfig BOARD_CONFIG = {
   .backlight_on_percent = 45,
-  .ambient_light_dark_threshold = 150,
-  .ambient_k_delta_threshold = 25,
+  .ambient_light_dark_threshold = 600,
+  .ambient_k_delta_threshold = 100,
   .dynamic_backlight_min_threshold = 5,
   .backlight_default_color = BACKLIGHT_COLOR_WARM_WHITE,
 };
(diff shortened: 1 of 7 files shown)


gmarull (commented):
thanks for the PR! please make sure to use real name in signed off/author as per guidelines

aveao:
ave is my real name but I'll swap these over to full name

aveao:
Done

jplexer:
> ave is my real name but I'll swap these over to full name

yeah he meant real name as full name, thank you!

aveao:
(ok google how do I get ave as my künstlername)

gmarull (commented):
is delta ~70uA (in average), or just the increase when acquiring new samples?

aveao:
I have no actual data, please do measure yourself if you can. I'm just basing it roughly between the two provided numbers on: https://www.lcsc.com/datasheet/C49230206.pdf

gmarull:
> I have no actual data, please do measure yourself if you can. I'm just basing it roughly between the two provided numbers on: https://www.lcsc.com/datasheet/C49230206.pdf

I'll do when I have a chance, but if delta is a continuous 70uA it's definitely a no-go.

aveao:
I did my power calculations wrong, assumed one shot not continuous, but the datasheet implies it is continuous :( Yeah at continuous 70uA that'd be definitely more notable.

PR #1316, opened 2026-05-18
applib: reset touch state via syscall

gmarull:
## Summary
- route touch_service_subscribe() reset through a privileged syscall
- avoid app MPU faults when subscribing to touch events during launch

Fixes FIRM-1957
Fixes FIRM-1958

## Testing
- ./waf build
- gitlint

### DIFF
--- a/src/fw/applib/touch_service.c
+++ b/src/fw/applib/touch_service.c
@@ -13,8 +13,6 @@
 #include "syscall/syscall.h"
 #include "system/passert.h"
 
-bool sys_touch_service_is_enabled(void);
-
 //! @return the per-task touch service state, or NULL if the current task is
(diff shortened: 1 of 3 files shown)


jplexer:
was going to ask claude to check through all exposed apis to see if they are correctly done via syscalls
did you do something like that and this is the only one or do I still need to check?

gmarull:
> was going to ask claude to check through all exposed apis to see if they are correctly done via syscalls did you do something like that and this is the only one or do I still need to check?

do that, I just found this one because your breakerbrick game crashes :-)

jplexer:
not breakerbrick 😭

jplexer (approved):

jplexer:
claude found nothin

PR #1318, opened 2026-05-18
third_party/moddable: unprivileged FFI dispatch

jplexer:
Tested with helloffi (everything works there) and tested with my POC (unprivileged, and trying to access kernel mem crashes the app (as expected)


gmarull (commented):
have we checked how feasible it is to run moddable event loop fully unprivileged?

jplexer:
needs a lot of changes!

gmarull:
> needs a lot of changes!

I think that's the path honestly, because running JS in privileged mode, even if "sandboxed" is screaming for a _fatal_ bug in the future

jplexer:
maybe at some point later but right now I think its too much of a hassle

gmarull (approved):

gmarull:
let's merge this for now, but discuss on isolating moddable completely if possible

PR #1321, opened 2026-05-18
drivers/ambient_light: raise level max to 16 bits on W1160

### LINKED ISSUE #1310: AMBIENT_LIGHT_LEVEL_MAX incorrectly set to 4096 for W1160

### Current Behavior

Ambient light sensor has some asserts for <= AMBIENT_LIGHT_LEVEL_MAX which is incorrectly set to 4096 despite W1160 being on 16 bit mode and capable of reaching 65535.

Unsure of practical implications yet, but I am trying to figure out why ambient sensor doesn't perform the way I'd hope it to and this was one thing I noticed.

### Expected Behavior

OPT3001 maxes out at 4096, W1160 at 65536.

### Steps To Reproduce

https://github.com/coredevices/PebbleOS/blob/main/src/fw/drivers/ambient_light.h#L20

### Version

v4.9.177

### Host OS

N/A

### Watch

Pebble Time 2 (Obelix)

### Anything else?

_No response_

jplexer:
The W1160 ALS is configured in 16-bit data format and ambient_light_get_light_level() can return up to 65535. The shared AMBIENT_LIGHT_LEVEL_MAX was 4096, which tripped
ambient_light_set_dark_threshold()'s assert and clamped backlight threshold prefs/settings UI to the OPT3001 12-bit range.

Select the constant by Kconfig so W1160 boards use the full sensor range while OPT3001 stays at 4096.

Fixes #1310

### DIFF
--- a/src/fw/drivers/ambient_light.h
+++ b/src/fw/drivers/ambient_light.h
@@ -17,7 +17,12 @@ typedef enum AmbientLightLevel {
 
 #define AMBIENT_LIGHT_LEVEL_ENUM_COUNT (AmbientLightLevelVeryLight + 1)
 
-static const uint32_t AMBIENT_LIGHT_LEVEL_MAX  = 4096;   // max 12 bits
+#ifndef CONFIG_AMBIENT_LIGHT_BITS
+// Fallback for header parsers (e.g. SDK generator) that don't preload autoconf.h.
+#define CONFIG_AMBIENT_LIGHT_BITS 12
+#endif
+
+static const uint32_t AMBIENT_LIGHT_LEVEL_MAX = (1U << CONFIG_AMBIENT_LIGHT_BITS);
 
 /** Initialize the ambient light sensor */
 void ambient_light_init(void);
--- a/src/fw/drivers/ambient/Kconfig
+++ b/src/fw/drivers/ambient/Kconfig
@@ -18,4 +18,13 @@ config AMBIENT_LIGHT_W1160
     help
       Support for the W1160 ambient light sensor.
 
+config AMBIENT_LIGHT_BITS
+    int "Ambient light sensor resolution (bits)"
+    default 12 if AMBIENT_LIGHT_OPT3001
+    default 16 if AMBIENT_LIGHT_W1160
+    help
+      Number of bits in the ambient light sensor reading. Used to derive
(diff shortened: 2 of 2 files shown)


src/fw/drivers/ambient_light.h
```diff
@@ -17,7 +17,11 @@ typedef enum AmbientLightLevel {
 
 #define AMBIENT_LIGHT_LEVEL_ENUM_COUNT (AmbientLightLevelVeryLight + 1)
 
+#ifdef CONFIG_AMBIENT_LIGHT_W1160
+static const uint32_t AMBIENT_LIGHT_LEVEL_MAX  = 65535;  // max 16 bits (W1160)
+#else
 static const uint32_t AMBIENT_LIGHT_LEVEL_MAX  = 4096;   // max 12 bits
+#endif
```
gmarull: this is a hack, can we decouple API from that? If not yet, maybe make hack a bit more "elegant" e.g.

```
config AMBIENT_LIGHT_BITS
    int "Ambient light resolution"
    default 12 if AMBIENT_LIGHT_...
    default 16 if ...
```

gmarull: maybe a percentage would be enough, or convert to lux...

jplexer: mhh sure

aveao: I don't think you can convert to lux with the data that's available, and percentage is unlikely to work well (the current cutoff for non-dark rooms is 150, which is 0.0022% of 65535).

src/fw/drivers/ambient_light.h
```diff
@@ -17,7 +17,12 @@ typedef enum AmbientLightLevel {
 
 #define AMBIENT_LIGHT_LEVEL_ENUM_COUNT (AmbientLightLevelVeryLight + 1)
 
-static const uint32_t AMBIENT_LIGHT_LEVEL_MAX  = 4096;   // max 12 bits
+#ifndef CONFIG_AMBIENT_LIGHT_BITS
+// Fallback for header parsers (e.g. SDK generator) that don't preload autoconf.h.
+#define CONFIG_AMBIENT_LIGHT_BITS 12
+#endif
```
gmarull, line 23: still hacky, but well, this all needs a proper refactor anyway

gmarull (approved):

PR #1326, opened 2026-05-19
fw/applib: Provide sdk music volume and sound for iOS find phone app and others

asyba:
I got tired of forgetting my iPhone in the garage and not being able to find it, so I did some tests to see how to run "Find My Phone" on iOS.

I know there's an app that allows you to do this on Android, but it's missing for iOS since it's more restrictive.
https://apps.repebble.com/find-my-phone_695da4c3fbfdde0009f99799

**Research:**
According to my research, on iOS we can use AudioServicesPlayAlertSound, AudioServicesPlaySystemSound, AudioServicesPlaySystemSoundWithCompletion, AVSpeechSynthesizer, and AVAudioPlayer.

But playing around with different options, not all of them work well for every test scenario. For example, if the volume is low/mute, it will only vibrate; if it's on silent (physical button), it will only vibrate. So, after experimenting, I concluded that the ones that will always play sound, regardless of whether it's on silent or in focus mode, are AVAudioPlayer and AVSpeechSynthesizer.

**Implementations**

- With AVAudioPlayer, we have to use a custom MP3 ringtone. I uploaded a short 3-second example.

- AVSpeechSynthesizer is useful, for example, for speaking a phrase or word aloud. I believe it's primarily used for accessibility, but it could also be useful for finding the phone or for other functionalities the community might come up with.

- I set the sound to a 3-part loop. If we make it infinite, it will keep playing continuously, and if we lose connection/context with the watch, we can't stop it. The only option is to manually close the mobile app. We need to be careful about leaving it playing indefinitely.

- I created a test watch app; here's the code: https://github.com/asyba/PlaySoundMobile obviously, it uses the modifications I made to the SDK in this PR.

- The mobile PR https://github.com/coredevices/mobileapp/pull/199 primarily allows calling SMS/alarm sounds, a default ringtone, synthesized voice, and sounds using the system sound ID.

The FW allows to call those sound methods plus expose the volume and get level control, to also allow the watchapps to set the level to max for find my phone feature.

### DIFF
--- a/include/pbl/services/music.h
+++ b/include/pbl/services/music.h
@@ -62,7 +62,7 @@ void music_get_pos(uint32_t *track_pos_ms, uint32_t *track_length_ms);
 int32_t music_get_playback_rate_percent(void);
 
 //! @return The volume percentage.
-uint8_t music_get_volume_percent(void);
+uint8_t music_service_get_volume_percent(void);
 
 //! Retrieve the current playback state.
 MusicPlayState music_get_playback_state(void);
@@ -76,7 +76,7 @@ bool music_is_playback_state_reporting_supported(void);
 bool music_is_progress_reporting_supported(void);
 
 //! @return True if the service supports reporting of the current volume.
-//! @see music_get_volume_percent
+//! @see music_service_get_volume_percent
 bool music_is_volume_reporting_supported(void);
 
 //! Sends the command to the server. Commands are "unreliable", they are sent at "best effort".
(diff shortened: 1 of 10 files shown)


jplexer:
I think this should wait until we have a better permission system: otherwise an app could play loud noises without my consent

asyba:
@jplexer like a system where people can allow/denied if the app/face can control volume, take screenshots, play sounds, turn off wifi, etc on the mobile connected?

How long will it take to start that new system? In the next 6 months? sooner?

jplexer:
I cannot give you an estimate really, I think @sjp4 was primarily working on this

sjp4:
Hopefully in the next month or two

PR #1330, opened 2026-05-19
clean up some things

jplexer:
another victim of a nerdsnipe


waftools/kconfig.py
```diff
@@ -109,6 +109,16 @@ def configure(conf):
...
 
+    # Bridge --release into Kconfig so symbols like CONFIG_MEMFAULT can
+    # use `default y if RELEASE`. Defined as a hidden bool in the
+    # top-level Kconfig.
+    release_synthesized = False
+    if getattr(conf.options, "release", False):
+        sym = kconf.syms.get("RELEASE")
+        if sym is not None and sym.nodes:
+            sym.set_value("y")
+            release_synthesized = True
```
gmarull: we should avoid this. To build for release, we can pass `-DCONFIG_RELEASE=y` at configure step, which should be the way forward

Kconfig
```diff
@@ -1,4 +1,20 @@
...
+      `./waf configure`. Gates features like Memfault crash reporting
+      that should only run on shipping firmware.
+
+config QEMU
+    bool "QEMU emulated board"
+    help
+      Set in QEMU board defconfigs (qemu_emery, qemu_flint,
+      qemu_gabbro). Mirrors the runtime is_qemu() / bld.env.QEMU
+      checks at Kconfig time so symbols can gate features that don't
+      make sense under emulation.
```
gmarull: ideally this should be a promptless symbol, someone else should select it, we do not have soc/ structure yet, though, so ok as a temporary solution

src/fw/applib/platform.h
```diff
@@ -17,22 +17,14 @@ typedef enum PlatformType {
 // Therefore, we need to switch on the platform for the platform type.
 #if !defined(SDK)
   // Variant boards should go before their base platforms, as they define the base platform too.
-  #if PLATFORM_QEMU_EMERY
+  #if CONFIG_PLATFORM_EMERY
```
gmarull: `#ifdef` (always for Kconfig, unset booleans are undefined)

Kconfig
```diff
@@ -41,5 +41,47 @@ config SCREEN_COLOR_DEPTH_BITS
...
+
+endchoice
+
+config MPU_TYPE_ARMV8M
+    bool
+    default y if CORTEX_M33
+
+config MPU_TYPE_ARMV7M
+    bool
+    default y if !MPU_TYPE_ARMV8M
```
gmarull: this should go into soc/, see Zephyr for inspiration. the arm settings should be under arch/ ideally.

gmarull: we should also use the CONFIG_SOC_*** naming

third_party/nonfree/wscript_build
```diff
@@ -1,7 +1,7 @@
-if bld.is_asterix() or bld.is_obelix() or bld.is_getafix():
-    if bld.is_asterix():
+if bld.env.CONFIG_PMIC_NPM1300:
```
gmarull: we should have a special Kconfig for the fuel gauge lib (we could use the PMIC without), then, the battery service have a nrf fuel gauge impl that depends on it

gmarull (commented):
nice cleanup, I've added a few comments for issues I think need addressing

waftools/kconfig.py
jplexer: so -DCONFIG_RELEASE=y should replace --release from waf?

gmarull: yeah, this is already supported, you can now override any Kconfig via `-DCONFIG_A=y -DCONFIG_B=30`

Kconfig
gmarull: well, you do later in fact!

gmarull (approved):

PR #1339, opened 2026-05-20
fw/light: fall back to on when ALS sensor is unhealthy [FIRM-1986]

jplexer:


gmarull (commented):
I'm a bit reluctant to merge these changes, specially now that we have some proposals in the ALS area. Maybe we should explore if partial continuous mode is more reliable first?

jplexer:
> Maybe we should explore if partial continuous mode is more reliable first?

that didnt get merged though, did it?

gmarull:
> > Maybe we should explore if partial continuous mode is more reliable first?
> 
> that didnt get merged though, did it?

nope, needs testing/power analysis, see if a "partial continuous" sampling would be better, etc.

jplexer:
got it got it

PR #1343, opened 2026-05-20
drivers/ambient: event-gated semi-continuous ALS sampling

jplexer:
Gates W1160 continuous sampling to short interaction windows (wake events open a 5s holdoff, settings menus prime explicitly) so the SNR win from bumping IT to 99 ms doesn't come with the always-on power cost. Average sensor current scales with wake rate × 5 s × ~70 µA which is roughly 0.1 µAh per wake, vs. the ~70 µA continuous draw that killed the previous continuous-mode attempt (#1313). Also bumps obelix dark thresholds to match the new IT and adds a software cache so primed reads stay snappy across suspend/resume (e.g. backlight-on bleed gating).

### DIFF
--- a/src/fw/board/boards/board_obelix.c
+++ b/src/fw/board/boards/board_obelix.c
@@ -576,8 +576,8 @@ const BoardConfigPower BOARD_CONFIG_POWER = {
 
 const BoardConfig BOARD_CONFIG = {
   .backlight_on_percent = 45,
-  .ambient_light_dark_threshold = 150,
-  .ambient_k_delta_threshold = 25,
+  .ambient_light_dark_threshold = 1400,
+  .ambient_k_delta_threshold = 200,
   .dynamic_backlight_min_threshold = 5,
   .backlight_default_color = BACKLIGHT_COLOR_WARM_WHITE,
 };
--- a/src/fw/drivers/ambient/ambient_light_w1160.c
+++ b/src/fw/drivers/ambient/ambient_light_w1160.c
@@ -6,8 +6,10 @@
 #include "drivers/ambient_light.h"
 #include "drivers/i2c.h"
 #include "drivers/periph_config.h"
+#include "drivers/rtc.h"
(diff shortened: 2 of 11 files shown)


src/fw/drivers/ambient/ambient_light_w1160.c
```diff
@@ -125,58 +185,147 @@ void ambient_light_init(void) {
...
+void ambient_light_release(void) {
+  if (!s_initialized) {
+    return;
+  }
+  mutex_lock(s_state_mutex);
+  PBL_ASSERTN(s_prime_refcount > 0U);
+  s_prime_refcount--;
+  prv_apply_sampling_state_locked();
+  mutex_unlock(s_state_mutex);
+}
```
gmarull: refcounting should be implemented in a layer above, or this will propagate to all drivers (see eg zephyr regulator API as a layering scheme)

src/fw/board/boards/board_obelix.c
```diff
@@ -576,8 +576,14 @@ const BoardConfigPower BOARD_CONFIG_POWER = {
...
 const BoardConfig BOARD_CONFIG = {
   .backlight_on_percent = 45,
-  .ambient_light_dark_threshold = 150,
-  .ambient_k_delta_threshold = 25,
+  // W1160 raw ALS counts scale linearly with integration time, and IT was
+  // bumped from 16.83 ms to 99 ms. Empirically the dark threshold lands
+  // around 1400 -- well above the ~9x naive scaling, because typical indoor
+  // light in our test scenes drives counts higher than the closed-PR
+  // prototype suggested. k_delta is held proportional (dark/7) so the
+  // AmbientLightLevel enum buckets keep similar relative width.
```
gmarull: too much details, keep in commit message, also test getafix

jplexer:
Merging because last comment was only re: getafix testing and comment and those have been resolve
and I want to cut a release

PR #1344, opened 2026-05-20
backlight: recover from silent AW2016 chip resets

jplexer:
The AW2016 backlight driver can silently power-cycle on a VBAT dip or I2C watchdog with no signal back to the MCU, leaving the chip stuck off while firmware still thinks the light is on; add a backlight_refresh() API and call it from the light service on no-op wake transitions so user taps/wrist-flicks heal the desync instead of getting absorbed by the fade-out timer (fixes FIRM-1929)

### DIFF
--- a/src/fw/drivers/backlight.h
+++ b/src/fw/drivers/backlight.h
@@ -18,6 +18,10 @@ void backlight_init(void);
 //! @param brightness Brightness level (0-100)
 void backlight_set_brightness(uint8_t brightness);
 
+//! Re-apply the cached driver state to the backlight hardware. No-op if the
+//! firmware believes the backlight is already off.
+void backlight_refresh(void);
+
 #ifdef CONFIG_BACKLIGHT_HAS_COLOR
 void backlight_set_color(uint32_t rgb_color);
 
--- a/src/fw/services/light/service.c
+++ b/src/fw/services/light/service.c
@@ -285,6 +285,11 @@ static void prv_change_state(BacklightState new_state) {
 
   if (s_current_brightness != new_brightness) {
     prv_change_brightness(new_brightness);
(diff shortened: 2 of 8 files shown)


src/fw/drivers/backlight.h
```diff
@@ -18,6 +18,14 @@ void backlight_init(void);
 //! @param brightness Brightness level (0-100)
 void backlight_set_brightness(uint8_t brightness);
 
+//! Re-apply the current driver state to the backlight hardware. Used to recover
+//! from silent chip resets on drivers (e.g. AW2016) that can power-cycle on a
+//! VBAT dip or I2C watchdog without any reset signal back to the MCU. No-op when
+//! the firmware believes the backlight is already off, or on drivers that can't
+//! desync from cached state. Safe to call from the light service whenever a wake
+//! action would otherwise be a no-op (FIRM-1929).
```
gmarull: too verbose, this is an interface, it just need to define the contract/use

src/fw/services/light/service.c
```diff
@@ -285,6 +285,13 @@ static void prv_change_state(BacklightState new_state) {
 
   if (s_current_brightness != new_brightness) {
     prv_change_brightness(new_brightness);
+  } else if (new_state == LIGHT_STATE_ON || new_state == LIGHT_STATE_ON_TIMED) {
+    // Wake-equivalent transitions normally short-circuit when brightness is
+    // unchanged. The AW2016 backlight driver can be silently power-cycled by a
+    // VBAT dip or I2C SCL watchdog with no reset signal back to the MCU, so
+    // we'd otherwise leave the chip stuck off while the firmware thinks it's
+    // on (FIRM-1929). Re-apply driver state on every user-initiated wake.
```
gmarull: we should not leak impl details at service level. Service level needs to work regardless of what's below, so details here are redundant

gmarull (approved):

PR #1348, opened 2026-05-20
fw/comm/ble/ancs: resubscribe when Notification Source goes silent

jplexer:
The ANCS alive-check fires a Control Point write every 15 minutes and treats iOS' reply as proof the watch is still receiving notifications. That probe only verifies the CP path. iOS can wedge in a state where it keeps responding to CP requests but its Notification Source push has been silently dropped (commonly after airplane-mode toggles and timezone swings during travel), so the watch reports itself healthy while no notifications arrive. Re-opening the iOS Pebble app is currently the only way to recover, which is what the FIRM-1692 reporter had to do after a 7-day silent outage.

Track how many consecutive alive-check intervals pass without any NS traffic. After ANCS_NO_NS_TIMEOUT_HOURS (6 h, derived from the alive-check interval so the two don't desync) we force a CCCD re-subscribe to nudge iOS into pushing again, skip the CP probe that round to avoid interleaving with the CCCD writes, and schedule the next alive check as a backup in case the CCCD response never lands. Any NS event (add/modify/remove) resets the counter, so users with normal notification volume never see the refresh fire; users with genuinely quiet phones absorb at most one harmless CCCD rewrite per 6 h window.

Fixes FIRM-1692

### DIFF
--- a/src/fw/comm/ble/kernel_le_client/ancs/ancs.c
+++ b/src/fw/comm/ble/kernel_le_client/ancs/ancs.c
@@ -101,6 +101,9 @@ typedef struct ANCSClient {
   NotificationQueueNode *queue;
   bool alive_check_pending;
   ANCSVersion version;
+  // Alive-check intervals with no NS traffic; reset by any NS notification.
+  // Triggers a CCCD re-subscribe at the threshold to recover silent NS.
+  uint8_t alive_checks_without_ns;
 } ANCSClient;
 
 static ANCSClient *s_ancs_client;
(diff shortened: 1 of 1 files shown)


src/fw/comm/ble/kernel_le_client/ancs/ancs.c
```diff
@@ -101,6 +101,13 @@ typedef struct ANCSClient {
   NotificationQueueNode *queue;
   bool alive_check_pending;
   ANCSVersion version;
+  // Counts alive-check intervals during which no Notification Source traffic
+  // arrived. When iOS' ANCS NS subscription silently wedges (CP keeps replying
+  // but NS push stops -- see FIRM-1692), the alive check is blind to it. After
+  // ANCS_ALIVE_CHECKS_WITHOUT_NS_BEFORE_RESUBSCRIBE consecutive silent
+  // intervals we re-write the CCCDs to nudge iOS into resuming pushes. Reset
+  // to 0 on every NS notification.
```
gmarull: shorten this, avoid refs to internal issues in the code

gmarull (approved):

PR #1352, opened 2026-05-21
fw/popups/phone_ui: fit caller-id frame to font height

jplexer:
Grow the caller-id text layer's frame alongside its bounds so the
rendered text isn't clipped when a larger title font pushes past the
style's static `caller_id_height` (e.g. GOTHIC_28_BOLD at
PreferredContentSizeLarge needs ~66px vs the static 60px).

Fixes FIRM-1944

### DIFF
--- a/src/fw/popups/phone_ui.c
+++ b/src/fw/popups/phone_ui.c
@@ -5,6 +5,7 @@
 #include "phone_formatting.h"
 
 #include "applib/fonts/fonts.h"
+#include "util/math.h"
 #include "util/trig.h"
 #include "applib/ui/action_bar_layer.h"
 #include "applib/ui/kino/kino_layer.h"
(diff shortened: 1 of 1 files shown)


src/fw/popups/phone_ui.c
```diff
@@ -416,8 +417,13 @@ static void prv_set_caller_id_text(PebblePhoneCaller *caller) {
   } else {
     s_phone_ui_data->caller_id_text_layer.layer.bounds.origin.y = DOUBLE_LINE_BOUND_OFFSET;
   }
-  s_phone_ui_data->caller_id_text_layer.layer.bounds.size.h =
-      lines * fonts_get_font_height(caller_id_font);
+  const int16_t content_height = lines * fonts_get_font_height(caller_id_font);
+  s_phone_ui_data->caller_id_text_layer.layer.bounds.size.h = content_height;
+  // Grow the frame to fit the chosen font so the bottom of the text isn't clipped — the
+  // style's static caller_id_height is sized for the default title font, but larger
+  // content-size settings (e.g. GOTHIC_28_BOLD at PreferredContentSizeLarge) push past it.
```
gmarull: we should be less verbose in comments, explain details in commit (classic from Claude, I'll tweak claude.md)

gmarull (approved):

PR #1353, opened 2026-05-21
Even more cleanup

jplexer:


src/fw/Kconfig
```diff
@@ -5,3 +5,49 @@ rsource "soc/Kconfig"
...
+      Expose the device-orientation / rotation prefs in shell and
+      Settings. Independent of CONFIG_SERVICE_ORIENTATION_MANAGER,
+      which controls whether the underlying service is built.
+
+config MODDABLE_XS
+    bool "Moddable XS JavaScript engine"
+    help
+      Build Moddable XS support for running JavaScript apps. The
+      effective JS engine can still be overridden at configure time
+      via --js-engine; PRF builds force it off.
```
gmarull: Stale comment (js-engine is removed). For PRF: simply do not set it to `y` in the prf .conf file.

gmarull: However, we do not have yet per-board settings at project level, so the PRF hack is still ok.

Kconfig
```diff
@@ -23,6 +23,90 @@ config IS_BIGBOARD
...
+choice
+    prompt "Board"
+    help
+      Hardware board variant. Boards are grouped into families
+      (BOARD_FAMILY_*) by SoC and HW design; some boards share a
+      family, while QEMU boards are 1-to-1 with their symbol.
+
+config BOARD_ASTERIX
+    bool "Asterix"
+    select BOARD_FAMILY_ASTERIX
```
gmarull: We should push all board Kconfig at each board directory (see Zephyr as a reference)

include/pbl/services/compositor/compositor_transitions.h
```diff
@@ -8,7 +8,7 @@
 #include "default/compositor_slide_transitions.h"
 #include "default/compositor_shutter_transitions.h"
 #include "legacy/compositor_app_slide_transitions.h"
-#if PLATFORM_ASTERIX
+#ifdef CONFIG_BOARD_FAMILY_ASTERIX
```
gmarull: We should avoid board level Kconfig in such places, this is probably more a "family" setting, or, a Flint choice (Flint would be any watch that behaves/is compatible with an asterix). I know it's now a mess, but still, worth fixing (or split to another PR)

gmarull (commented):
good stuff, see my comments. we'll need a few iterations to get this right (zephyr level), I'm okay merging intermediate stages though. maybe some can be iterated further now.

include/pbl/services/compositor/compositor_transitions.h
jplexer: - i think this one is actually even more simpler: black and white watches

Kconfig
```diff
@@ -23,6 +23,64 @@ config IS_BIGBOARD
...
+
+config QEMU_MACHINE
+    string
+    default "pebble-emery" if BOARD_QEMU_EMERY
+    default "pebble-flint" if BOARD_QEMU_FLINT
+    default "pebble-gabbro" if BOARD_QEMU_GABBRO
+    depends on QEMU
+    help
+      QEMU machine identifier passed to qemu-system-arm via the
+      `-machine` flag.
```
gmarull: we should avoid that: at the time we load Kconfig, we know the board (from `--board`), just pass it as env var $BOARD to Kconfig, see Zephyr

Kconfig
```diff
@@ -23,6 +23,48 @@ config IS_BIGBOARD
...
+      Hidden symbol selected by boards in the Getafix family.
+
+config BOARD_FAMILY_NAME
+    string
+    default "asterix" if BOARD_FAMILY_ASTERIX
+    default "obelix" if BOARD_FAMILY_OBELIX
+    default "getafix" if BOARD_FAMILY_GETAFIX
+    default "qemu_emery" if BOARD_QEMU_EMERY
+    default "qemu_flint" if BOARD_QEMU_FLINT
+    default "qemu_gabbro" if BOARD_QEMU_GABBRO
```
gmarull: modify default in board Kconfig.defconfig? see Zephyr patterns

gmarull (approved):

PR #1308, opened 2026-05-17
status_bar_layer: Draw clock with bold fonts by default

aveao:
The main way of rendering time in apps, status_bar_layer, has always used a tiny font and made it very difficult to view the time when e.g. there's a notification on screen. This has been one of my bigger pet peeves over the years.

This PR makes it render in a bold font, which is a starting point overall, which helps with making them more visible from a comfortable distance. To not break backwards compat with apps that utilize status_bar_layer for custom text, this PR only changes the font if StatusBarLayerMode is StatusBarLayerModeClock (the default).

---

I do think that for obelix and gabbro, using a larger font (24_BOLD vs 18_BOLD with this PR vs 18 on main rn) would be an improvement still. Hop on discord: https://discord.com/channels/221364737269694464/1412426648384438403/1505568794616135832

PR #1369, opened 2026-05-22
Introduce an option to resize status bar on notifications to a larger font

### LINKED ISSUE #1323: status_bar_layer: Increase size to 24 from 18 on obelix for better visibility
example pics: https://github.com/coredevices/PebbleOS/pull/1308#issuecomment-4475609657

eric seal of approval: https://github.com/coredevices/PebbleOS/pull/1308#issuecomment-4482649184

ideas on approaches:

- could expose options for status bar height and font size within the StatusBarLayerConfig, [like this](https://github.com/aveao/PebbleOS/commit/f1d816960534bffcf2fab84b2a20a74cab428b82), then system apps can move to setting it to 24, then apps can move to it as they choose. I think this'd be generally a good addition, but it's an SDK change and it's not guaranteed that apps would notice this and move over. It'd be best to do a more forceful move, so I like JP's idea for this:
- [JP's initial idea on discord](https://discord.com/channels/221364737269694464/1412426648384438403/1505576099932012645): "I guess what you could do is give apps on the current SDK revision the old size and then bump the SDK revision for the new size so that new apps would then inherit the new size"

aveao:
Closes #1323

Building on top of https://github.com/coredevices/PebbleOS/pull/1356 and https://github.com/coredevices/PebbleOS/pull/1308

This adds a new option to the newly added status bar setting* on notifications to allow setting it to "Big & Bold", which is one font size higher on all platforms. For obelix and gabbro, this is going from 18 to 24 point, and IMO it looks great. This is what I've been using for days and it's very preferable, unlike 18 bold the spacing is better as well so numbers don't risk blending together.

(*: also renamed it to "Clock Style")

---

Let me know if you have a better idea for the `status_bar_layer_set_mode`, it's not super optimal but seemed necessary to resize on mode change, as that doesn't trigger layout change events. Same with if you don't think this needs a golden and as such the import approach on tests can be changed.

---

Untested in person on flint or gabbro (only emulator). Please give it a try!


gmarull (approved):

jplexer (approved):

ericmigi:
I think this looks much better than the bold option. Let's try it out but if it works is there any reason to have the smaller text on gabbro and obelix - should this just be the new default and only option? I can't really think of a reason to have a smaller font here since it doesn't really have any benefit right?

aveao:
Very glad you like it.

Since day 1?2? of getting my PT2 I've been using 18 bold OS-wide which is more of a matter of taste (I find it helps but the distance between letters is lower on 18, I can see that being a problem for some, especially with anything like astigmatism, speaking as someone with quite a bit of it), and 24 bold on notifications (which to me is a clear winner).

I don't see why I'd drop down to the smaller size, I suspect most would prefer the larger size, but getting some added opinions may be worthwhile. The sizing does currently make the clock look a bit large compared to the rest of the notifications UI if set to default text size, personally the function of seeing the time is more valuable than the aesthetics, but it being an option also leaves that choice to the users even if the default would be this option. Up to you ultimately, I wouldn't mind this being the only option.

aveao:
Oh, above is for obelix.

This PR adds it for asterix and gabbro too, I can see the screen real estate being more in demand for smaller devices (but this was something I wished for even on the PTS)

And I would honestly suggest to think about making gabbro's default status bar larger and making the option for larger clock even larger than this. It's hard to predict how it'd look irl but based on the pictures I've seen, the status bar looks tiny on that screen.

PR #1378, opened 2026-05-26
Music App Prefs: Allow disabling priority in launcher, configure vibe strength

nullstalgia:
Howdy! Happy to rejoin the Pebble family, but in open source this time! It was really fun editing the source of my favorite smartwatch, looking forward to doing more!

I keep Music bound to "Quick Launch: Hold Select", so the prioritization in the Launcher just gets annoying when the app at the top gets shifted without my consent.

Additionally the HAPTIC_FEEDBACK vibe score uses a short 100% pattern, and I felt that was a little aggressive (especially with the repeating vibes from holding Up/Down), so I've added an option to allow changing it with the mobile app's settings page.

---

My work is based off of PR #1071, and in which I noticed the desire to bundle these preferences by application. I even attempted to do so within [the first commit](https://github.com/coredevices/PebbleOS/commit/ba5bf727c4f3937830aeb9e883b9b513b1299402), consolidating:

```
  "musicShowVolumeControls",
  "musicShowProgressBar",
  "musicPrioritizeWhenPlaying",
  "musicLongPressVibeIntensity",
```

into

```
  "musicAppPreferences",
```

***But***, I don't know any Kotlin! (I'm a C and Rust fellow!) So following the flow of something like `"hrmPreferences"` is non-trivial for me at the moment. **I'd be more than happy to help return it to a consolidated struct**, but I'd need someone to help me with the Kotlin side of things!

### DIFF
--- a/src/fw/services/blob_db/settings_blob_db.c
+++ b/src/fw/services/blob_db/settings_blob_db.c
@@ -109,8 +109,7 @@ static const char *s_syncable_settings[] = {
   "workerId",
 
   // Music preferences
-  "musicShowVolumeControls",
-  "musicShowProgressBar",
+  "musicAppPreferences",
 };
 
 static const size_t s_num_syncable_settings = ARRAY_LENGTH(s_syncable_settings);
--- a/src/fw/services/blob_db/settings_blob_db.c
+++ b/src/fw/services/blob_db/settings_blob_db.c
@@ -109,7 +109,10 @@ static const char *s_syncable_settings[] = {
   "workerId",
 
   // Music preferences
-  "musicAppPreferences",
+  "musicShowVolumeControls",
+  "musicShowProgressBar",
+  "musicPrioritizeWhenPlaying",
+  "musicLongPressVibeIntensity",
 };
 
 static const size_t s_num_syncable_settings = ARRAY_LENGTH(s_syncable_settings);
--- a/include/pbl/services/music.h
+++ b/include/pbl/services/music.h
@@ -7,6 +7,8 @@
 #include <stdint.h>
 #include <string.h>
 
+#include "util/attributes.h"
+
 #define MUSIC_BUFFER_LENGTH 64
 
 typedef enum {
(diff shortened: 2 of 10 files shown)


src/fw/services/blob_db/settings_blob_db.c
```diff
@@ -109,8 +109,7 @@ static const char *s_syncable_settings[] = {
   "workerId",
 
   // Music preferences
-  "musicShowVolumeControls",
-  "musicShowProgressBar",
+  "musicAppPreferences",
```
jplexer: we gotta be careful here, since the mobile app also needs a corresponding change

nullstalgia: I'm aware! And in case my description wasn't super clear, that isn't how the PR is _currently_ laid out, only in the first commit; later I reverted back to the old discrete entries + two new discrete preference keys, like so.

https://github.com/coredevices/PebbleOS/pull/1378/changes/90fe3fa782f916340191b830367d6e94df8c9fad#diff-2604530760c3fbc778755b27d105d89c5633f1621153c1fd6c844bfbbf08305dR112-R115

But to make myself clear once more, **I do think the single struct is the preferable method**, I simply do not have the requisite Kotlin knowledge to make the correct changes needed in the mobile app. I was hoping I could get some assistance on that side, as I'm far more comfortable with embedded firmware than mobile applications.

jplexer: @sjp4 can you help maybe :)

sjp4: A single struct sounds problematic for forwards/backwards compatibility (especially in fw). Adding individual settings keys will be much simpler from that perspective.

nullstalgia: I can see that. I just wanted to try to adhere to the comment stating ["`NEW PREFERENCES DO __NOT__ BELONG HERE WITHOUT A VERY GOOD REASON`"](https://github.com/nullstalgia/PebbleOS/blob/33f7bc4093bbb99e9a1d0f1102f6d5a355ea2aa2/src/fw/shell/prefs.h#L13), and avoid cluttering the namespace as much as possible. If you think it's acceptable, then I'm happy. Otherwise I'm also happy to help make changes. :)

nullstalgia:
Let me know if there's any action required from me, since otherwise this might be good to go? 

It looks like other PRs poking this module have come in though, haven't had time to look super closely at the conflicts but they seem trivial.

ericmigi:
sorry, probably not going to merge this. It's a bit too niche.

nullstalgia:
@ericmigi If I'm allowed to try to vouch one final time, 

I can acquiesce on the vibration strength part\*, but I do think the option to disable prioritizing the Music app in the launcher is worth considering on its own merit.

I try to keep my Quick Launch shortcuts and Launcher menu order as consistent as possible so I can burn them into muscle memory, and the Music app changing its position (to the top no less, offsetting every other item) really messes with that. Doubly so since I already have Music bound to the "Hold Center" Quick Launch. I'd be happy to trim the PR to just that.

\*: I'll admit that I am weirdly tactically sensitive, but I do also _way_ prefer how the feedback feels at 40% strength on my PT2 (Medium-Low on the Mobile PR). Plus there wasn't a way in the kernel previously to scale a vibe score's strength at runtime and that was fun to add, but I do also understand the merit of `static const` for as much as possible. A fork with instead a single new pattern for the feedback wouldn't be too hard for me to upkeep ;P

PR #1386, opened 2026-05-26
fw/kernel: pool TaskTimer allocations

jplexer:
TaskTimer creates churn very heavily on this device class (NimBLE allocates one callout TaskTimer per ble_npl_callout — ~55 alive at once just for the BLE stack — and total creates over a session can exceed 16k). Each create/free was a kernel_malloc/kernel_free of sizeof(TaskTimer) (40 B), which left 44 B holes scattered through the heap. Coredump from a fragmented Asterix shows 28 free blocks averaging 135 B; the dominant fragmentation source by count is the TaskTimer churn.

Serve TaskTimers from a fixed 192-slot slab pool kept in .kernel_bss instead. Create/delete now move slots on/off an intrusive free list with O(1) cost and never touch the heap allocator after init. The pool falls back to kernel_malloc on exhaustion so worst-case behaviour matches the old code rather than failing harder. 192 slots is ~1.7x observed peak (111) — the 2.4 KB of slab headroom above peak costs ~3 KB of kernel heap (.kernel_bss takes the RAM back from .kernel_heap by linker script) and trades it for eliminating per-TaskTimer fragmentation.

### DIFF
--- a/src/fw/kernel/Kconfig
+++ b/src/fw/kernel/Kconfig
@@ -1,0 +1,8 @@
+# SPDX-FileCopyrightText: 2026 Core Devices LLC
+# SPDX-License-Identifier: Apache-2.0
+
+config TASK_TIMER_POOL_SIZE
+    int "TaskTimer slab pool size"
+    default 192
+    help
+      Number of TaskTimer slots reserved in .kernel_bss
--- a/src/fw/kernel/task_timer.c
+++ b/src/fw/kernel/task_timer.c
@@ -5,7 +5,6 @@
 #include "task_timer_manager.h"
 
 #include "kernel/pebble_tasks.h"
-#include "kernel/pbl_malloc.h"
 #include "os/mutex.h"
 #include "os/tick.h"
 #include "system/logging.h"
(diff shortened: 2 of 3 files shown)


src/fw/kernel/task_timer.c
```diff
@@ -46,6 +46,66 @@ typedef struct TaskTimer {
   bool defer_delete:1;
 } TaskTimer;
 
+// TaskTimer slab pool. Avoids fragmenting the kernel heap with heavy
+// create/delete churn. Falls back to kernel_malloc if exhausted.
+#define TASK_TIMER_POOL_SIZE 192
```
gmarull: Should be Kconfigurable

src/fw/kernel/task_timer.c
```diff
@@ -46,6 +46,66 @@ typedef struct TaskTimer {
...
+  s_task_timer_pool_free_head = timer;
+  mutex_unlock(s_task_timer_pool_mutex);
+}
+
+static TaskTimer *prv_timer_alloc(void) {
+  TaskTimer *timer = prv_pool_alloc();
+  if (timer) {
+    return timer;
+  }
+  PBL_LOG_WRN("TaskTimer pool exhausted, falling back to kernel_malloc");
```
gmarull: Not a good idea, if we fail, pool needs to be redimensioned

jplexer:
github I pushed a commit 20 hours ago!!!

src/fw/kernel/Kconfig
```diff
@@ -0,0 +1,11 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+config TASK_TIMER_POOL_SIZE
+    int "TaskTimer slab pool size"
+    default 192
+    help
+      Number of TaskTimer slots reserved in .kernel_bss. TaskTimer
+      churn (e.g. NimBLE callouts) would otherwise fragment the kernel
+      heap. Pool exhaustion is fatal — size this above observed peak
+      usage. Each slot costs sizeof(TaskTimer) bytes of kernel BSS.
```
gmarull: too many impl details, this should be simple

gmarull (approved):

PR #1387, opened 2026-05-26
resources/lang: avoid glyphs missing from LECO fonts in ca/es

jplexer:
The LECO_*_BOLD_NUMBERS fonts used to render values in the sleep and heart-rate summary cards include only digits, uppercase letters and a small punctuation set. Translations that fall outside that set render as replacement boxes:

- ca_ES translates the time-unit strings "H" / "M" and the formats "%dH" / "%dM" to lowercase ("h" / "m", "%d h" / "%d min"). The sleep summary card showed garbled glyphs in place of the unit suffixes.
- es_ES translates "BPM" to "LAT/MIN". The heart-rate summary card showed "85 ▢A▢/MI▢" because L, T and N are missing from LECO_26_BOLD_NUMBERS_AM_PM. Reuse "PPM", which already matches the "%d BPM" → "%d PPM" translation elsewhere.

Fixes FIRM-2032
Fixes FIRM-2056

### DIFF
--- a/resources/normal/base/lang/ca_ES/tintin.po
+++ b/resources/normal/base/lang/ca_ES/tintin.po
@@ -2329,12 +2329,12 @@ msgstr "FC mitjana"
 #: ../src/fw/services/activity/health_util.c:32
 #, c-format
 msgid "%dH"
-msgstr "%d h"
+msgstr "%dH"
 
 #: ../src/fw/services/activity/health_util.c:38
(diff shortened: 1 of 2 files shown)


gmarull (changes requested):
We should extend font set instead

gmarull:
related https://github.com/coredevices/PebbleOS/issues/1399

PR #1402, opened 2026-05-28
add system wide volume limit

jplexer:


src/fw/apps/system/settings/vibe_patterns.c
```diff
@@ -160,6 +175,34 @@ static void prv_selection_changed_cb(SettingsCallbacks *context, uint16_t new_ro
   vibe_score_destroy(score);
 }
 
+#ifdef CONFIG_SPEAKER
```
gmarull, line 178: we should ifdef using the right kconfigs, SERVICE_SPEAKER here?

jplexer: CONFIG_SERVICE_SPEAKER is set globally in prj_normal.conf so it wouldn't gate per-board

gmarull: oh, we should then fix that as well. prj_normal.conf should just contain common settings, speaker is definitely not. We could have per-board overlays at firmware level like Zephyr. Can be done as a follow-up.

gmarull (approved):

PR #1431, opened 2026-06-03
fw/drivers/hrm/gh3x2x: adjust fifo water marker

ck-telecom:
increase gh3x2x fifo water marker to max 800, less interrupt to host wakeups.


gmarull:
will check on HW

PR #1435, opened 2026-06-03
fw/apps/system: Add an accel_tap service to show progress bar for 5s

FouzR:
Adds a accel tap service that temporarily shows the progress bar for 5 seconds if the show progress bar setting is unchecked.

AI(Claude Sonnet 4.6 Low) was only used for troubleshooting build issues, so there might be a few odd things here and there that _might_ be bandaid fixes =p. Though the main logic was implemented by me 

Verified on an obelix_pvt device with -DCONFIG_RELEASE=y

### DIFF
--- a/src/fw/apps/system/music.c
+++ b/src/fw/apps/system/music.c
@@ -6,6 +6,7 @@
 #include "applib/app.h"
 #include "applib/event_service_client.h"
 #include "applib/app_timer.h"
+#include "applib/accel_service.h"
 #include "applib/fonts/fonts.h"
 #include "applib/preferred_content_size.h"
 #include "applib/tick_timer_service.h"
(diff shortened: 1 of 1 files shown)


src/fw/apps/system/music.c
```diff
@@ -784,9 +815,12 @@ static void prv_update_track_progress(MusicAppData *data) {
   if (data->pause_track_pos_updates) {
     return;
   }
-  if (!shell_prefs_get_music_show_progress_bar()) {
+
+  if (!data->temporarily_show_progress){
+  if (!shell_prefs_get_music_show_progress_bar()){
```
jplexer: && instead of nested?

jplexer (commented):
if I get it correctly the accel tap is being subscribed even if progress bar is supposed to show at all times, maybe make it a setting?

FouzR:
Yeah, I can make it a setting

Currently, it just subs if showing progress is present. I could also tie it to when the progress bar view is disabled, which should suffice I think

FouzR:
@jplexer I believe these changes should be enough?

FouzR:
(accidentally messed up that logic, now it does work, and is tested on the hardware)

FouzR:
Rebased and squashed to one commit now 👍

jplexer (approved):

PR #1436, opened 2026-06-03
Quiet time improvements

jmsunseri:
https://github.com/user-attachments/assets/e466074e-9e67-4079-98ba-f0d8e0d280ca


gmarull (changes requested):
please, follow commit guidelines and push PRs that are reviewable (e.g. scoped enough)

jmsunseri:
> please, follow commit guidelines and push PRs that are reviewable (e.g. scoped enough)

I can move the day picker refactor into a PR of its own.  Would that be sufficient?

jmsunseri:
@gmarull now that the other PR is merged this one is a little smaller and i did a bit of a refactor to try and limit the size of the MR while keeping it feature complete.  There were other more aggressive improvements possible but hopefully this struck a nice balance.

jmsunseri:
any update on this?

ericmigi:
pinged Gerard

jmsunseri:
It's not my fault it's stale I wanted it merged months ago.  It's very discourage to see hard work go to waste

ericmigi:
been busy (gerard is also off so we're down to 1 fw eng). we'll get to it (at some point!). It's also cosmetic (unless I'm misreading it)

jmsunseri:
Cosmetic?  No I wouldn't call it cosmetic. With the current code it's
impossible I set up a quiet time that's as customized as the alarms. In
this MR we are borrowing the customizing functionality from the alarms UI
and giving it to quiet time.


Sincerely,
Justin Sunseri

ericmigi:
@sjp4 are we syncing quiet time now? this will need a corresponding mobileapp change too right?

gmarull:
please rebase

jmsunseri:
> please rebase

done.  New model wanted to make a couple small updates to cover and edge case and make some additional adjustments and tests.  I have not had a chance to play with this thoroughly but I plan to do so over the next day

ericmigi:
@gmarull ready to merge?

src/fw/services/notifications/do_not_disturb.c
```diff
@@ -119,14 +124,100 @@ static void prv_push_manual_dnd_first_use_dialog(ManualDNDFirstUseSource source)
...
+}
+
+static bool prv_is_any_qt_schedule_active_now(void) {
+  struct tm time;
+  rtc_get_time_tm(&time);
+  for (int i = 0; i < MAX_QUIET_TIME_SCHEDULES; i++) {
+    if (!s_qt_schedule_cache[i].is_used || !s_qt_schedule_cache[i].enabled) continue;
+    bool days[DAYS_PER_WEEK];
+    quiet_time_get_scheduled_days(&s_qt_schedule_cache[i], days);
+    if (!days[time.tm_wday]) continue;
```
gmarull, line 206: Schedules that cross midnight stop at midnight on custom day sets. Gating on `days[time.tm_wday]` before checking the window means a "Wednesdays 22:00-07:00" schedule is quiet on Wednesday 00:00-07:00 and 22:00-24:00, and NOT quiet at Thursday 02:00. Legacy had the same quirk only at the weekday/weekend boundary so it went unnoticed; with single custom days it is very visible.

A wrapping window should belong to the day it started on:

```c
static bool prv_schedule_is_active(const struct tm *now, const QuietTimeScheduleConfig *s) {
  bool days[DAYS_PER_WEEK];
  quiet_time_get_scheduled_days(s, days);
  int now_m = now->tm_hour * 60 + now->tm_min;
  int from_m = s->from_hour * 60 + s->from_minute;
  int to_m = prv_schedule_end_minutes(s);
  if (from_m <= to_m) {
    return days[now->tm_wday] && now_m >= from_m && now_m < to_m;
  }
  int yesterday = (now->tm_wday + DAYS_PER_WEEK - 1) % DAYS_PER_WEEK;
  return (days[now->tm_wday] && now_m >= from_m) || (days[yesterday] && now_m < to_m);
}
```

`prv_set_schedule_mode_timer` has the matching flaw: it adds a midnight transition instead of the end time on the day after the start day. Compute per schedule the seconds until `from` on the next scheduled day and the seconds until `to` (next day if wrapping), take the minimum, and drop the midnight entry. Please add tests for Saturday 01:00 on a Weekdays schedule (inactive) and Monday 01:00 (inactive).

src/fw/services/notifications/do_not_disturb.c
```diff
@@ -284,6 +366,21 @@ void do_not_disturb_get_schedule(DoNotDisturbScheduleType type,
 
 void do_not_disturb_set_schedule(DoNotDisturbScheduleType type, DoNotDisturbSchedule *schedule) {
```
gmarull, line 367: `do_not_disturb_set_schedule`, `do_not_disturb_set_schedule_enabled` and `do_not_disturb_toggle_scheduled` are now only called from tests, and each one writes slot 0/1 unconditionally. If a user deletes the migrated weekday schedule and creates a custom one, it lands in slot 0 and any future call here overwrites it.

Please remove the three setters and the dual-write, keep the legacy keys read-only for migration, and port the tests that use them to the `quiet_time_*` API.

src/fw/services/notifications/alerts_preferences.c
```diff
@@ -730,6 +829,27 @@ void alerts_preferences_handle_blob_db_event(PebbleBlobDBEvent *event) {
...
+      QuietTimeScheduleConfig _tmp;
+      if (settings_file_get(&file, key, key_len, &_tmp, sizeof(_tmp)) == S_SUCCESS) {
+        s_qt_schedule[i] = _tmp;
+        matched_key = qt_key;
+      }
+      goto done;
+    }
+  }
+
+  // Legacy DND schedule keys (dndWeekdaySchedule, etc.) are already handled by
```
gmarull, line 847: With this PR, phone writes to `dndWeekdaySchedule` / `dndWeekendSchedule` (still in the syncable list) reload the legacy struct and kick the DND service, but the service only evaluates the QT slots. Editing the weekday schedule from the mobile app becomes a silent no-op.

Either map legacy inserts onto slot 0/1 when that slot still has kind Weekdays/Weekends, or drop the legacy keys from the syncable list and land `qtSchedule*` support in the mobile app first. This is the question raised earlier in the thread and it needs an answer before merge.

include/pbl/services/notifications/do_not_disturb.h
```diff
@@ -22,6 +23,30 @@ typedef struct PACKED DoNotDisturbSchedule {
...
+  QT_KIND_CUSTOM,
+} QuietTimeKind;
+
+//! QuietTimeScheduleConfig stores a single quiet time schedule slot.
+//! The scheduled_days array is always present but only meaningful when kind == QT_KIND_CUSTOM.
+//! For other kinds, the day mask is derived from the kind at runtime.
+//! is_used indicates whether the slot contains a valid schedule (unused slots are zeroed out).
+typedef struct PACKED QuietTimeScheduleConfig {
+  bool is_used;
+  QuietTimeKind kind;
```
gmarull, line 41: This struct is persisted and synced to the phone, but its layout is compiler-defined: `QuietTimeKind` is an enum (4 bytes on ARM and host today, nothing pins it) and `bool` is not guaranteed to be 1 byte either. Please use `uint8_t kind` and `uint8_t` for the flags so the wire format is explicit, and treat a `settings_file_get` size mismatch on load as an error rather than silently dropping the slot.

src/fw/services/notifications/alerts_preferences.c
```diff
@@ -172,6 +182,52 @@ static void prv_migrate_legacy_dnd_schedule(SettingsFile *file) {
   }
 }
 
+static void prv_migrate_qt_schedules(SettingsFile *file) {
+  if (settings_file_exists(file, s_qt_schedule_keys[0], strlen(s_qt_schedule_keys[0]))) {
```
gmarull, line 186: The migration marker is fragile: this re-runs on every boot until `qtSchedule0` exists, but a slot is only written when its legacy schedule looks non-default. Weekday at default plus weekend customised means slot 1 gets rewritten from legacy values at every boot, undoing edits or deletes made in the new UI.

Always write both keys (an empty slot is fine), or use a dedicated migration flag.

gmarull (changes requested):
Thanks for the rebase. Overall direction is good, but a few things need to be fixed before this can go in. Inline comments below cover the blocking ones; I have a longer list of non-blocking cleanups (unused API, dead includes, the deferred window push, unrelated blob_db fixes that should be their own PR, and history that does not bisect) that I can add once these are addressed.

PR #1440, opened 2026-06-05
fw/apps/system/settings: Add Quick Lanuch Combo settings in watch

FouzR:
Adds quick launchsettings on the watch for the two Combo actions(Back+Up and Up+Down). This is to separate out this setting from the mobile app ~~and allow users not using Core's Pebble App to set it~~

Verified by building it on obelix_pvt and qemu_emery(builds but can't use combo quick launch. Assuming something is up with how qemu handles button inputs)

### DIFF
--- a/src/fw/apps/system/settings/quick_launch.c
+++ b/src/fw/apps/system/settings/quick_launch.c
@@ -25,7 +25,7 @@
 #include "system/passert.h"
 #include "system/status_codes.h"
 
-#define NUM_ROWS (NUM_BUTTONS + 2)  // 4 hold buttons + 2 tap buttons (up and down)
+#define NUM_ROWS (NUM_BUTTONS + 2 + 2)  // 4 hold buttons + 2 tap buttons (up and down) + 2 Combo
 
 typedef enum {
   ROW_TAP_UP = 0,
@@ -34,6 +34,8 @@ typedef enum {
   ROW_HOLD_SELECT,
   ROW_HOLD_DOWN,
   ROW_HOLD_BACK,
+  ROW_HOLD_BACK_UP,
+  ROW_HOLD_UP_DOWN,
 } QuickLaunchRow;
 
 typedef struct QuickLaunchData {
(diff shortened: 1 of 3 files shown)


FouzR:
IMG_5880


Picture of it working in hardware

src/fw/apps/system/settings/quick_launch.c
```diff
@@ -130,37 +140,53 @@ static void prv_select_click_cb(SettingsCallbacks *context, uint16_t row) {
...
+      button = BUTTON_ID_DOWN;
+      is_tap = false;
+      is_combo = true;
       break;
     default:
       return;
   }
   
-  quick_launch_app_menu_window_push(button, is_tap);
+  quick_launch_app_menu_window_push(button, is_tap, is_combo);
```
gmarull, line 189: it would be cleaner to convert button to a button bitmap (and adjust ButtonId enum so that it is directly bit mappable)

FouzR: So like say, 4 bits, each bit corresponding to a button and setting bits per button pressed, and using that to find the combo? Which would also allow for more combinations(and a possible UP+SELECT+DOWN)?

gmarull: correct

FouzR: I also noticed that all the combo get/set/enabled/is_enabled functions are just specific to the two combos available, unlike all the other quick lanuch actions

would it be preferrable to squash them into a single function and pass the button bitmap

sjp4:
These were deliberately not included in the watch settings - designed to be mobile-only configurable

FouzR:
Oh? Why have some be mobile only? I can understand if setting it on the watch is cumbersome(like the RGB backlight) but it makes no sense when you already have quick launch on the watch for the other buttons and not have all of them

sjp4:
There are too many settings on the watch - the menu-diving and scrolling becomes a bad experience, so we have to draw the line somewhere (in this case, the distinction is "advanced" settings which most users won't touch).

You said that you want to be able to set these but don't use the core app - which app are you using? libpebble3 supports these fully, so any app should be able to set them.

FouzR:
Oh I didn't actually test those apps. I do use the Core app, but prefer setting things on the watch :p

(also it was a bad assumption on my end, since I heard stuff like the RGB backlight wasn't in micropebble, so maybe there needs to be PRs there to add em)

I do get it being an impediment to the UX on the watch and having too many options isn't a good idea

Maybe there could a middle ground "advanced options" thing that shows these additional options(instead of clumping all of them into debugging)

sjp4:
We're not averse to the idea of a "i'm an advanced user" toggle kind of thing

FouzR:
So it would either need to be its own setting menu(with maybe a mobile app toggle with all the warnings needed) or it would have to be a pref that's checked by each menu to display the elements(so as to have the relevant options in their menus, like the motion sensitivity for backlight, which already is behind "debug" for the mobile app side of things)

sjp4:
Probably should discuss this on a new issue

FouzR:
> Probably should discuss this on a new issue

Just want clarification why the _PR_ was closed but you mentioned it as an issue 

we can have the discussion about advanced settings in general on an issue, but I still think for this, it's fine to have it on the watch(once I clean up the PR)

PR #1446, opened 2026-06-06
fw/kernel/util: add hibernate mode for SF32LB

ck-telecom:


gmarull (commented):
we should have a generic API, implemented at soc level

src/fw/soc/sf32lb/sf32lb52x/hibernate.c
```diff
@@ -0,0 +1,13 @@
...
+
+#include "util/attributes.h"
+
+#include <bf0_hal.h>
+
+NORETURN enter_hibernate(void) {
+  HAL_PMU_EnterHibernate();
+
+  /* Should not reach here */
+  while (1) {}
```
gmarull: use code unreachable (add macro if not available)

gmarull (commented):
this needs better polish, a platform can select `HAS_HIBERNATE` Kconfig, `HIBERNATE` depend on it?

src/fw/soc/Kconfig
```diff
@@ -18,6 +18,13 @@ config SOC_QEMU
 
 endchoice
 
+config HIBERNATE
+    bool "Enable hibernate mode"
+    default y if BOARD_FAMILY_OBELIX
```
gmarull: -1, this doesn't scale. Please, check how shutdown works in Zephyr for a good reference...

ck-telecom: could you give me a hint? I am not sure how scale this interface

gmarull: Just check Zephyr, there's shutdown API for a well designed case.

src/fw/soc/sf32lb/sf32lb52x/soc_power.c
```diff
@@ -0,0 +1,14 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "bf0_hal.h"
+
+#include "drivers/soc_power.h"
+
+static void sf32lb52_enter_hibernate(void) {
```
ck-telecom: TODO: here we shoud set wakeup pin(if not the same pin in each platform, maybe need board level?), but I didn't kown which pin coz I don't got any sch

gmarull (changes requested):
This PR is now mixing things, pmic shutdown and new hibernate feature. Not what we need. Leave shutdown as is, no need to change that. We want a NEW hibernate feature, modeled after Zephyr shutdown, also available from settings menu as is shutdown.

ck-telecom:
Could you please check this PR? I test it on my watch and it works, check time before shutdown, then check time after power up the watch

ericmigi:
Thanks for the PR @ck-telecom! Just checking if you addressed Gerard's feedback above?

include/pbl/soc/sf32lb/hibernate.h
```diff
@@ -0,0 +1,6 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+NORETURN enter_hibernate(void);
```
gmarull: definitely not the place for this header, this is a platform agnostic interface, why in soc?

soc/sf32lb/sf32lb52x/hibernate.c
```diff
@@ -0,0 +1,20 @@
...
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "util/attributes.h"
+
+#include <bf0_hal.h>
+
+static void prv_set_wakeup_source() {
+  HAL_PMU_SelectWakeupPin(0, HAL_HPAON_QueryWakeupPin(hwp_gpio1, 34));
+  HAL_PMU_EnablePinWakeup(0, AON_PIN_MODE_HIGH);
+  hwp_pmuc->WKUP_CNT = 0x50005;
```
gmarull: no magic numbers pls

soc/Kconfig
```diff
@@ -53,3 +54,16 @@ config MPU_TYPE_ARMV8M
...
+    help
+      Platform supports deep hibernate mode for standby. SoCs that
+      provide this capability should select this symbol.
+
+config HIBERNATE
+    bool "Enable hibernate mode"
+    depends on HAS_HIBERNATE
+    help
+      Use deep hibernate mode when entering standby. This provides
+      better power savings.
```
gmarull: not a soc level Kconfig

src/fw/kernel/util/standby.c
```diff
@@ -9,6 +9,11 @@
 #include "system/logging.h"
 #include "system/reset.h"
 #include "system/passert.h"
+#ifdef CONFIG_HIBERNATE
+#include "pbl/soc/sf32lb/hibernate.h"
+#endif
```
gmarull: what I said: we're including soc level stuff in platform agnostic code (kernel standby)

src/fw/kernel/util/standby.c
```diff
@@ -23,6 +28,8 @@ static NORETURN prv_enter_standby(void) {
 }
 #endif
 
+#endif /* CONFIG_HIBERNATE */
```
gmarull: wrong: we want to keep both -> full shutdown AND soc shutdown/hibernate

gmarull (changes requested):
I think this still has some major design problems, and lacks e.g. a settings menu entry to trigger it.

ck-telecom:
@gmarull updated

PR #1447, opened 2026-06-06
apps: add quick launch music actions

gloriousCode:
Add Quick Launch-only system actions for music playback controls so users can bind long-press shortcuts to previous track, next track, play/pause, volume up, and volume down. Each action sends the existing music service command and exits with an action-performed reason so the shell returns to the watchface.

I don't want to have to open an additional app to change/play pause. Let me do it from the watch face!


src/fw/apps/system/music_actions.c
```diff
@@ -0,0 +1,99 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
```
jplexer: not authored by core devices llc

jplexer (commented):
nit regarding the copyright

src/fw/apps/system/music_actions.c
gloriousCode: Sorry about that!

PR #1448, opened 2026-06-06
workout: streamline selecting and starting a workout

s00se:
It currently takes upwards of 5 button presses to start a different workout. This reduces the amount of unnecessary button presses


asyba:
and how are you saving the amount of presses?

PR #1375, opened 2026-05-24
toggle: add quick launch shortcut to wipe notification history

s00se:

PR #1450, opened 2026-06-06
toggle: add quick launch shortcut to wipe notification history

s00se:
https://github.com/coredevices/PebbleOS/pull/1375


gmarull (commented):
commit still fails to have proper area description, and commit author seems != signed off

s00se:
I think this is good now? If not, please let me know with a little more specific information where this should live in the area description

jplexer (approved):

PR #1452, opened 2026-06-07
Make SDK a little bit more C++-friendly

imax9000:
By no means this introduces any actual support for C++. It only tweaks a couple of small things that g++ treats as errors and that cannot be reasonably resolved or worked around within an app project.

### DIFF
--- a/src/fw/applib/pebble_warn_unsupported_functions.h
+++ b/src/fw/applib/pebble_warn_unsupported_functions.h
@@ -44,6 +44,11 @@
 #define stat(...)   _Static_assert(0, "stat() is not supported")
 
 // List of unsupported memory calls
+
+// alloca() happens to be defined as a macro in the libc implementation
+// used in the SDK. #undef is a no-op if there's no such macro.
+#undef alloca
+
 #define alloca(...)   _Static_assert(0, "alloca() is not supported")
 #define mmap(...)     _Static_assert(0, "mmap() is not supported")
 #define brk(...)      _Static_assert(0, "brk() is not supported")
--- a/src/fw/applib/graphics/gcolor_definitions.h
+++ b/src/fw/applib/graphics/gcolor_definitions.h
@@ -18,12 +18,12 @@
 //! @param green Green value from 0 - 255
 //! @param blue Blue value from 0 - 255
 //! @param alpha Alpha value from 0 - 255
-//! @return GColor created from the RGB values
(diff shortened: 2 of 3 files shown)


src/fw/applib/pebble_warn_unsupported_functions.h
```diff
@@ -44,6 +44,9 @@
 #define stat(...)   _Static_assert(0, "stat() is not supported")
 
 // List of unsupported memory calls
+#ifdef alloca
+#undef alloca
+#endif
```
gmarull: this needs better explanation, e.g. why is alloca affected and not others

imax9000: Done

src/fw/applib/pebble_warn_unsupported_functions.h
```diff
@@ -44,6 +44,12 @@
 #define stat(...)   _Static_assert(0, "stat() is not supported")
 
 // List of unsupported memory calls
+
+// alloca() happens to be defined as a macro in glibc and
+// a few other libc implementations. #undef is a no-op if
+// there's no such macro.
+#undef alloca
```
gmarull: why is glibc relevant here?

imax9000: Headers under `toolchain/arm-none-eabi/arm-none-eabi/include` in the SDK appear to be coming from glibc or a fork of it. But point taken.

src/fw/applib/pebble_warn_unsupported_functions.h
```diff
@@ -44,6 +44,11 @@
 #define stat(...)   _Static_assert(0, "stat() is not supported")
 
 // List of unsupported memory calls
+
+// alloca() happens to be defined as a macro in the libc implementation
+// used in the SDK. #undef is a no-op if there's no such macro.
+#undef alloca
```
gmarull, line 50: I've checked, and `alloca` uses a builtin under `__GNUC__`, so in theory, it is supported. Maybe this was added because it was not desired (I checked and in 2016 it was also using a builtin). The problem is not C++ specific in fact.

Checked again: I'm fine with the solution

gmarull (approved):

imax9000:
Fixed lint errors. @gmarull , could you please approve again to unblock the checks?

jplexer (approved):

PR #1455, opened 2026-06-08
apps/system/timeline: fix timeline scrolling bugs

soxfox42:
After the changes in PebbleOS to add a full timeline app (the one with both past and future), I've found some buggy behaviour when scrolling the timeline, particularly when combined with Quick Launch.

Here's what I've found and fixed in this PR:
- Timeline Full uses the wrong animation direction when crossing the past/future boundary.
  Fixed by using the animation for the destination when switching.
- If Timeline Full is launched by either the Tap or Hold Up quick launch, it gets stuck in the past.
  Fixed by launching Timeline Future, instead of attempting to relaunch Timeline Full, which was blocked by the system.
- After Timeline Past is quick-launched, scrolling down takes you to the future, instead of back to the watchface.
  Fixed by correctly setting the launch arguments for all timeline quick launches.

Happy to split this into multiple PRs if needed, and feedback is of course greatly appreciated :)

### DIFF
--- a/src/fw/apps/system/timeline/timeline.c
+++ b/src/fw/apps/system/timeline/timeline.c
@@ -173,8 +173,8 @@ static void prv_swap_timeline(void *data) {
   const bool is_future = (s_app_data->timeline_model.direction == TimelineIterDirectionFuture);
   const bool to_timeline = true;
   const CompositorTransition *transition = PBL_IF_RECT_ELSE(
-      compositor_slide_transition_timeline_get(is_future, to_timeline, timeline_model_is_empty()),
-      compositor_dot_transition_timeline_get(is_future, to_timeline));
+      compositor_slide_transition_timeline_get(!is_future, to_timeline, timeline_model_is_empty()),
+      compositor_dot_transition_timeline_get(!is_future, to_timeline));
   static TimelineArgs args = (TimelineArgs) {
     .force_full = true
   };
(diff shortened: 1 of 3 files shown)


gmarull:
@jplexer ping

jplexer (approved):

gmarull:
sign-off doesn't match commit author?

soxfox42:
> sign-off doesn't match commit author?

All the commits seem to match for me. I used `git commit -s`, so they should all be using my configured name/email.
image

jplexer:
> sign-off doesn't match commit author?

yea they do?

gmarull:
> > sign-off doesn't match commit author?
> 
> 
> 
> yea they do?

committer is not the same as author, any reason?

soxfox42:
> committer is not the same as author, any reason?

Not sure if we're looking at different things here, but like I showed above, the committer and author are both me (with [Antigravity](https://antigravity.google/) as co-author as required by CONTRIBUTING.md).

soxfox42:
Oh, I've just spotted what you mean - GitHub has added another user to the commit who is definitely not present in the raw commit data. That seems like a GitHub bug.

Edit: Did some testing in a private repo, seems like this is due to the Antigravity coauthor line. I don't know why shimonenator seems to be connected with the antigravity@google.com email, but that specific email causes them to appear. As far as I can tell, Google Antigravity doesn't actually *have* a suitable address to use there, it just made one up. Is there another way I can document the AI usage so I can avoid this problem?

jplexer:
I guess you could use "Assisted-by: AGENT_NAME:MODEL_VERSION" in this case

PR #1461, opened 2026-06-09
doc: add developer doc for general development

ck-telecom:


gmarull (changes requested):
we have `docs/`, extend if needed

Issue #1265, opened 2026-05-09
Allow pairing of multiple devices even if only one active


luigi311:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

Allow pairing of multiple devices instead of only allowing a single device that needs to be forgotten before another can be paired.

In my case i have multiple phones that i switch between depending on what im doing/test/developing and the phones not in use get turned off/bluetooth disabled so the pebble itself only needs to handle a single connection but as of right now i have to go through the entire unpairing and then repairing process when i need to switch instead of just disconnecting/connecting to the phone i want that i have already paired once before.

testarossa47:
Would really appreciate this feature, too. I constantly switch between two phones and it would be great if the watch could remember multiple phones, while only being connected to one at a time

PR #1465, opened 2026-06-09
Feature/dual phone notifications

Coral-coder:
## Summary

- Add `multi_phone.h` with `PhoneSlot` type and `MAX_PHONE_CONNECTIONS=2`
- Refactor `ancs.c` to maintain a per-slot array of `ANCSClient` instances; GATT events are routed to the correct slot via characteristic lookup
- Update `kernel_le_client.c` with per-connection slot tracking and a service-discovery shim that passes slot context to ANCS; AMS and PPoGATT remain slot-0 only
- Add `bt_persistent_storage_get_all_ble_ancs_bondings()` and relax stale-bonding pruning to retain up to `MAX_PHONE_CONNECTIONS` entries

## Test plan

- [ ] Pair two iOS phones simultaneously and verify notifications arrive from both
- [ ] Disconnect one phone, verify the other continues delivering notifications
- [ ] Reconnect the disconnected phone, verify both resume
- [ ] Verify ANCS action (dismiss/reply) works with two phones connected
- [ ] Verify single-phone behavior is unchanged

### DIFF
--- a/include/pbl/services/bluetooth/bluetooth_persistent_storage.h
+++ b/include/pbl/services/bluetooth/bluetooth_persistent_storage.h
@@ -65,10 +65,12 @@ bool bt_persistent_storage_get_ble_pairing_by_addr(const BTDeviceInternal *devic
                                                    SMIdentityResolvingKey *IRK_out,
                                                    char name_out[BT_DEVICE_NAME_BUFFER_SIZE]);
 
-//! Returns the first ANCS supported bonding that is found
-//! The case of having multiple supported ANCS bondings isn't handled well yet.
-//! When this happens this could easily be changed to a for_each_ancs_supported_bonding(cb)
+//! Returns the first ANCS supported bonding that is found.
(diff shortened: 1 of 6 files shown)


Coral-coder:
Fixes issue #1265

gmarull (changes requested):
supporting multiple pairings requires some more work, there's a reason we do not allow them right now.

Coral-coder:
Yea… that’s why I’m submitting this…

jplexer:
it doesnt fix the issues that #1265 was describing

Coral-coder:
OkayyyyWhat blockers exist to prevent us from allowing multiple pairings? Am I missing something here?EOF#>On Jun 9, 2026, at 12:38, Joshua Jun ***@***.***> wrote:﻿jplexer left a comment (coredevices/PebbleOS#1465)
it doesnt fix the issues that #1265 was describing

PR #1477, opened 2026-06-10
fw/applib/graphics: fall back between base and extension fonts on glyph miss

phamson02:
## Problem

`prv_font_res_for_codepoint()` routes each codepoint to exactly one font: Latin-classified codepoints (`<= U+02AF` and the `U+2000–U+2BFF` symbols range) always go to the **base** font, while everything else goes to the language-pack **extension** font when one is loaded.  If this routed font (e.g. base font) does not contain the glyph, the renderer falls straight back to the wildcard box (▯), even when the other font (e.g. extension font) contains a valid glyph.

This means that **Latin-classified codepoints missing from the base fonts** (e.g. Vietnamese ơ/ư at U+01A1/U+01B0) can never be supplied by a language pack. This has been a known issue in the Vietnamese Pebble community since at least 2016 (issue: https://github.com/MarSoft/pebble-firmware-utils/issues/17) as language packs could not fix the missing ơ/ư.

I originally attempted to patch the base Gothic fonts to include these characters, but noticed that this fonts cover only about 360 of the roughly 3,650 codepoints this classification routes to them. With this fallback logic, any Latin-classified codepoint missing from the base font can instead be supplied by a community extension pack, which seems like a better long-term fix (Happy to discuss more on this if needed).

This also fixes the inverse issue: **glyphs that exist only in the base font but are not Latin-classified** (e.g. ﬁ U+FB01 and π U+03C0, both present in GOTHIC_18) stop rendering as soon as *any* language pack is installed, because they are then routed to an extension font that does not contain them.

## Fix

On a miss for the requested codepoint, try the other of base/extension before falling back to the wildcard.

Safety properties:

- The extra lookup only runs on the miss path, and only when an extension is loaded (`font_info->extended`) — app fonts and pack-less systems see zero change.
- The wildcard and space fallbacks still come exclusively from the base font, preserving the existing "extension pack may be deleted" guarantee.
- Glyph cache entries are keyed by resource id, so base and extension lookups (including cached negative results) cannot collide.

## Testing

- New regression test `test_text_resources__extension_miss_falls_back_to_base`: ﬁ (U+FB01) is present in GOTHIC_18 but absent from the Chinese language-pack fixture; with the extension installed, the lookup must return the base-font glyph. Verified the test **fails without the fix** (returns the 7 px wildcard instead of the 6 px ligature) and passes with it.
- Existing suite passes, including the "missing in both fonts → wildcard" case (`./pbl test -M ".*test_text_resources.*"`).
- Firmware builds clean for qemu_emery.

🤖 Generated with [Claude Code](https://claude.com/claude-code) and manually edited by me

### DIFF
--- a/src/fw/applib/graphics/text_resources.c
+++ b/src/fw/applib/graphics/text_resources.c
@@ -596,10 +596,36 @@ static const GlyphData *prv_get_glyph_in_font(FontCache *font_cache, Codepoint c
   return prv_get_glyph_metadata_from_spi(codepoint, font_cache, font_res, need_bitmap);
 }
 
+// Leaf glyph lookup against an explicit sub-font resource, bypassing the base/extension routing.
+// Used for the base<->extension sibling fallback.
+static const GlyphData *prv_get_glyph_in_font_res(FontCache *font_cache, Codepoint codepoint,
+                                                  const FontResource *font_res, bool need_bitmap) {
+  prv_check_font_cache(font_cache, font_res);
+  return prv_get_glyph_metadata_from_spi(codepoint, font_cache, font_res, need_bitmap);
+}
+
+//! The other of base/extension, used as fallback when the routed font lacks
+//! the glyph. NULL when no extension is loaded or when the routed font is not
+//! part of this FontInfo (e.g. the emoji font).
+static const FontResource *prv_alt_font_res(const FontResource *font_res,
+                                            const FontInfo *font_info) {
+  if (!font_info->extended) {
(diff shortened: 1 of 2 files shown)


ericmigi:
do you mind sharing some pictures of the watch rendering before and after your change? so we could visualize the change more easily (as non Vietnamese speakers 😉 )

SteffanDonal:
Commenting as a spectator, this seems like it'd allow us to add more emoji via an extension font

PR #1481, opened 2026-06-11
test: clear out the BROKEN_TESTS quarantine list

Mearman:
Most of the BROKEN_TESTS list in tests/wscript_build has been dead weight since the open-source import. Tests got quarantined to make `./waf test` pass and then never looked at again while the firmware moved on under them. This works through the list and re-enables 42 of the 50, verified in the CI Linux container (qemu_gabbro: 322 tests, 0 failures).

A few buckets.

**Stale includes and one stray byte.** Some just needed a renamed header leaf, a dead Bluetopia include dropped, or in stubs_common.h a UTF-8 BOM stripped out from in front of `#pragma once`. The BOM was desyncing clang into that misleading ButtonId/NUM_BUTTONS error cascade.

**The display migration.** The test platforms are emery (obelix, 200x228 rect) and getafix (gabbro, 260x260 round) now, and a lot of the snapshot tests still carried basalt-era goldens and rect-only assumptions. Regenerated the goldens from the current render, and where a test mixes round-only and rect-only symbols I guarded each test body with `#if PBL_ROUND` / `#if PBL_RECT`. The functions stay defined because clar scans the file as text. Dropped the obsolete `*_1bit` graphics entries too, the template loop only emits 8-bit now.

**API and enum drift.** wakeup_handle_clock_change split off a `significant` variant; PreciseBatteryChargeState grew a `pct` field; the platform enum grew Flint and Gabbro so the old `PlatformTypeEmery + 1` sentinel is now a valid value. Each test moved to the current contract.

**Bluetopia to NimBLE.** Biggest chunk. The BLE tests still pulled in the deleted SS1 fakes. Added fake_bt_driver_advert and fake_bt_driver_gatt that drive the real bt_driver_* / bt_driver_cb_* contracts: they record advertising intervals and payloads, build packed GATTService blobs and feed them through the real discovery callbacks, so the assertions still mean something rather than passing against hollow stubs. Covers the advertising, GATT-client, ANCS and ppogatt tests.

**Three production changes**, all deliberate. app_order_storage and battery_curve got `*_reset_for_tests()` hooks because both keep static state that leaked between tests with no reset path (inert in firmware). And tools/timezones.py had a real bug: the DST section wrote variable-length entries while timezone_database/service.c reads a fixed 16-byte stride, so the link section was misaligned and US/Pacific resolved to -1. Pad every DST id to 16 bytes.

**Still quarantined (8).** Two assert BLE features the firmware doesn't implement yet, so the honest fix is the feature not the test: gap_le_connect (watch-as-central, gated behind BLE_MASTER_CONNECT_SUPPORT which is defined nowhere, PBL-32761) and gatt_service_changed_server (the service-changed indication is an empty stub at both the service layer and the NimBLE driver, being done separately). Four are pixel-diff graphics tests: draw_circle_8bit, fill_circle_8bit, draw_line, draw_rotated_bitmap. A golden is the committed PNG compiled to a pbi, and for heavily anti-aliased output (arcs, radials, dotted lines, rotated bitmaps) that PNG to pbi round-trip isn't exact, so a golden made by saving the render back to PNG doesn't reproduce the render. The renders are right, regenerating the goldens needs the proper pbi tooling rather than a PNG copy. The last two are floor and pow: floor.c includes the host `<math.h>` after pblibc_private.h has renamed log/pow, which breaks glibc's vector-math macros (it builds on macOS, whose libc headers don't have them). They need the include order in the libc math sources sorting out first.

Also flips the test job to `waf test -k` so a single failing test doesn't hide every other failure behind it.

Everything was checked against the CI Linux container (pebbleos-docker:v5), not just locally, since the quarantine is about that environment.


gmarull (commented):
if AI assisted please, add Co-authored-by

include/pbl/services/battery/battery_curve.h
```diff
@@ -19,6 +19,10 @@ void battery_curve_set_compensation(BatteryCurveVoltageCompensationKey key, int
 
 void battery_curve_set_full_voltage(uint16_t voltage);
 
+//! Restore the discharge curve mutated by battery_curve_set_full_voltage().
+//! For test isolation only; inert in normal firmware operation.
+void battery_curve_reset_for_tests(void);
```
gmarull, line 25: ifdef to unittest?

include/pbl/services/process_management/app_order_storage.h
```diff
@@ -13,6 +13,9 @@ typedef struct PACKED AppMenuOrderStorage {
 
 void app_order_storage_init(void);
 
+//! Reset app_order_storage state for testing - clears cached "file missing" flag
+void app_order_storage_reset_for_tests(void);
```
gmarull: same, maybe a T_STATIC could do as well

tests/fakes/fake_bt_driver_gatt.c
```diff
@@ -0,0 +1,399 @@
+/* SPDX-FileCopyrightText: 2024 Google LLC */
```
gmarull: wrong copyright/s?

include/pbl/services/battery/battery_curve.h
Mearman: Done. Wrapped the declaration and definition in `#if UNITTEST` so it's not built into the firmware.

include/pbl/services/process_management/app_order_storage.h
Mearman: Went with the ifdef rather than `T_STATIC`: nothing in the firmware calls it, so `T_STATIC` would just leave an unused static (and the warning that comes with it).

tests/fakes/fake_bt_driver_gatt.c
Mearman: Oops that was definitely AI copypasta. Did the same on the other new fakes: `fake_bt_driver_advert.{c,h}` and `fake_event_gatt_service_buffer.h`.

gmarull (approved):

PR #1484, opened 2026-06-11
fw/drivers/imu/lis2dw12: harden against FIFO/INT1 stream stalls

jplexer:
## Summary

Investigation of FIRM-2490 ("Step count not working, started working after toggling health") showed the accel sample stream on obelix stalls: steps stop counting, activity sees frozen orientation, and stationary mode engages on a moving wrist. Recreating the accel session (the health toggle) fixed it because it fully re-arms the FIFO and INT1 — this series makes the driver do that itself.

Root cause mechanics:
- INT1 is edge-triggered while the sensor's FIFO-threshold output is level-based: if the drain work is delayed past FIFO overrun (~280 ms slack at 25 Hz), the line latches high and no further edges arrive.
- The existing INT1 watchdog timed the last INT1 *edge*, but shake/wake-up function interrupts share the INT1 pad (CTRL7 INT2_ON_INT1), so wrist motion kept feeding the watchdog while the FIFO stream was dead — fully silent stalls.
- Its recovery only rewrote FIFO_CTRL (field logs show it firing 3x in 52 s without keeping the stream alive) and discarded queued samples.
- accel_peek returns the cached last FIFO sample during active sampling, so a dead stream freezes stationary/orientation data and the stall masks itself.

## Changes

1. Watchdog tracks successful FIFO burst reads instead of INT1 edges, with a 2x period margin (1x tripped on normal jitter: 1004 ms vs 1000 ms period).
2. Shared recovery helper (watchdog + overrun path): quiesce INT routing (forces a latched-high pad to produce a fresh edge), drain queued samples, re-assert ODR/FIFO/INT routing, clear latched sources, log a recovery counter.
3. accel_peek detects stale cached data and queues the stall check itself, breaking the self-masking loop.
4. Drain the FIFO on subscriber reconfiguration instead of discarding up to a full FIFO of samples (resolves existing FIXME).
5. Floor the stall threshold at 1 s so small-batch subscribers (e.g. per-sample at 25 Hz) don't trip spurious recoveries.

Fixes FIRM-2490
Related: FIRM-2285, FIRM-1626, FIRM-1141

## Testing

- obelix_pvt and getafix_dvt (both lis2dw12 boards) build clean
- `./waf test` green (200 suites, including test_accel_manager)
- On-watch validation pending (QEMU does not emulate the lis2dw12): watch for `FIFO stream stalled for N ms` / `Recovering accel stream (count N)` followed by steps continuing to count, and no recovery storms

### DIFF
--- a/src/fw/drivers/imu/lis2dw12/lis2dw12.c
+++ b/src/fw/drivers/imu/lis2dw12/lis2dw12.c
@@ -29,8 +29,10 @@
 // - For some reason (needs more investigation), INT1 is sometimes left HIGH
 //   due to FIFO overruns, and without edge change, we cannot detect such
 //   events. To mitigate this, there is a watchdog timer that re-arms FIFO
-//   if no INT1 event is detected within the expected time window based on the
-//   ODR and FIFO threshold.
+//   if no FIFO samples have been read within the expected time window based
+//   on the ODR and FIFO threshold. Sample reads (not INT1 edges) are tracked
+//   because shake/wake-up events share the INT1 pad and would otherwise mask
+//   a dead FIFO stream.
 
 // Time to wait after reset (us)
 #define LIS2DW12_RESET_TIME_US 5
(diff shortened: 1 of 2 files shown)


gmarull (commented):
let me take this carefully, first question: has this been tested on hw?

jplexer:
regression-tested normal steps/shake on getafix_dvt2, and verified the recovery path with a fault-injection build (killed INT routing / FIFO mode / ODR behind the driver's back) all recovering sucessfully

gmarull:
See #1701

PR #1496, opened 2026-06-12
build: container runners for host tests and firmware builds

Mearman:
Host tests compile against the host libc, which is different on macOS and the Linux CI container. Tests pass on one and fail on the other because of linker behaviour (-fno-common), libc headers (glibc's vector-math macros), compiler versions, and so on. The CI Docker image (`ghcr.io/coredevices/pebbleos-docker:v5`) is the canonical environment, but until now you had to manually mount the source and run ad-hoc Docker commands to reproduce it.

`ci-test.sh` reproduces the test.yml steps inside the Docker image: safe.directory, pip install, waf configure, waf test. Passes through waf arguments so you can scope runs.

`build_in_container.sh` does the same for the firmware build workflows (build-firmware, build-prf, build-qemu). Validates the board name, appends the right waf targets for QEMU boards, accepts CONFIGURE_EXTRA flags.

`.devcontainer/devcontainer.json` opens VS Code or Codespaces into the same Docker image so the toolchain matches CI exactly.

CONTRIBUTING.md updated to document the workflow: container runners before pushing, native `./waf` for iteration.

Merge order: **1** (independent)

### DIFF
--- a/docs/development/testing-locally.md
+++ b/docs/development/testing-locally.md
@@ -1,0 +1,16 @@
+# Testing locally
+
+The CI test environment runs inside a Docker container
+(`ghcr.io/coredevices/pebbleos-docker:v5`) to ensure consistent libc versions
+and tooling. Host unit tests compile against the host libc, which differs
+between macOS and the Linux CI environment, so running tests locally without the
+container may produce different results.
+
+To reproduce CI locally, use the container runners provided in `tools/ci/`:
+
+- `./tools/ci/run_tests.sh` — run the host test suite (same as CI)
+- `./tools/ci/build_firmware.sh` — build firmware in the CI container
+
+The native `./waf test` and `./waf build` commands are faster for iteration but
+can diverge from CI. Always run the container scripts before pushing to verify
+consistency.
--- a/tools/ci/build_firmware.sh
+++ b/tools/ci/build_firmware.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 Core Devices LLC
+# SPDX-License-Identifier: Apache-2.0
+# build_firmware.sh — Firmware build runner for PebbleOS in Docker
+#
+# Reproduces build-firmware.yml, build-prf.yml, and build-qemu.yml steps
+# inside ghcr.io/coredevices/pebbleos-docker:v5
+#
+# Usage: ./tools/ci/build_firmware.sh --board <board> [options]
+#
+# Environment variables (all optional):
+#   PEBBLEOS_DOCKER_IMAGE — Docker image (default: ghcr.io/coredevices/pebbleos-docker:v5)
+#   BOARD                  — Board to build (required unless set via --board flag)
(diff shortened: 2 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -76,6 +76,17 @@ Additional requirements:
...
+## Testing Locally
+
+The CI test environment runs inside a Docker container (`ghcr.io/coredevices/pebbleos-docker:v5`) to ensure consistent libc versions and tooling. Host unit tests compile against the host libc, which differs between macOS and the Linux CI environment, so running tests locally without the container may produce different results.
+
+To reproduce CI locally, use the container runners provided in `scripts/`:
+
+- `./scripts/ci-test.sh` — run the host test suite (same as CI)
+- `./scripts/build_in_container.sh` — build firmware in the CI container
+
+The native `./waf test` and `./waf build` commands are faster for iteration but can diverge from CI. Always run the container scripts before pushing to verify consistency.
```
gmarull: likely belongs to `docs/` (as probably other stuff here maybe, it's getting too long)

scripts/build_in_container.sh
```diff
@@ -1,3 +1,4 @@
+/* SPDX-License-Identifier: Apache-2.0 */
```
gmarull: amend original commits

Mearman:
thanks for the feedback @gmarull.
will action those now

tools/ci/build_firmware.sh
```diff
@@ -0,0 +1,106 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 Core Devices LLC
```
gmarull, line 2: not by us, also, use e.g. `tools/ci`, so no new toplevel folder

scripts/build_in_container.sh
```diff
@@ -0,0 +1,106 @@
...
+Valid boards:
+  - asterix
+  - obelix_dvt
+  - obelix_pvt
+  - getafix_evt
+  - getafix_dvt
+  - getafix_dvt2
+  - qemu_flint
+  - qemu_emery
+  - qemu_gabbro
```
gmarull, line 67: candidate to stall, no hardcoding of names pls

docs/development/testing-locally.md
```diff
@@ -6,10 +6,10 @@ and tooling. Host unit tests compile against the host libc, which differs
 between macOS and the Linux CI environment, so running tests locally without the
 container may produce different results.
 
-To reproduce CI locally, use the container runners provided in `scripts/`:
+To reproduce CI locally, use the container runners provided in `tools/ci/`:
 
-- `./scripts/ci-test.sh` — run the host test suite (same as CI)
-- `./scripts/build_in_container.sh` — build firmware in the CI container
+- `./tools/ci/run_tests.sh` — run the host test suite (same as CI)
```
gmarull, line 11: in general: amend, main tree should not contain intermediate work

tools/ci/run_tests.sh
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: 2026 Core Devices LLC
```
gmarull, line 2: ping: not us!

PR #1503, opened 2026-06-12
fw: run glyph flash loads on a shared scratch stack

jplexer:
## Problem

v4.11.0 introduced a 100%-reproducible stack overflow when rendering notifications with non-Latin text (FIRM-2475; user reports FIRM-2447/2491/2498). Two changes collided:

- `56eeca9a` (FIRM-2160 fix) moved the glyph bitmap pre-load into `walk_line()`'s render pass, so the first render of an uncached glyph pays the deep flash chain (`resource_load → pfs_read → flash_read_bytes → task_timer_start`) on the render stack.
- `7cf5af22` added the Arabic/RTL shaping path with two 128-byte stack buffers in the same `walk_line` frame.

Non-Latin glyphs are perpetual cache misses (huge codepoint space, PFS-backed language packs), so every line render pays multiple deep loads on the 2 KB App/KernelMain stacks. This is the third relocation of the same load (render_glyph → measure pass → walk_line render pass); every call site that needs glyphs is already at the bottom of a deep UI stack, so there is no shallow place left to move it to. The dedicated syscall stack doesn't help here: the Notifications app is a privileged system app and KernelMain is privileged, so their syscalls never SVC-trap.

## Fix

Give the load its own stack instead of relocating it again:

- **`fw/kernel/scratch_stack.c`**: a mutex-serialized 2 KiB borrowed stack in KERNEL_RAM. Privileged callers switch PSP/PSPLIM onto it for the duration of one callback (naked-asm swap); blocking while borrowed is safe because the CM33 port saves/restores both PSP and PSPLIM per task context. Unprivileged callers fall through to a direct call (their deep syscalls already relocate to the per-task syscall stack at the SVC boundary), as do ISRs, reentrant calls, and non-ARMv8-M builds.
- Two invariants keep FreeRTOS's switch-out check (`pxTopOfStack <= pxStack` → overflow hook) happy while borrowed: the scratch stack is **linker-placed above the kernel task stacks**, and borrowing is **whitelisted** to App/Worker (whose method-1 trip `vApplicationStackOverflowHook` already ignores) and KernelMain/KernelBG (whose stacks sit below the scratch stack). Other tasks may have heap-allocated stacks above it and fall back to a direct call.
- **`text_resources.c`**: the three flash-touching glyph paths (cache-miss load, cached-entry bitmap load, font reload) run via `scratch_stack_call()`. Cache hits don't touch the mutex, so the hot render path is unchanged. The miss path covers metadata-only loads too, so the FIRM-2160 measure-pass case is also off-stack now.
- **Analytics**: `stack_free_scratch_bytes` reports the high-water for sizing, mirroring the syscall-stack metrics.

## Validation

- obelix (ARMv8-M, feature live) and asterix (ARMv7-M, direct-call fallback) build clean; scratch stack verified above `__kernel_main_stack_start__`/`__kernel_bg_stack_start__` in the ELF.
- Unit tests: unchanged vs main (one pre-existing unrelated timezone failure); host builds use the inline fallback.
- QEMU (qemu_emery, CM33): booted, exercised launcher/notifications/settings text rendering, no faults.
- **Hardware**: validated on PT2 with a non-Latin language pack — the FIRM-2491 repro (open notifications, scroll) no longer crashes. This exercises the FP extended-frame path QEMU can't.

Fixes FIRM-2475


gmarull (commented):
this is becoming very complex, making me skeptical about the overall solution (maybe it's ok, but LLMs can get out of control easily). Can the root problem be described in a few words?

jplexer:
Loading a glyph means reading the font from flash mid-render, and that read chain needs more stack than any 2 KB task stack has left at that depth. We've already moved the load twice (#975, #1422) and each move just overflowed a different stack, so the fix gives the load its own stack instead of relocating it a third time.

gmarull:
> Loading a glyph means reading the font from flash mid-render, and that read chain needs more stack than any 2 KB task stack has left at that depth. We've already moved the load twice (#975, #1422) and each move just overflowed a different stack, so the fix gives the load its own stack instead of relocating it a third time.

isn't it an option to increase app/main stacks (where I guess we run through that deep call stack)?

jplexer:
> > Loading a glyph means reading the font from flash mid-render, and that read chain needs more stack than any 2 KB task stack has left at that depth. We've already moved the load twice (#975, #1422) and each move just overflowed a different stack, so the fix gives the load its own stack instead of relocating it a third time.
> 
> isn't it an option to increase app/main stacks (where I guess we run through that deep call stack)?

we could and i"d be fine with that

gmarull:
> > > Loading a glyph means reading the font from flash mid-render, and that read chain needs more stack than any 2 KB task stack has left at that depth. We've already moved the load twice (#975, #1422) and each move just overflowed a different stack, so the fix gives the load its own stack instead of relocating it a third time.
> > 
> > 
> > isn't it an option to increase app/main stacks (where I guess we run through that deep call stack)?
> 
> we could and i"d be fine with that

If we have enough RAM to eg double, let's take trivial approach and optimize later if needed

jplexer:
#1508

Issue #1510, opened 2026-06-12
RFC: Gyroscope API (gyro_data_service) — a pickleball pet project that might be useful to others


dbortz:
## Problem

This started as a personal pet project: I wanted to track wrist movement
during pickleball games with my Pebble Time 2, and accelerometer-only motion
classification wasn't cutting it — swing/shot recognition really wants angular
velocity. The watches that do this well on other platforms all lean on a gyro.

It turns out the hardware is already there: both Pebble Time 2 (obelix) and
Pebble 2 Duo (asterix) carry an ST LSM6DSO(W) 6-axis IMU. On obelix the chip
is currently entirely unused (the LIS2DW12 handles all motion features and the
LSM6DSO driver isn't compiled in); on asterix the driver runs accel-only with
the gyro explicitly powered down. There's even a TODO in
`src/fw/drivers/imu/lsm6dso/lsm6dso.c` anticipating gyro support (the ULP
power-mode note). But there's no way for an app to reach it — no gyro API in
the SDK.

I went ahead and built it for myself on a fork, and it works — so I wanted to
surface it here in case it's interesting to the community, since a published
app can't depend on a private-fork API.

## Proposed change

A gyroscope data service mirroring the accelerometer stack layer by layer,
working today on my PT2 (branch: `dbortz/PebbleOS@gyro-service`, PR to follow):

- **Driver**: new `drivers/gyro.h` interface modeled on `drivers/accel.h`,
  implemented for the LSM6DSO. The accel and gyro halves can be enabled
  independently (`CONFIG_ACCEL_LSM6DSO` / `CONFIG_GYRO_LSM6DSO`); shared chip
  ownership (init, I2C, the single FIFO with tag-dispatched records) lives in
  a small `lsm6dso_core.c`. Sample collection is FIFO-batched and
  timer-polled, since obelix doesn't route an LSM6DSO interrupt pin. This also
  resolves the ULP-mode TODO (accel leaves ULP while the gyro runs, datasheet
  §6.2.1).
- **Service**: `gyro_manager`, a structural mirror of `accel_manager`
  (shared circular buffer, per-client subsampling, batched delivery,
  syscalls). One deliberate difference: the gyro is **fully powered down
  whenever there are no subscribers** (~3 µA chip power-down) — no ambient OS
  use, so zero battery cost unless an app asks for data.
- **applib/SDK**: `gyro_data_service_subscribe()` /
[...]

PR #1511, opened 2026-06-12
drivers/imu: add LSM6DSO gyroscope driver

dbortz:
Splitting the original combined PR into smaller, layered PRs as requested (#1510 has the full context/RFC). This is the **first** of three:

1. **this PR** — the LSM6DSO gyroscope driver
2. *(next)* `gyro_manager` service + board enablement
3. *(after that)* applib `gyro_service` + SDK export (`PBL_GYRO`)

I'll open each subsequent PR once the previous lands.

## What this PR does

- Refactors the LSM6DSO chip management (init, I2C, the shared hardware FIFO) out of the accelerometer driver into a small `lsm6dso_core`, so the accel and gyro halves can be compiled independently (`CONFIG_ACCEL_LSM6DSO` / `CONFIG_GYRO_LSM6DSO`). FIFO records from the two sensors are interleaved and dispatched by tag.
- Adds the `drivers/gyro.h` interface and its LSM6DSO implementation (`lsm6dso_gyro.c`). The gyro is FIFO-batched and timer-polled (not all boards route an LSM6DSO interrupt pin) and stays powered down until samples are requested.
- Resolves the existing ULP-mode TODO in `lsm6dso.c`: the accelerometer now leaves ultra-low-power mode while the gyro is active (datasheet §6.2.1).

## On building / coverage

The driver isn't enabled on any board here — the service that consumes it (and the board enablement that links the two) is the next PR, so they can be reviewed separately as you asked. Concretely:

- The **`lsm6dso_core` refactor is exercised in CI** via the existing accelerometer on `asterix` (no behavior change).
- **`lsm6dso_gyro.c` and the gyro-specific `#if CONFIG_GYRO_LSM6DSO` paths compile once the service PR enables a board** — this is the usual driver-then-enable split. I've verified the full 3-PR series builds clean on `obelix` and `asterix`.

If you'd rather have the driver and service land together (so every line is exercised in this PR's own CI), I'm happy to combine those two — just say the word.

### DIFF
--- a/src/fw/drivers/gyro.h
+++ b/src/fw/drivers/gyro.h
@@ -1,0 +1,96 @@
+/* SPDX-FileCopyrightText: 2026 Dave Bortz */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#pragma once
+
+#include <inttypes.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+/*! Gyroscope driver interface
+ *  ==========================
+ *
+ * The gyroscope driver mirrors the accelerometer driver interface (see
+ * drivers/accel.h for the design rationale): it is a dumb interface to the
+ * gyroscope hardware with no circular buffers, clients, threads or
+ * subsampling. All of that is handled by higher level code (the gyro
+ * manager service).
+ *
+ * Power: the driver must keep the gyroscope powered down whenever no data
+ * has been requested (sampling interval and num samples both zero). Unlike
+ * the accelerometer, the gyroscope is comparatively power hungry, so there
+ * are no always-on event detection features (shake, tap) in this interface.
+ *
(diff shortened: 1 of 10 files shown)


gmarull (changes requested):
split changes into smaller PRs, eg first driver, then service, etc.

dbortz:
Thanks for the review! Split as suggested — this PR is now just the driver (the `lsm6dso_core` refactor + the gyro driver). The `gyro_manager` service + board enablement will be the next PR, and the applib/SDK export (`PBL_GYRO`) the one after that. I'll open them in sequence as each lands. Branches are pushed (`gyro-pr2-service`, `gyro-pr3-sdk`) and the full series builds on obelix and asterix.

PR #1513, opened 2026-06-12
bluetooth: support concurrent BLE clients

lesserevil:
## Summary
- Allow the gateway phone link and one external BLE client to coexist.
- Track multiple local-slave BLE connections instead of a single boolean.
- Classify NimBLE pairings as gateway only after Pebble pairing service use.
- Prune stale gateway bonds without deleting non-gateway BLE client bonds.

## Scope
This is the BLE infrastructure branch needed by workout HRM sharing. It intentionally does not add the HRM GATT service, HRM advertising primitive, or workout publishing logic.

## Tests
- `git diff --check`
- `./pbl test -M 'test_bluetooth_persistent_storage|test_nimble_hrm_service'`
- `./pbl build`
- `gitlint --commits HEAD~1..HEAD`

Note: `test_gap_le_connect` remains in the repository `BROKEN_TESTS` list, so the new concurrent-slave-connection unit test is not run by the default test target until that suite is re-enabled.


lesserevil:
Sorry for the spam. This PR and the other three in the set have been running on my watch just fine, and the new feature works. 

Please let me know if you need more info/testing/etc.

ericmigi:
Pretty cool. Can you share how you tested this and which other BLE devices you tested with?

lesserevil:
I have tested it on my PT2, and connected to a Garmin 1040 computer and used a BLE sniffer on my phone to check that the packets start/stop when the workout starts/stops on the watch.

lesserevil:
I've updated the PRs against the latest 'main', and I'm still able to run this on my own device. 

Please let me know if there something you want me to change or if you need more information. Thanks!

ericmigi:
It's tough for us to schedule time to merge this, since it would take a little bit of effort for one of our engineers to get up and running on this. But we see you, and will get to this eventually. sorry for the delay!

sruggier:
I'm not the original author of the PR, but it looks like it's probably fair to remove the stale label and reopen it.

PR #1519, opened 2026-06-13
ble: simultaneous dual-phone connections with full ANCS and PPoGATT support

Coral-coder:
## Summary

PebbleOS hard-codes a single phone connection throughout the BLE stack. This series removes that assumption end-to-end, supporting any two phones simultaneously regardless of OS.

## What changes

**`multi_phone.h`** introduces `PhoneSlot` (`uint8_t`), `MAX_PHONE_CONNECTIONS 2`, and `PHONE_SLOT_INVALID 0xFF`.

**NimBLE** raises `BLE_MAX_CONNECTIONS` from 1 to 2 on both nrf52 and sf32lb52 targets.

**`gap_le_advert` / `gap_le_connect`** replaces single-connection booleans with `uint8_t` slave connection counters so advertising and reconnect logic correctly tracks whether all slots are full.

**`ancs.c`** replaces the `s_ancs_client` singleton with `s_ancs_clients[MAX_PHONE_CONNECTIONS]`. All entry points take a `PhoneSlot slot` parameter. GATT characteristic events route to the correct slot via `prv_slot_for_characteristic()`.

**ANCS action routing** adds a uid-to-slot map so watch-initiated actions (dismiss, reply, answer, decline) go to the correct phone. Without this, uid=1 on phone A and uid=1 on phone B would collide and actions could fire on the wrong device.

**`ppogatt.c`** extends the same treatment to Android: `PPoGATTClient` gains a `slot` field, rediscovery state becomes per-slot, and create/destroy/service-discovery all take `PhoneSlot`. Both protocols now operate independently on separate slots.

**`kernel_le_client`** replaces the single-bonding init lookup with `bt_persistent_storage_get_all_ble_ancs_bondings`; slot wrappers route service discovery events to the correct protocol handler per connection.

**Bonding storage** retains up to `MAX_PHONE_CONNECTIONS` BLE pairings and prunes only when a third is added.

**PRF stubs and unit tests** updated throughout, including dual-phone bonding storage and GAP advertising expectations.

## Two gateways problem

Prior firmware treated BLE bonding as single-phone: storing a new gateway pairing replaced the previous one, and advertising stopped as soon as one phone connected. That made a second phone impossible even when the link layer could support two connections.

This series fixes that at three levels:

- **Storage:** keep up to two distinct gateway bondings instead of replacing on every new pair
- **Advertising:** resume advertising when one slot is free so a second phone can discover the watch
- **Protocol clients:** per-slot ANCS and PPoGATT instances so each gateway gets its own service state

An earlier draft of this work did not fully address the gateway side of dual-phone support. This PR is a clean resubmit with that fixed end to end.

## What this enables

- iPhone + iPhone: independent ANCS streams, one per slot
- Android + Android: independent PPoGATT sessions, one per slot
- iPhone + Android: ANCS on one slot, PPoGATT on the other

## Known limitations

**AMS** (music controls) remains slot-0 only.

**Notification storage** still keys items by `ancs_uid` without a phone slot. Watch-initiated actions are routed correctly via the uid-to-slot map in `ancs.c`, but phone-initiated removals and storage-level lookups can still collide if two phones reuse the same UID. A full fix would need slot-aware fields in `TimelineItemHeader` and `notification_storage`.

Note: this PR overlaps with #1513 on `gap_le_connect.c` (`bool` to `uint8_t` counter). Happy to rebase on top of that once it lands.

## Test plan

- [ ] CI green: build-firmware, build-prf, build-qemu (emery/flint/gabbro), nix, tests
- [ ] Pair two iPhones, verify independent ANCS notification streams
- [ ] Pair two Android phones, verify independent PPoGATT sessions
- [ ] Pair iPhone + Android, verify ANCS on one slot and PPoGATT on the other
- [ ] Dismiss/reply on phone A, confirm it does not affect phone B
- [ ] Verify advertising resumes with one slot occupied, stops with both full

### DIFF
--- a/include/pbl/services/bluetooth/bluetooth_persistent_storage.h
+++ b/include/pbl/services/bluetooth/bluetooth_persistent_storage.h
@@ -65,10 +65,12 @@ bool bt_persistent_storage_get_ble_pairing_by_addr(const BTDeviceInternal *devic
                                                    SMIdentityResolvingKey *IRK_out,
                                                    char name_out[BT_DEVICE_NAME_BUFFER_SIZE]);
 
-//! Returns the first ANCS supported bonding that is found
-//! The case of having multiple supported ANCS bondings isn't handled well yet.
-//! When this happens this could easily be changed to a for_each_ancs_supported_bonding(cb)
+//! Returns the first ANCS supported bonding that is found.
 BTBondingID bt_persistent_storage_get_ble_ancs_bonding(void);
+
+//! Collects all BLE ANCS bondings into @p out, up to @p max_count entries.
+//! @return number of bondings written into @p out.
+int bt_persistent_storage_get_all_ble_ancs_bondings(BTBondingID *out, int max_count);
 
 //! Returns true if the bondings is BLE and supports ANCS
 bool bt_persistent_storage_is_ble_ancs_bonding(BTBondingID bonding);
--- a/resources/normal/base/lang/tintin.pot
+++ b/resources/normal/base/lang/tintin.pot
@@ -692,6 +692,14 @@ msgstr ""
 msgid "Open the Pebble app on your phone to connect."
 msgstr ""
 
+#: ../src/fw/apps/system/settings/bluetooth.c
+msgid "Open the Pebble app on another phone to connect."
+msgstr ""
+
+#: ../src/fw/apps/system/settings/bluetooth.c
+msgid "Forget a device to pair a new one."
+msgstr ""
+
 #: ../src/fw/apps/system/settings/bluetooth.c:452
 msgid "Forget this device to pair a new device."
 msgstr ""
--- a/src/fw/apps/system/settings/bluetooth.c
+++ b/src/fw/apps/system/settings/bluetooth.c
@@ -16,6 +16,7 @@
 #include "comm/bt_lock.h"
 #include "comm/ble/gap_le_connection.h"
 #include "comm/ble/gap_le_device_name.h"
+#include "comm/ble/kernel_le_client/multi_phone.h"
 #include "drivers/rtc.h"
 #include "kernel/pbl_malloc.h"
 #include "kernel/ui/system_icons.h"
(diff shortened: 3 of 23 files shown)


gmarull (changes requested):
Supporting multiple connections has wide implications: user experience, power consumption etc. For now, it's not a priority on our side.

Coral-coder:
Thanks for telling me you don’t care without telling me you don’t care
It’s a major priority for me on my side and I’m moving forward with it.

I want notifications from both my phones at the same time. That’s the whole reason I bought this watch.

gmarull:
> Thanks for telling me you don’t care without telling me you don’t care It’s a major priority for me on my side and I’m moving forward with it.
> 
> I want notifications from both my phones at the same time. That’s the whole reason I bought this watch.

This feature was never advertised, and one day it may be available, just not right now due to the implications I mention. However, you can maintain and sideload your own firmwares.

Coral-coder:
Oh, come on! this feature existed on the previous Pebble generation

Would you be happier if I just created a toggle in the Debug or developer menu to allow multiple connections?

Because it seems like the problems that you indicate are only for users that actually want or have two phones attached to their watch.

jplexer:
> Oh, come on! this feature existed on the previous Pebble generation

Not officially, no. It was a side effect of having both BLE and BT Classic. We only have BLE nowadays

Coral-coder:
Well, can I stop being stonewalled and actually bring the official feature? 

Getting a little frustrated that I'm basically being told "no we don't want this"

That was a serious driving factor in me buying a new Pebble
That's what made it better...

jplexer:
Also sidenote if you used AI for this PR you must note it in the commits done by AI (as per the Contribution Guidelines)

gmarull:
> Well, can I stop being stonewalled and actually bring the official feature?
> 
> Getting a little frustrated that I'm basically being told "no we don't want this"
> 
> That was a serious driving factor in me buying a new Pebble That's what made it better...

We're not gonna merge this feature anytime soon. I already mentioned a way forward (maintain/sideload your own firmwares).

PR #1528, opened 2026-06-15
build: adopt board-declared, west-style runners

gmarull:
Move device-runner selection into the board manifests and restructure the runners as a self-contained package modeled on Zephyr's west binary runners.

- Boards declare their supported runners in <board>.yml; the first is the default. Drop the hardcoded RUNNERS map and the configure-time --runner.
- New tools/runners package with RunnerCaps/RunnerConfig and a capabilities()/do_add_parser()/create()/run()->do_run() dispatch. Runners shell out directly, so waftools/{openocd,sftool,nrfutil}.py are removed.
- Runners own their CLI args (e.g. --tty for sftool) via do_add_parser(); --resources stays a native flash option and the runner fails if it cannot program resources.
- OpenOCD configuration lives in the board: boards/asterix/support/openocd.cfg (cmsis-dap baked in). Drop the generated openocd.cfg, the template, the JTAG fragments and --openocd-jtag.
- Add native ./waf flash/run/debug commands that dispatch through the runner, alongside the existing ./pbl commands; both share the runner package.


jplexer:
what do you mean I can review prs from linear now???

gmarull:
> what do you mean I can review prs from linear now???

omg no pls

jplexer (approved):

PR #1534, opened 2026-06-16
applib: expose alarm service to apps

leifenberg-eyal:
Add a read-only AlarmService applib API so apps (e.g. watchfaces) can show an alarm indicator. alarm_service_peek_next() reports whether the wearer has an enabled alarm scheduled and, if so, the time it fires.

Alarm state lives in the kernel, so this adds a
sys_alarm_get_next_enabled() syscall wrapping the existing alarm_get_next_enabled_alarm(), the applib wrapper, and the SDK export (rev 104, SDK minor 0x65). The API is read-only: apps cannot create, modify or delete alarms. The system supports up to 10 alarms.

Note: usable by apps only once shipped in a firmware + SDK release.

Configured cleanly for `obelix_pvt`.

🤖 Generated with assistance from Claude (Claude Code); see the `Co-Authored-By`
trailer on the commit.

### DIFF
--- a/src/fw/applib/alarm_service.c
+++ b/src/fw/applib/alarm_service.c
@@ -1,0 +1,10 @@
+/* SPDX-FileCopyrightText: 2026 Core Devices LLC */
+/* SPDX-License-Identifier: Apache-2.0 */
+
+#include "alarm_service.h"
+
+#include "syscall/syscall.h"
+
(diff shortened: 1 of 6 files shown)


gmarull (approved):
lgtm, @jplexer has final word here

jplexer (approved):

PR #1540, opened 2026-06-16
workout: fix HH:MM:SS duration clipping on narrow displays

ericmigi:
Once a workout passes one hour, the duration field switches from `MM:SS` to `HH:MM:SS` (8 chars). On rectangular displays ≥200px tall (obelix/emery) the reduced number font (LECO_42) is still too wide for 8 chars in the space left of the action bar, so the seconds got clipped (rendered e.g. as `04:20:-`).

Fix: use the narrower LECO_36 font for hour-long durations on those displays. Round displays (gabbro) have room to spare and smaller displays already use a font that fits, so both keep their existing font.

Verified by the `test_workout_active` render tests; the obelix fixture for `sports_pace_long_values` is updated to show the full `04:20:39`.

| Before | After |
| --- | --- |
| `04:20:-` (clipped) | `04:20:39` |


jplexer (approved):

gmarull (approved):
pls push branches to your fork on next PR

PR #1541, opened 2026-06-16
clock: add hourly speaker chime

Coral-coder:
Add a speaker chime tone that plays at the top of each hour. The setting appears in Sounds & Haptics as 'Hourly Chime' directly below the existing 'Hourly Notice' vibe row. Selecting it toggles the feature on or off.

When enabled, the watch plays the Chime tone (triangle-wave ascending C5-E5-G5-C6) via speaker_service at the user's configured volume, independent of the vibration setting. The chime respects the system mute flag (both always-on and DND-scoped mute).

Persisted under the key 'hourlyChimeSound' in the notifpref settings file; defaults to off.


Coral-coder:
Vibed with Cursor. waiting for it to actually chime before I claim ready for review.

PR #1545, opened 2026-06-16
Feature: Dual Phone Support / Full stack Implementation - Complete

Coral-coder:
## Dual-phone Bluetooth concurrency

This adds full-stack support for two phones connected to the watch over BLE simultaneously. Both phones maintain live Pebble Protocol sessions; the watch decides which one is the "gateway" (the source of notifications and system messages) while the other remains a secondary peer.

### What "full stack" means here

**BLE advertising (GAP layer)**
After the first phone connects, the BLE hardware stops advertising by default. We restart advertising from the kernel background task so the second phone can still see and connect. The watch stays discoverable until both slots are occupied.

**Connection management (`kernel_le_client.c`)**
The watch tracks up to two phone slots. Each slot gets its own `BTDeviceInternal` identity. A pinned-gateway bonding ID (user preference, persisted to flash) means the preferred phone always wins the gateway slot when it reconnects -- no user action required.

**GATT discovery and PPoGATT (`ppogatt.c`)**
PPoGATT is the BLE transport that carries Pebble Protocol. We made it per-slot: each phone gets its own independent GATT client state, sequence-number window, and reset/ACK machinery. UUID deduplication (which previously blocked a second session from the same app UUID) is now scoped per slot. If `comm_session_open` fails after a reset handshake, the client schedules a bounded GATT rediscovery retry rather than silently dropping the connection.

**Comm session routing (`session.c`)**
The session layer now allows two PPoGATT sessions to coexist. The system session -- used for notifications, firmware updates, app install, etc. -- is routed exclusively to the gateway slot. The secondary phone's session is kept alive but receives only app-targeted traffic. Put-bytes teardown is gated on the gateway session closing, preventing premature cleanup when the secondary phone is still active.

**ANCS (`ancs.c`)**
Apple Notification Center Service subscriptions are maintained per-slot, so both an iPhone and an Android phone can deliver notifications concurrently. Action responses (dismissing a notification, replying) are routed back to whichever phone originated the notification.

**AMS (Apple Media Service)**
AMS (music controls) follows the gateway: when the gateway changes, `ams_destroy` / `ams_create` transfers the media subscription to the new gateway phone.

**Gateway selection**
When the active gateway disconnects while a secondary phone is still connected, the secondary is immediately and transparently promoted to gateway -- no delay, no popup. The pinned preference is preserved in storage, so the moment the preferred phone reconnects it reclaims the gateway slot and AMS automatically.

**Settings and pairing (`bluetooth.c`, `pairability.c`)**
The Bluetooth settings screen marks the active gateway with a star. The watch stays pairable until both connection slots are full, allowing a second phone to pair without requiring the first to disconnect first.

**Persistent storage**
The active gateway bonding ID is persisted and loaded at boot. Bonding ID lists are cached at init to avoid storage calls while the BT lock is held.

### Files changed

| Layer | Files |
|---|---|
| GAP / advertising | `gap_le_advert.c`, `gap_le_slave_reconnect.c` |
| Connection + gateway | `kernel_le_client.c`, `multi_phone.h` |
| PPoGATT transport | `ppogatt.c`, `ppogatt.h` |
| Comm session | `session.c`, `put_bytes.c` |
| ANCS / AMS | `ancs.c`, `ams.c` |
| Pairability / settings | `pairability.c`, `bluetooth.c` |
| Persistent storage | `bluetooth_persistent_storage.c` |
| Unit tests | `test_ppogatt.c`, `test_kernel_le_client.c`, `test_ancs.c`, `test_put_bytes.c` |

No changes to the Pebble Protocol wire format or the SDK API surface. Existing single-phone behaviour is unchanged.

Vibed with Claude, Coral, And Cursor. (Three heads are better than one)

Cheers!


Coral-coder:
Feel free to put a pin in this one for as long as you need... 
No rush of course, But in the event you ever want to implement this, here it is...

gmarull (changes requested):
Already mentioned this is not a priority right now, so closing.

Coral-coder:
@ericmigi while you wait, mind taking a look here? 

If you want to see it, it is already built on my repo.

ericmigi:
we discussed this - it would be a huge lift to test and ensure that this doesn't break who knows what other parts of the OS. definitely cool but not something the internal team can prioritize testing/merging right now. I'd encourage you to keep working on it, running it on your own device to see how it works! Keep us updated here if you'd like

Coral-coder:
Thank you for that answer. I appreciate hearing it from you. 

It is working well on my watch with minimal impact to battery. 
It isn't perfect, I can't get all notifications from both phones at the same time everytime... but I can come and go with both phones and the transition between them is seamless. 
Only issue I run into is keeping the same apps and watchfaces on both phones... (that's a different problem tho)

ericmigi:
one of the main problems is blobdb - that's designed to only work with one phone at a time. plus apps/faces as you identified. and appmessage...and a lot of other things.

PR #1549, opened 2026-06-17
third_party/hal_sifli/sf32lb52: mark kernel RAM non-cacheable

jplexer:
Fixes FIRM-2513

### DIFF
--- a/src/fw/kernel/memory_layout.h
+++ b/src/fw/kernel/memory_layout.h
@@ -10,7 +10,7 @@
 #define KERNEL_READONLY_DATA SECTION(".kernel_unpriv_ro_bss")
 
 enum MemoryRegionAssignments {
-  // SF32LB52: SiFli's system_bf0_ap.c programs MPU regions 0..4 in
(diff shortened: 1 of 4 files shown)


gmarull (changes requested):
this needs better explanation. why cache invd ops can't be used?

jplexer:
We have some hardfaults that are in the SiFli HAL's DMA descriptor state, not our buffers, so cache ops mean clean/invalidate inside vendor code on every DMA path, all 32B-aligned. Already tried targeted flushes (4bde4ee0/3ff6d0d06), but issue still persists.

Non-cacheable kernel RAM just kills the whole problem, practically free (D-cache still covers XIP flash), and it's the known-good state from before we enabled MPU.

gmarull:
@jplexer pls share some more problem details with @rabbitsaviola

rabbitsaviola:
On mstar invalidate op is to clean first and then invalidate, that means if cache is dirty and invalidate is called,  dirty data would be written back to ram first and then cacheline is invalidated. That means there's no standard invalidate op on mstar.

In our usage, the entire internal RAM is configured as non-cachable as there's no benefit to configure it as cachable.

PR #1553, opened 2026-06-18
battery: show percent while charging

gloriousCode:
Add the current battery percentage to charging dialogs while keeping fully charged messages percentage-free.
Emery:
qemu_emery_charging_42 copy qemu_emery_fully_charged copy
Flint: 
qemu_flint_charging_42qemu_flint_fully_charged
Gabbro: 
qemu_gabbro_charging_42qemu_gabbro_fully_charged

### DIFF
--- a/src/fw/shell/normal/battery_ui.c
+++ b/src/fw/shell/normal/battery_ui.c
@@ -34,6 +34,10 @@ typedef struct {
   ResourceId warning_icon;
 } BatteryWarningDisplayData;
 
+typedef struct {
+  uint8_t percent;
+} BatteryChargingDisplayData;
+
 // UI Callbacks
 ///////////////////////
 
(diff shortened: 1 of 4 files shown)


src/fw/shell/normal/battery_ui.c
```diff
@@ -34,6 +34,11 @@ typedef struct {
   ResourceId warning_icon;
 } BatteryWarningDisplayData;
 
+typedef struct {
+  uint8_t percent;
+  char text[32];
```
ck-telecom: maybe text[32] is useless? using a temp variable instead when dialog_set_text()?

ericmigi (commented):
UI looks nice. @gmarull do you think it's accurate enough during charging? One reason we didn't show this in first era was the uncertainty was too high.

gloriousCode:
Happy to even fuzz it down to a small battery icon with three bars where the latest bar flashes at levels 0-33, 34-66, 67-100. I just wanted to see _something_ as an indicator for how much its charged. Though it really doesn't take long to charge either

gmarull:
> Happy to even fuzz it down to a small battery icon with three bars where the latest bar flashes at levels 0-33, 34-66, 67-100. I just wanted to see _something_ as an indicator for how much its charged. Though it really doesn't take long to charge either

It would be nice to see the cup getting filled, not sure how easy it is though... I'll check on device next week the accuracy of the % while charging, but I expect to be ok.

ericmigi:
cup filing would be amazing! I love that it would give a fuzzy more calm way of giving this info

ericmigi:
@gloriousCode want to dust this off? We'd be happy to merge it in.

PR #1556, opened 2026-06-18
fw: revert flint/diorite app segment to 66 KiB

gmarull:
Commit d2cd45a9 ("fw/process_management: double the app stack to 4 KiB") bumped app_segment for all platforms, taking flint (and diorite) to 68 KiB. Combined with the 30 KiB runtime_reserved this makes the App RAM region 98 KiB (0x18800), which the ARMv7-M MPU cannot express: a region must be n*2^k bytes with n <= 8 subregions, and 98 KiB = 2^11 * 49.

Since db56c6c4 the subregion math runs at runtime rather than at build time (tools/mpu_calc.py), so the illegal size no longer fails the build and instead trips PBL_CROAK in compute_block_layout() while booting asterix (a flint board):

  mpu_armv7m.c: MPU region cannot fit subregion alignment:
  base=0x20027800 size=0x18800

Revert flint and diorite (the ARMv7-M platforms) to 66 KiB, restoring the 96 KiB (0x18000) App RAM region that maps cleanly to 6 x 16 KiB subregions. ARMv8-M platforms program the limit register directly and are unaffected, so they keep 68 KiB.

### DIFF
--- a/src/fw/process_management/app_manager.c
+++ b/src/fw/process_management/app_manager.c
@@ -201,7 +201,11 @@ void prv_dump_start_app_info(const PebbleProcessMd *app_md) {
 }
 
 #define APP_STACK_JS_SIZE (8 * 1024)
+#if defined(CONFIG_PLATFORM_EMERY) || defined(CONFIG_PLATFORM_GABBRO)
 #define APP_STACK_NORMAL_SIZE (4 * 1024)
+#else
+#define APP_STACK_NORMAL_SIZE (2 * 1024)
+#endif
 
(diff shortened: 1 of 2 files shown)


jplexer (approved):
but shouldnt it only be flint?

gmarull:
> but shouldnt it only be flint?

well, diorite is (was) also armv7, is old platforms info used elsewhere? because chalk etc should also be adjusted then, original commit tweaked all

jplexer:
Old platform info is used for v2 and v3 sdk ram sizes, which actually means yes we should re-adjust everything legacy back to the original

jplexer (approved):

PR #1557, opened 2026-06-18
Move "stop mode" logic to each soc

gmarull:


src/fw/soc/qemu/freertos.c
```diff
@@ -0,0 +1,126 @@
+/* SPDX-FileCopyrightText: 2024 Google LLC */
```
jplexer, line 1: oops?

jplexer (approved):

src/fw/soc/qemu/freertos.c
gmarull: it's the copyright of origin file... (this PR just copy/pasted) it's probably not much left from google there, but original wasn't updated either

jplexer: ah gotcha

PR #1547, opened 2026-06-17
applib/text_layout: pre-load glyph bitmaps in RTL rendering path

pebble-aibot:
## Summary

- The BiDi (RTL) rendering path in `walk_line()` was missing glyph bitmap pre-loading before `render_glyph()` calls, unlike the standard LTR path which pre-loads via `text_resources_get_glyph()` to ensure cache hits
- Without pre-loading, the RTL path triggers the full flash I/O call chain (`text_resources_get_glyph → prv_load_glyph_bitmap → sys_resource_load_range`) from deep in the stack, on top of 256 bytes of RTL shaping/reversal buffers already allocated on the stack
- This contributes to KernelMain stack overflows during notification text rendering (FIRM-2575/FIRM-2643 crash family)
- Adds `text_resources_get_glyph()` pre-load calls before every `render_glyph()` in the BiDi rendering block — for RTL segments, LTR segments within mixed text, and the trailing suffix glyph

Fixes FIRM-2643
Fixes FIRM-2575

## Test plan

- [ ] Verify text rendering with RTL text (Arabic, Hebrew) in notifications
- [ ] Verify text rendering with mixed LTR/RTL text
- [ ] Verify no regression in standard LTR notification rendering
- [ ] Monitor KernelMain stack high-water mark during notification rendering with RTL content

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1561, opened 2026-06-18
applib: fix stack overflow when rendering Hebrew notifications

alonmln:
## Summary

This fixes a watch crash in the RTL text rendering path. When a notification contains Hebrew in the sender/author field, or in one of the available responses shown after pressing Select, text rendering can overflow the stack and crash the watch.

The crash comes from the Hebrew/RTL rendering path keeping several 128-byte scratch buffers on the render stack. This change keeps a single RTL output buffer on the stack, walks UTF-8 sequence starts backward instead of storing decoded codepoints in another stack array, and allocates the Arabic shaping scratch buffer only when Arabic shaping is actually needed. Hebrew rendering keeps the same visual behavior while using less stack.

## Testing

- Tested in QEMU
- Tested on physical Pebble 2 hardware
- Verified Hebrew notification author and response text no longer crash the watch
- gitlint --commits origin/main..HEAD
- git diff --check origin/main..HEAD

### DIFF
--- a/src/fw/applib/graphics/rtl_support.c
+++ b/src/fw/applib/graphics/rtl_support.c
@@ -8,9 +8,8 @@
 
 #include <string.h>
 
-// Maximum number of codepoints we can handle in a single reversal
-// Keep small for stack safety on embedded systems (32 codepoints * 4 bytes = 128 bytes)
-// 32 is sufficient for all real Hebrew words including long morphological forms
+// Maximum number of codepoints we can handle in a single reversal.
+// 32 is sufficient for real Hebrew words including long morphological forms.
 #define MAX_RTL_CODEPOINTS 32
 
 bool utf8_contains_rtl(const utf8_t *start, const utf8_t *end) {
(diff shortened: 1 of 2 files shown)


gmarull (approved):

gmarull:
Worth noting this overlaps with #1547, which targets the same KernelMain RTL stack overflow (FIRM-2575/2643) from the opposite end: #1547 pre-loads glyphs so the deep flash-load chain (`text_resources_get_glyph → prv_load_glyph_bitmap → sys_resource_load_range`) no longer runs with `render_glyph`'s frame stacked under it, while this PR shrinks the buffers (`shaped_buffer`/`codepoints[]`) that sit on the stack throughout the loop.

They are complementary, not competing — each reduces a different term of the same high-water sum, and applying both gives the largest margin. Suggest coordinating so they land together rather than one superseding the other.

alonmln:
@gmarull thanks for the quick follow up!
What do you suggest I do? Merge `fix/text-render-preload-rtl-glyphs` into this branch?

gmarull:
> @gmarull thanks for the quick follow up! What do you suggest I do? Merge `fix/text-render-preload-rtl-glyphs` into this branch?

both seem to target same issue, it's worth analyzing the two PRs

alonmln:
I checked #1547. Agreed, they look complementary rather than competing.

#1547 avoids entering the deep glyph/resource-loading path from inside `render_glyph()` in the BiDi path. This PR reduces the RTL scratch-buffer stack usage itself.

So they reduce different parts of the same stack high-water mark. I don’t think one supersedes the other; landing both should give the best margin.

PR #1563, opened 2026-06-18
Improvements for sleep level management in SF32LB

gmarull:
With this change, deep WFI will be used more often (e.g. during I2C transfers). The stop mode API will be removed soon (on nRF5 we just need an API to gate "full WFI" (WFI + flash sleep, etc.)). The whole stop mode API concept was broken/poorly implemented.


gmarull:
merging to tag new release

PR #1573, opened 2026-06-19
CPU governor: adaptive 3-step frequency scaling + idle-draw reductions

Coral-coder:
## Summary

Adds an adaptive 3-step CPU governor that scales the SF32LB52x between three frequency tiers based on actual workload, plus a set of idle-draw reductions. This keeps the UI and media playback responsive under load while lowering background power when idle.

The motivating problem: on a fixed clock the UI/compositor could fall behind in realtime (e.g. music playback time lagging) because the CPU did not ramp up for render work. The governor now ramps the CPU from render-time feedback so frames draw on time.

## Changes

1. **`powermode: add adaptive 3-step CPU governor`** — cpumode gains a tier-frequency setter and drops the unused 24 MHz idle frequency. The powermode service runs three tiers — light (48 MHz), medium (144 MHz), high (240 MHz) — stepping up on demand or when a unit of work overruns its tier budget, and stepping back down through a short boost/tail window. A boot gate holds high performance until the launcher is up.
2. **`compositor: drive CPU governor from render timing`** — measures the CPU-bound cost of each app/modal render and reports it to the governor, and requests a brief high-performance boost around each display flush. A render-depth guard collapses nested app+modal renders into a single report.
3. **`kernel: wire CPU governor into boot and app lifecycle`** — boot-complete signal, app/HRM high-performance holds, removes the modal manager's standalone hold now that the governor owns scaling, and guards the calls out of PRF builds.
4. **`power: reduce idle draw across ALS, battery, and touch`** — polls the fuel gauge less often and skips PMIC reads when stop mode is blocked, extends the ALS cache and Obelix backlight thresholds, lengthens the CST816 touch wedge watchdog, and relaxes stationary/background-activity poll intervals.

## Notes

- Deep sleep entry threshold is unchanged (`RTC_TICKS_HZ/20`), so tap-to-wake / shake-to-wake behavior is preserved.
- No Bluetooth / connectivity changes are included.
- Draft pending on-device validation of realtime rendering.

### DIFF
--- a/include/pbl/services/activity/activity_private.h
+++ b/include/pbl/services/activity/activity_private.h
@@ -52,7 +52,7 @@ typedef uint16_t ActivityScalarStore;
 #define ACTIVITY_LAST_SLEEP_MINUTE_OF_DAY  (21 * MINUTES_PER_HOUR)
 
 // Default HeartRate sampling ON time
-#define ACTIVITY_DEFAULT_HR_ON_TIME_SEC (60)
+#define ACTIVITY_DEFAULT_HR_ON_TIME_SEC (45)
 
 // Turn off the HR device after we've received X good quality samples
 #define ACTIVITY_MIN_NUM_GOOD_SAMPLES_SHORT_CIRCUIT (10)
--- a/include/pbl/services/activity/activity_private.h
+++ b/include/pbl/services/activity/activity_private.h
@@ -52,7 +52,7 @@ typedef uint16_t ActivityScalarStore;
 #define ACTIVITY_LAST_SLEEP_MINUTE_OF_DAY  (21 * MINUTES_PER_HOUR)
 
 // Default HeartRate sampling ON time
-#define ACTIVITY_DEFAULT_HR_ON_TIME_SEC (45)
+#define ACTIVITY_DEFAULT_HR_ON_TIME_SEC (60)
 
 // Turn off the HR device after we've received X good quality samples
 #define ACTIVITY_MIN_NUM_GOOD_SAMPLES_SHORT_CIRCUIT (10)
--- a/include/pbl/services/powermode_service.h
+++ b/include/pbl/services/powermode_service.h
@@ -4,9 +4,14 @@
 #pragma once
 
 #include <stdbool.h>
+#include <stdint.h>
 
 //! Initialize the power mode service.
 void powermode_service_init(void);
+
+//! Called once the launcher has finished booting. Low-power CPU mode is not
+//! entered until this runs so early init stays at high performance.
+void powermode_service_boot_complete(void);
 
 //! Enable or disable the power mode service. When disabled, request and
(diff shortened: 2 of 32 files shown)


Coral-coder:
I absolutely do not expect this would get merged… 
But it does a great job for me at least of extending battery.

gmarull:
Thanks @Coral-coder for the PR. Since we're actually working on power analysis, I took a quick shot at the PR on a PR2 (Idle on tictoc, connected at slow interval). What really matters from this PR is the last commit, ie, reduce what we do "too often". We have a binary frequency scaling mechanism (really primitive, under debug settings, the one you are extending). From our experience, it's not too effective, even though you gain something (but at the cost of complexity, potential latencies, etc.). What matters most in SF32LB chips is to spend as much as possible on deepsleep, which we definitely do not right now as last commit suggest (15ms out of every second are not in that state right now, even when 'idle'). Note on a watch this is where we spend most of the time (or we should), so it's the primary path to optimize. Other scenarios could potentially benefit from a governor, or stuff like GPU, using native color formats, graphics compression, etc.

This PR:

Screenshot From 2026-06-19 21-27-12

Main:

Screenshot From 2026-06-19 21-31-05

Main + 910396c148b150d456133e9978e9fee649e417db

Screenshot From 2026-06-19 21-35-58

cc: @jplexer

Coral-coder:
That’s awesome! thanks for taking a look and for the insight! 😁

I noticed also that it does a check for daylight savings time every second
As well as some other things that don’t really need to run every second like the battery check.
I can’t recall if I already changed that on my end…
When I changed the Deep Sleep on my end, odd quirks started to pop up like frame render misses and serious lagging… 

I also experimented with 24 MHz and forcing deep sleep whenever it wasn’t drawing the screen but my implementation was buggy. 
It didn’t like to wake from 24mhz (accel/touch)

Coral-coder:
trimmed back this PR

Coral-coder:
.3 % / HR

Coral-coder:
now as low as .1 %/ HR !!!

jplexer:
hey! some pretty cool stuff, doing some further testing with this today

Coral-coder:
OOOH!!!! I can't wait to see the power graph comparisons!

Just to note... I don't have this set to both power profiles, so you will have to set the profile to low power in debug settings.

gmarull (changes requested):
Thanks for the proposal, however, these kind of changes are risky and can impact user experience. While some ideas make sense, it's unclear whether they have a true impact or not. For PRs like these to have a chance to get into mainline, the author must submit them in small chunks, and proof that each chunk improves things (eg by showing current traces).

Coral-coder:
Removing Governor, trimming all the rest out into bite sized prs, and resubmitting... (the governor does work tho... it isn't necessary to always run at high power)

PR #1575, opened 2026-06-19
Remove legacy persist

gmarull:


jplexer (approved):

gmarull:
@jplexer did AI get it right that files with old scheme cannot exist?

jplexer:
> did AI get it right that files with old scheme cannot exist?

mhh no, just checked further, we would need a migration
2.x was was p%06d and 3.x until now was ps%06d

PR #1588, opened 2026-06-22
applib: reorder bidi boundary spaces and measure Arabic ligatures

kaluaim:
## What

Two bidi line-layout fixes in `text_layout.c` that share the same segment-collection loop, so they ship together:

  - **Boundary spaces** — a space between an RTL and an LTR run is direction-neutral. Kept inside a run, it was reversed with that run, and the segment reorder then carried it to the run's far edge, collapsing the gap between the two runs. Peel trailing spaces into their own neutral segment so they stay put between the runs.
  - **Ligature width** — width measurement shaped each codepoint in isolation, so a Lam-Alef pair is measured as two glyphs while the renderer draws one ligature. Measure with `arabic_shape_pair()` (skipping the consumed Alef) so layout width matches what is drawn. This also covers the line/truncation measurement path (`update_dimensions`), closing the measurement gap raised on #1585.

Adds `rtl_segment_content_end()` and raises `MAX_BIDI_SEGMENTS` to 16.

## Depends on

Depends on **#1585** (`arabic_shape_pair()`). #1586 is now merged, so its digit commit is dropped and this PR has been rebased on `main`. The PR still shows #1585's commit until it merges — **please merge after #1585**, after which a rebase narrows the diff to just the `text_layout` change.

## Before / After

cmp_rtlltr

## Test

`test_rtl_support` (incl. content_end) plus `test_line_layout`, `test_char_iterator`, `test_word_iterator` pass.

### DIFF
--- a/src/fw/applib/graphics/text_layout.c
+++ b/src/fw/applib/graphics/text_layout.c
@@ -263,15 +263,25 @@ bool word_init(GContext* ctx, Word* word, const TextBoxParams* const text_box_pa
   Codepoint curr_cp = utf8_iter_state->codepoint;
   WordState state = WordStateStart;
   state = word_state_update(state, curr_cp);
+  // arabic_shape_pair() can fold this codepoint and the next into a single
+  // glyph (the Arabic Lam-Alef ligature), reporting the next one as consumed.
+  // Its advance is then already counted, so skip it on the next iteration.
+  bool skip_ligature_member = false;
 
   do {
     iter_next(&char_iter);
     Codepoint next_cp = utf8_iter_state->codepoint;
 
     if (state == WordStateGrowing || state == WordStateIdeograph) {
-      Codepoint width_cp = arabic_shape_codepoint(prev_cp, curr_cp, next_cp);
-      word->width_px += prv_codepoint_get_horizontal_advance(&ctx->font_cache,
-          text_box_params->font, width_cp);
+      if (skip_ligature_member) {
+        skip_ligature_member = false;
+      } else {
+        bool consumed_next = false;
+        Codepoint width_cp = arabic_shape_pair(prev_cp, curr_cp, next_cp, &consumed_next);
+        word->width_px += prv_codepoint_get_horizontal_advance(&ctx->font_cache,
+            text_box_params->font, width_cp);
+        skip_ligature_member = consumed_next;
+      }
     }
 
     prev_cp = curr_cp;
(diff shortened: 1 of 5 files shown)


src/fw/applib/graphics/text_layout.c
```diff
@@ -263,15 +263,25 @@ bool word_init(GContext* ctx, Word* word, const TextBoxParams* const text_box_pa
...
-      word->width_px += prv_codepoint_get_horizontal_advance(&ctx->font_cache,
-          text_box_params->font, width_cp);
+      if (skip_ligature_member) {
+        skip_ligature_member = false;
+      } else {
+        bool consumed_next = false;
+        Codepoint width_cp = arabic_shape_pair(prev_cp, curr_cp, next_cp, &consumed_next);
+        word->width_px += prv_codepoint_get_horizontal_advance(&ctx->font_cache,
+            text_box_params->font, width_cp);
+        skip_ligature_member = consumed_next;
```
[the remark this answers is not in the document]
kaluaim: Addressed in `388e9e4`. The line-width and truncation/ellipsis pass goes through `walk_line()`'s general loop + `update_dimensions_char_visitor_cb`, which measured each codepoint separately — so a Lam-Alef pair counted as two glyphs while `word_init` and the render path count one.

Now that loop measures with `arabic_shape_pair()` and gives the consumed Alef zero width, and the dimension visitor reuses the advance the loop already computed instead of recomputing per codepoint. So word width, line width, and the truncation pass all count the ligature once and agree with what is drawn — and `last_visited_char` can no longer land between the Lam and the Alef. Non-Arabic text is unaffected (`arabic_shape_pair` is a no-op there).

Added `test_line_layout__lam_alef_width_counts_ligature_once`: `بلا` (Beh+Lam+Alef) measures 2 advances, not 3.

gmarull:
@kaluaim needs rebase

src/fw/applib/graphics/arabic_shaping.c
```diff
@@ -249,6 +249,50 @@ Codepoint arabic_shape_codepoint(Codepoint prev_cp, Codepoint curr_cp, Codepoint
   return prv_get_shaped_codepoint(entry, form);
 }
 
+// Lam-Alef mandatory ligatures: Lam (0x0644) followed by an Alef variant
+// collapses to a single glyph. Only isolated and final forms exist (Alef
+// never joins to a following letter).
+typedef struct __attribute__((__packed__)) {
```
gmarull: any reason to pack here?

kaluaim: No real reason — I'd mirrored the existing `ArabicShapingEntry` (also packed). It's three `uint16_t` (6 bytes, no padding), so packing is a no-op. Dropped in `6a13db2`.

gmarull (approved):

gmarull:
rebase

PR #1599, opened 2026-06-24
drivers/display/sf32lb: run LCDC ISR from RAM [FIRM-1740]

jplexer:
If the JDI half-frame LCDC interrupt slips past max_line, its mid- and end-of-frame interrupts merge, EOF is never sent, and the compositor wedges until the silent-loss watchdog fires.

Fix: put the reprogram path (LCDC1_IRQHandler, display_jdi_irq_handler, HAL_LCDC_IRQHandler) in .ramfunc so it runs while flash is busy.


gmarull (commented):
Note: why does this fix the problem? ie, why flash fetch interferes with LCDC? @rabbitsaviola ?

PR #1602, opened 2026-06-24
enable psram on emery/gabbro

devsnek:
only tested on emulator so far (https://github.com/coredevices/qemu/pull/6). hardware init is mostly copy/paste from sifli.

### DIFF
--- a/src/fw/soc/sf32lb/sf32lb52x/init.c
+++ b/src/fw/soc/sf32lb/sf32lb52x/init.c
@@ -9,6 +9,59 @@
 #define HCPU_FREQ_MHZ 240
 #define PWRKEY_RESET_CNT (32000 * 15)
 
+#define PBL_PSRAM_SIZE_MB 16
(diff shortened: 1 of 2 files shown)


gmarull (changes requested):
We're not going to enable PSRAM just for the sake of enabling it. As of today, it has no usage on the system, and yet it consumes some power by just being on.

devsnek:
@gmarull I would have been happy to discuss other approaches as well, for example only enabling psram if an app requests it. I don't think just closing a PR like this is a very appropriate behavior for a project seeking open source contributions.

gmarull:
> @gmarull I would have been happy to discuss other approaches as well, for example only enabling psram if an app requests it. I don't think just closing a PR like this is a very appropriate behavior for a project seeking open source contributions.

Happy to discuss as soon as the proposal comes with context, example usecases, etc. Features reaching userspace have wide implications.

PR #1620, opened 2026-06-28
Obelix/Getafix prog and flash memory map

ck-telecom:


jplexer (approved):
Cool! I would also add real image of how the watch connects to the pblprog but can be added later

PR #1333, opened 2026-05-20
i18n: add Arabic (ar_SA) language pack

AhmedTheGeek:
## Summary

Adds an Arabic (ar_SA) language pack and two small fixes in `applib/graphics` needed to make multi-word Arabic notification bodies display without truncation. Built on top of the Arabic shaping and RTL work in #783.

### Render fix
`arabic_shape_text` silently dropped codepoints past index 16. Multi-word RTL lines (e.g. `لا تنسَ شراء الخبز` = 18 codepoints with spaces) lost their trailing letters before reversal and render, producing visible mid-word truncation like `لا تنسَ شراء الخ` instead of the full `الخبز`.

- `MAX_SHAPE_CODEPOINTS` 16 -> 64 in `arabic_shaping.c` so the shape array matches the per-line buffer ceiling.
- `shaped_buffer` / `rtl_buffer` in `walk_line` 64 -> 128 bytes so the larger shape output has room (Arabic presentation forms at U+FE70+ are 3-byte UTF-8 vs the 2-byte basic block, so shaping expands the byte length).

That's it for the firmware change. No layout-side modifications, so existing unit tests (menu-cell pixel diffs etc.) are unaffected.

### Language pack
- `resources/normal/base/lang/ar_SA/` with Noto Sans Arabic Regular + Bold, a 397-codepoint subset covering U+0600-U+06FF and the Arabic Presentation Forms-B range (U+FE70-U+FEFC) that `arabic_shape_text` emits, a `lang_map.json` modelled on the Hebrew template, and the upstream `LICENSE.NotoSansArabic.txt` (SIL OFL 1.1).
- One-line addition to `FontResourceGenerator`: a `pixelHeight` field in lang_map.json entries that overrides the height parsed from the slot name. Used by the ar_SA pack to render Arabic glyphs at 12/14/17/20 px in the 14/18/24/28 slot reservations so they sit at the same visual height as the surrounding Latin Gothic instead of overrunning the line metrics (Arabic font em-boxes reserve vertical space for diacritics).

### Tooling
- `tools/install_lang_pack.py` pushes a `.pbl` to a running emulator via the pypkjs websocket using `PutBytes` with `ObjectType.File` and filename `lang` (the PFS path the resource storage layer watches, per `resource_storage_file.c` and `pfs_resource_table.auto.c`).

## Test plan

- [x] Builds clean for `qemu_emery`, `qemu_flint`, and `qemu_gabbro`.
[...]

PR #1622, opened 2026-06-28
applib: treat Arabic diacritics as transparent for cursive joining

kaluaim:
## What

Arabic combining marks (harakat — fatha, damma, kasra, shadda, sukun…) sit on a base letter and must not break the cursive connection between the letters around them. `arabic_shape_text` used each codepoint's immediate neighbours as the joining context, so a mark between two letters shaped them as if disconnected — e.g. a shadda before a Lam-Alef left the ligature isolated instead of joined. Vocalized words fragmented.

Skip transparent marks (U+0610–061A, U+064B–065F, U+0670, U+06D6–06ED) when choosing the previous/next letter for shaping, and carry any mark consumed inside a Lam-Alef pair through to the output.

## Before / after

Notification `ثَلَاثَة بِلَا مُلَاحَظَات` (fatha / kasra / damma). Before: words fragment where the marks break the joins; after: they connect.

cmp_nsm

## Test

`diacritic_transparent_join`: a shadda between Meem and Lam-Alef keeps the Meem medial and the ligature final (U+FEFC), not isolated (U+FEFB). `./pbl test` passes.

### DIFF
--- a/src/fw/applib/graphics/arabic_shaping.c
+++ b/src/fw/applib/graphics/arabic_shaping.c
@@ -293,6 +293,16 @@ Codepoint arabic_shape_pair(Codepoint prev_cp, Codepoint curr_cp, Codepoint next
   return arabic_shape_codepoint(prev_cp, curr_cp, next_cp);
 }
 
+bool arabic_is_transparent(Codepoint cp) {
+  return (cp >= 0x0610 && cp <= 0x061A) ||
+         (cp >= 0x064B && cp <= 0x065F) ||
+         (cp == 0x0670) ||
+         (cp >= 0x06D6 && cp <= 0x06DC) ||
+         (cp >= 0x06DF && cp <= 0x06E4) ||
+         (cp >= 0x06E7 && cp <= 0x06E8) ||
+         (cp >= 0x06EA && cp <= 0x06ED);
+}
+
 size_t arabic_shape_text(const utf8_t *src, size_t src_len,
                          utf8_t *dest, size_t dest_size) {
   if (src == NULL || dest == NULL || src_len == 0 || dest_size == 0) {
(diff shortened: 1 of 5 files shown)


ericmigi:
Which Arabic language pack are you using? We're looking to include one in the language drop-down

kaluaim:
> Which Arabic language pack are you using? We're looking to include one in the language drop-down

@ericmigi  I'm using **Vazirmatn** ([ar_SA-vazirmatn.pbl.zip](https://github.com/user-attachments/files/29457479/ar_SA-vazirmatn.pbl.zip))  Not my favorite, but in my opinion the best of what worked on Pebble so far — I'll keep looking for others.

@AhmedTheGeek also added ar_SA from Noto in #1333 ([pack here](https://github.com/coredevices/PebbleOS/pull/1333#issuecomment-4501021842)) — so Noto or Vazirmatn both work.

gmarull (approved):

ericmigi:
Which one do you think we should include as the default in the languages drop-down? We'll defer to the experts here 😀

kaluaim:
@ericmigi I’d say **Vazirmatn** ([ar_SA-vazirmatn.pbl.zip](https://github.com/user-attachments/files/29457479/ar_SA-vazirmatn.pbl.zip)) for now.

I’m still on the hunt for the perfect Arabic font that works with Pebble and will update you once I find it 😄

AhmedTheGeek:
I agree with @kaluaim I tried the Vazirmatn version it looks way better than Noto.

ericmigi:
cool! can you do a PR to github.com/coredevices/mobileapp to include that file in the list of language packs please?

ericmigi:
we don't bake language packs into the OS

kaluaim:
@ericmigi  Done! → coredevices/mobileapp#280

PR #1627, opened 2026-06-29
fw/comm/ble: hand clients a single service instance, rotating on failure [FIRM-2018]

jplexer:
Some peer GATT servers (notably buggy Android companions) leave a stale duplicate of a service behind -- same UUID, different handle range -- when they re-add services without removing the old one. PPoGATT then started a handshake (meta read) per instance, and the concurrent reads deadlocked the single ATT bearer when the stale instance never answered, so the whole connection failed. With two PPoGATT instances the meta read failed with ENOTCONN once the phone gave up and disconnected; with one it succeeded.

Hand each kernel LE client exactly one instance: collect every instance that exposes all required characteristics, and try the newest (highest-handle) one first. We can't tell which instance is live at discovery time, so PPoGATT advances the selection after a failed handshake and resets it on a successful session; across reconnects this cycles through the complete instances until the live one connects. Newest-first matches the Android re-add bug, where the server only retains a reference to the most recently added characteristics.

This drops support for multiple concurrent instances of the same service UUID (legacy multi-app PPoGATT); the Core companion uses a single Hybrid server.

### DIFF
--- a/src/fw/comm/ble/kernel_le_client/kernel_le_client.c
+++ b/src/fw/comm/ble/kernel_le_client/kernel_le_client.c
@@ -205,37 +205,78 @@ static void prv_handle_all_services_invalidated(void) {
   }
 }
 
+//! Selects which instance of a duplicated service we hand to a client. Advanced
+//! by PPoGATT after a failed handshake and reset after a successful session, so
+//! that across reconnects we eventually try every instance. See
+//! prv_handle_services_added and kernel_le_client.h.
+static uint8_t s_service_instance_attempt;
+
+void kernel_le_client_reset_service_instance_attempt(void) {
+  s_service_instance_attempt = 0;
+}
+
+void kernel_le_client_advance_service_instance_attempt(void) {
+  s_service_instance_attempt++;
+}
+
 static void prv_handle_services_added(
     PebbleBLEGATTClientServicesAdded *added_services, BTDeviceInternal *device) {
-  // loop through the new services
-  for (int s = 0; s < added_services->num_services_added; s++) {
-    // get the uuid for the service
-    Uuid service_uuid = gatt_client_service_get_uuid(added_services->services[s]);
+  for (int c = 0; c < KernelLEClientNum; c++) {
+    const KernelLEClient * const client = &s_clients[c];
 
-    // are any clients looking for this uuid?
-    for (int c = 0; c < KernelLEClientNum; c++) {
-      const KernelLEClient * const client = &s_clients[c];
-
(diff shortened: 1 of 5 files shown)


src/fw/comm/ble/kernel_le_client/kernel_le_client.c
```diff
@@ -205,37 +205,78 @@ static void prv_handle_all_services_invalidated(void) {
...
-    Uuid service_uuid = gatt_client_service_get_uuid(added_services->services[s]);
-
-    // are any clients looking for this uuid?
-    for (int c = 0; c < KernelLEClientNum; c++) {
-      const KernelLEClient * const client = &s_clients[c];
+  for (int c = 0; c < KernelLEClientNum; c++) {
+    const KernelLEClient * const client = &s_clients[c];
 
+    // Collect every complete instance of this client's service. Some peer GATT
+    // servers (notably buggy Android companions) leave a stale duplicate of a
```
gmarull, line 228: has this been confirmed to be buggy Android, or a bug in the mobile app?

gmarull: @sjp4

sjp4: https://linear.app/core-dev/issue/FIRM-2018/nimble-bug-pt2-fails-to-connect-to-android-12-stuck-in#comment-7ffcb7b5 seems like they enable the option to ignore other pebble apps being installed, so that is the likely cause

gmarull: So I guess that's it, no workarounds needed on our side.

sjp4: I'm not sure to be honest - they enabled the pref to allow other apps to be installed without blocking connection, but looking again, it doesn't show another app being detected (could be a different package that we don't know about / a custom built coreapp or something). I'd say that rotating between the two is probably a complex workaround - maybe a really simple workaround to just use the first one it finds, rather than get stuck?

gmarull: note: won't reverse ppog fix all these problems?

sjp4: That's the hope (although I supposed we still need a mechanism to stop multiple clients subscribing to it)

gmarull: that's easier to handle, let's wait till then.

PR #1628, opened 2026-06-29
fw/kernel/fault_handling: kill process on app/worker-heap OOM

jplexer:
Moddable watchfaces (e.g. "Pebble Digital Classic", FIRM-2725) run their whole event loop inside the privileged `moddable_createMachine` syscall, so an app-heap OOM hits the privileged branch of `trigger_oom_fault` and reboots the whole OS — crash-looping the device into recovery + re-pair.

Fix: route by which heap ran out, not privilege. An OOM on a process's own app/worker heap kills just that process via `sys_app_fault`; only a kernel-heap OOM (or kernel-task OOM) still reboots. The kill path is gated on a safe context (not in an ISR, scheduler running). OOM-only — hardfaults still reboot.

Tested: builds for `obelix@pvt`.

Fixes FIRM-2725

### DIFF
--- a/src/fw/kernel/fault_handling.c
+++ b/src/fw/kernel/fault_handling.c
@@ -9,8 +9,11 @@
 
 #include "applib/app_logging.h"
 #include "kernel/memory_layout.h"
+#include "mcu/interrupts.h"
 #include "mcu/privilege.h"
 #include "pbl/services/system_task.h"
+#include "process_state/app_state/app_state.h"
+#include "process_state/worker_state/worker_state.h"
 #include "syscall/syscall.h"
 #include "syscall/syscall_internal.h"
 #include "system/logging.h"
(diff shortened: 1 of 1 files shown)


src/fw/kernel/fault_handling.c
```diff
@@ -157,19 +160,39 @@ NORETURN trigger_fault(RebootReasonCode reason_code, uint32_t lr) {
...
-      reboot_reason_set(&reason);
-      reset_due_to_software_failure();
-  } else {
+  // An OOM on a process's own heap is the process's fault, not the kernel's, even
+  // when we happen to be privileged because the allocation came from a syscall.
+  // Rocky/Moddable apps run their whole event loop inside the moddable_createMachine
+  // syscall, so they are always privileged; without this a watchface that exhausts
+  // its app heap would reboot (and crash-loop) the whole system. Kill just the
+  // offending process instead: sys_app_fault cooperatively marks the current task
+  // safe-to-kill and suspends it, which is safe from privileged context too.
```
gmarull: extremely verbose comments, also mentions Rocky which no longer exists

jplexer: i love claude not listening to claude.md

gmarull (approved):

PR #1634, opened 2026-06-30
Revert #1587

jplexer:
Reverts #1587

### DIFF
--- a/src/fw/applib/graphics/text_render.c
+++ b/src/fw/applib/graphics/text_render.c
@@ -66,17 +66,10 @@ void render_glyph(GContext* const ctx, const uint32_t codepoint, FontInfo* const
   // Bitfiddle the metrics data:
   GRect glyph_metrics = get_glyph_rect(glyph);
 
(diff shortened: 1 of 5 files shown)


gmarull (approved):

gmarull:
@kaluaim FYI

kaluaim:
@jplexer @gmarull Thanks for the heads-up on the revert. Could you share what regressed, or steps to reproduce it? I'd like to look into it and fix the issue so the change can be re-submitted — this baseline alignment is important for correct Arabic rendering.

jplexer:
https://www.reddit.com/r/pebble/comments/1uj01b1/latest_update_seems_to_have_broken_my_favourite/
see this reddit thread
I think you're handling this at the wrong spot, I think this is something that should be adjusted at language pack generation time

kaluaim:
@jplexer Apologies for the inconvenience here. You were right — I was looking at this the wrong way; doing it at language-pack generation time is the correct fix. I've implemented that in #1649. Thanks for the guidance!

PR #1636, opened 2026-06-30
Add QEMU protocols to inject health values (steps, heart rate, etc.)

jplexer:


src/fw/board/boards/board_qemu_emery.c
```diff
@@ -85,6 +88,15 @@ static MicDevice MIC_DEVICE = {
 };
 MicDevice * const MIC = &MIC_DEVICE;
 
+#ifdef CONFIG_HRM
```
gmarull: redundant?

gmarull (approved):

Issue #1423, opened 2026-06-02
Voice dictation drops the last word / end of last word (audio tail not flushed on stop)


Emilien-Etadam:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Current Behavior

During voice dictation, the transcription frequently loses the last word, or
the end of the last word, roughly one time out of two. This is independent of
the transcription backend; the audio sent to the phone is missing its tail.

Root-cause analysis (traced in source, not yet verified on hardware). Three
mechanisms in the firmware capture/stop path discard the end of the recording:

1. Stop ordering in `prv_stop_recording` (src/fw/services/voice/voice.c).
   `audio_endpoint_stop_transfer()` is called before `mic_stop()`. The first
   call sends StopTransfer and invalidates the session, so any audio captured
   between it and the actual mic stop can no longer be sent. The inline comment
   states this ordering is intentional, but its effect is that the tail frames
   are dropped.

2. Residual stranded in the PDM driver (src/fw/drivers/mic/sf32lb52/pdm.c).
   `prv_dispatch_samples_system_task` only delivers full frames
   (`if (available < frame_size_bytes) break;`) and the loop is gated by
   `is_running`. At stop, the full frames not yet dispatched plus the trailing
   sub-frame remainder are left in the circular buffer and never delivered.
   This is what makes a whole word disappear, not just a few ms.

3. No final Speex flush (src/fw/services/voice/voice.c +
   src/fw/services/voice/voice_speex.c). `prv_audio_data_handler` drops any
   non-full final frame (`if (sample_count != expected_samples) return;`), and
   the Speex encoder is never flushed (no zero-padded final frame, no trailing
   frame to clear the wideband QMF delay).

A secondary amplifier: `audio_endpoint_add_frame`
(src/fw/services/audio_endpoint/service.c) uses a never-block write and drops
frames when the BT send buffer is full (`s_dropped_frames++`). Under BLE
congestion at the end of an utterance this removes the final frames, which is a
likely reason the loss is intermittent.

### Expected Behavior

The complete utterance, including the last word, is captured, encoded, and sent
to the phone before the session is torn down.

### Steps To Reproduce

1. Start a dictation (e.g. reply to a notification).
2. Speak a short phrase and let it end normally.
3. Observe that the transcription is frequently missing the last word or its
   ending, intermittently across attempts.

### Version

v4.9.184

### Host OS

N/A

### Watch

Pebble Time 2 (Obelix)

### Anything else?

[...]

Issue #719, opened 2026-01-16
Raw Audio API


andree182:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

Access to raw audio data would be great, so that apps like e.g. guitar tuners, tone generators or navigation instructions player could be created.

### Alternatives

Using internal API or raw access to hardware registers.

pizzaisdavid:
I don't know how it should work and many people probably already know this, but I learned that there is a dictation API:
https://developer.repebble.com/guides/events-and-services/dictation/

But it doesn't return any audio raw audio data but the transcribed text.
It could probably still be used as a starting point.

This seems to be some corresponding code:
https://github.com/coredevices/PebbleOS/blob/a84c84a124c38583b8a73091440a9d1d6e141a10/src/fw/applib/voice/dictation_session.c#L158

matejdro:
My use case for this goes hand in hand with #416 - after watchapps can have more storage, we could record voice dictation offline and then transfer/transcribe it after bluetooth reconnects

jpmtrabbold:
I would love this. As I said in this [duplicate](https://github.com/coredevices/PebbleOS/issues/789), I was vibe-coding an app to simulate the Index behavior (take a note by holding one of the buttons, and sending to an llm to sort things into reminders, events, notes, etc) and it works great when it's offline, but one feature that I couldn't implement was to store the dictation audio when the watch is offline, so that it can sync it when it pairs with the phone again.....so the raw audio would come in handy.

On the other hand... @ericmigi, do you foresee giving the pebble watch a bit of that Index behavior? It's been fun to develop that for myself, but if that's in the roadmap I probably shouldn't spend so much time on it...

ericmigi:
Yup that's on the roadmap!

Keatr0n:
A raw audio api would be really nice for things like decibel meters too. I've been vibe coding a watch face and things like a warning if the environment is too loud would be nice.

GeoffreyPlitt:
+1 Can't wait for this. I'm dying to make a dictation app that allows long silences, which isn't possible with the current dictation API. Or field recorder.

eiis1000:
Another +1! Getting raw audio would mean I can use more sophisticated transcription methods, and also means I can save both the audio *and* the transcription...

PR #1641, opened 2026-07-01
voice: add on-device audio recording with playback and SDK API

julpel8:
## Summary

Adds on-device audio recording to flash, playback through the watch speaker, deferred
transcription through the phone, and a companion protocol so a phone can manage stored
recordings.

Recordings use the existing Speex codec and can be created without a phone connection.
Once Bluetooth is available, an app — or the phone itself — can send a stored recording
through the existing dictation transport and receive the transcription result.

## What's included

**On-device microphone recording**

- Stores Speex-encoded audio in PFS.
- Uses pre-allocated temporary files to avoid flash erases during capture.
- Compacts recordings to their final size when stopped.
- Auto-finalizes a recording after 120 s; valid recordings share a 1 MiB storage budget.
- Supports enumeration (including paged listing), deletion, validation, and cleanup.
- Handles recording-ID wraparound without overwriting a stored recording.
- Enforces recording ownership for third-party apps, and deletes an app's recordings when
  it is uninstalled.
- Keeps recording and live dictation mutually exclusive.
- Reports a failure reason (busy, mic busy, storage full, …) for diagnostics and UI.

**Speaker playback**

- Decodes stored Speex frames to 16 kHz PCM.
- Streams audio through the existing speaker service.
- Isolates playback sessions from stale callbacks via session IDs.

**Deferred transcription**

- Streams stored Speex frames to the phone without transcoding.
- Paces transfers to prevent send-buffer overflow and dropped frames.
- Reports upload progress to the UI.
- Prevents deletion of a recording while it is being transcribed.
- Delivers results through the existing dictation event path, correlated with the
  originating app or window.

**SDK API**

- Adds APIs to start, stop, cancel, query, enumerate, play, delete, and transcribe
  recordings.
- Includes the required syscalls and exported SDK symbols.
- Bumps the SDK revision to 110 (process info SDK version 5.0x6b).
- Returns transcription results asynchronously through a callback.

**Companion management protocol**

- New message IDs `0x10`–`0x17` on the existing voice endpoint (11000 / `0x2af8`) for
  listing, deleting, playing/stopping, and transcribing recordings.
- Advertises `voice_recording_management_support` in the system capabilities bitfield,
  only on builds with both `CONFIG_MIC` and `CONFIG_SERVICE_VOICE_ENDPOINT`.
- Request handlers tolerate trailing bytes for forward compatibility and reject malformed
  requests with an explicit result code.
- Documented in `docs/reference/voice_recordings.md` (architecture, app API, limits, wire
  format, result codes, implementation map).

**Settings test interface**

- Adds a "Voice Memos TEST" page to record, play, transcribe, and delete recordings
  directly on the watch.
- Also exposes recording quality and record/playback gain. These tuning controls are
  internal and deliberately not part of the SDK.
- This currently serves as a test interface. Whether it should remain as a user-facing
  Voice Memos interface is an open question.

## Open questions

**Recording lifecycle:** Recordings currently persist until deleted (by the app, the
settings UI, the phone, or app uninstall) and are capped by a 120 s duration and a 1 MiB
budget. Should the OS instead apply a retention policy, or delete automatically after a
successful transcription? What should happen when the budget is reached — reject, or evict
the oldest recording?

**Audio quality:** Quality and gain are configurable through the settings menu but not
exposed to apps. Should they stay internal, become fixed defaults, or be part of the SDK?

**Storage information:** Should the watch display the storage used and available for
recordings? If so, should this represent the dedicated recording budget or the actual free
flash space? Should this information also be exposed through the SDK and over the
companion protocol?

**Settings UI:** Should the current settings page remain available as a user-facing Voice
Memos interface, or should it be removed once SDK and companion integration are validated?

**Protocol result codes:** `Unsupported` and `Failed` are defined but never emitted today;
some start failures are reported as `Not found`. Worth splitting before a companion
depends on them?

## Validation

- Built for obelix@pvt (Pebble Time 2).
- Tested on real Pebble Time 2 hardware:
  - record and stop;
  - display the saved recording with its duration and size;
  - play audio through the speaker;
  - transcribe a stored recording;
  - delete recordings.
- Unit tests: new voice-recording playback tests, plus voice-endpoint tests covering the
  valid and malformed recording-management requests.

## Notes for reviewers

This PR covers recording, playback, deferred transcription, and the companion protocol as
one end-to-end feature. It is relatively large and can be split into separate reviewable
changes if preferred.

The capture stop path still shares the microphone tail-drop behavior described in #1423:
residual PDM data and the final partial Speex frame are not flushed.


julpel8:
@jplexer @gmarull My PR is heavy, should I split it in 3 PR (record, playing and transcribe) ?

ericmigi:
cool! keep in touch with @jplexer as they will help merge this if it's good enough

ericmigi:
would be cool to see an easy to use companion pkjs side that receives/decodes/streams audio from watch

julpel8:
Is this heavy battery consumption related to audio feature ?

image

ericmigi:
are you using a debug or release build? debug builds enable UART RX which is very high power consumption

julpel8:
> are you using a debug or release build? debug builds enable UART RX which is very high power consumption

Yes, I don't set CONFIG_RELEASE. Thank you !

eiis1000:
Would really love to see this merged! There are so many apps I want to make with this feature...

ericmigi:
we're tracking it! just haven't had time yet sorry

baibhavbista:
This is fantastic! offline on-device recording is going to be a really welcome addition.

However, I think there's a missing piece: `audio_recording_transcribe()` already transports the stored recording to the phone, but only exposes Pebble's transcript — there's currently no supported way for the initiating app (or its PKJS companion) to receive the recording itself. And a class of apps needs the audio, not just text.

Two reasons from my own use case (I'm building a Roam Research capture watchapp for PT2) that specifically need the recording, not just a transcript:
- **My own transcription** — the built-in model struggles with my accent, so I need to re-transcribe the audio on my own backend.
- **Storing the recording** — I want to save the recording in Roam alongside a timestamped transcript, so it's replayable and navigable later. In our equivalent apple watch app, this recording has proved to be a godsend in noisy situations where the AI misspells a lot of stuff.

Others in #719 have asked for the same (saving audio + transcript), so exposing an API to get the audio out would help a range of apps, not just mine.

One possible shape (drafted with AI assistance; the technical claims were checked against the PR source): a read API like `audio_recording_read(id, offset, buf, len)` (plus the codec params) that lets an app pull the stored Speex frames and forward them to a server. That said, see my follow-up below for a refinement I'd actually prefer, aligned with @ericmigi's earlier comment.

Happy to help in any way I can: whether testing or by contributing a stab at a follow-up PR (DCO + AI-disclosure understood). Thank you!

cc @ericmigi @jplexer to hear your thoughts on my request

baibhavbista:
Following up on my comment above, I just noticed @ericmigi's earlier note about "an easy to use companion pkjs side that receives/decodes/streams audio from watch." That's exactly the shape that would work for me, and it's a better fit than the read API I mentioned above. 

(Below is from Claude, so please take with grain of salt again)

---

`audio_recording_transcribe()` already streams the stored recording to the phone — if that same transport could hand the audio to the **initiating app's PKJS companion** (decoded, or as frames + codec metadata) instead of only the built-in transcriber, that would be all I need. PKJS could own the HTTP upload to my backend and return the result to the watch over AppMessage. An app-scoped export/stream that preserves the recording ID (`AudioRecordingInfo.id`) as a correlation ID, plus completion/error reporting, would be enough.

It also preserves this PR's offline model: keep the recording on-watch until the export is acknowledged, then the app can delete it.

(A byte-read API like `audio_recording_read()` would also work, but it'd make every app re-chunk the audio through AppMessage even though this PR already has a native transport — so I'd favor the companion stream if that matches where you're headed.)

---

Happy to help spec, test, or contribute a follow-up PR.

julpel8:
Thanks @baibhavbista!

Getting the raw audio out makes a lot of sense.

I’ve actually already forked [Brain Dump](https://github.com/adrienthiery/pebble-brain-dump-app) by @adrienthiery and got it listing the on-device recordings via audio_recording_list().
Adrien was interested in this feature  too. I’ll add the PKJS code to ship the audio out to a backend in my fork branch soon.

julpel8:
@baibhavbista After reviewing it, I realized I hadn’t pushed the changes that make it easier to list, edit and listen to the audio.
PS: I vibe write a doc in docs/reference/voice_recordings.md

jplexer:
please rebase, fix build, and clean up commits a little

meded90:
I want to create a system with notes that AI then transcribes. I really need this API.

julpel8:
@jplexer I rebase and clean some stuff

julpel8:
@ericmigi @baibhavbista I vibecoded a POC for the compagnon app you asked for. I'm on holydays so I give enough info to use and improve it:  

# Audio Recording PoC — `feature/audio-recording`

- Companion app:
  <https://github.com/julpel8/pebble-brain-dump-app/tree/feature/audio-recording>
- Firmware branch it runs against:
  <https://github.com/julpel8/PebbleOS/tree/audio/record-pkjs>

You can download fw and app here:
https://drive.google.com/drive/folders/10b5KBbNVmoiLYhsHXldbUWrlDZIgyjfm?usp=drive_link

## What it does

On-watch voice memos that don't need a phone to capture, with an optional copy to the phone for playback and export.

**On the watch (`src/c/main.c`).** DOWN from Home opens the recordings list. Row 0 starts a new capture (SELECT to start/stop, UP to discard); each stored recording opens a detail screen — UP transcribes, SELECT plays back / long-press deletes, DOWN exports to the phone.

**On the phone (`src/pkjs/`).** Metadata syncs on connect. Any recording without cached audio is auto-downloaded in the background, decoded to WAV, and cached in `localStorage`. The settings page lists them with inline playback, a save/share button, and Transcribe/Delete commands routed back to the watch.

## Wire protocol

Twelve message keys, phone→watch commands (`SYNC`, `EXPORT`, `TRANSCRIBE`, `DELETE`, `CANCEL_EXPORT`, `TRANSFER_ACK`) and watch→phone events (`LIST_BEGIN/ITEM/END`, `TRANSFER_BEGIN/CHUNK/END`, `STATUS`, `TRANSCRIPT`).

The transfer is **watch-driven**: chunks are pumped from `outbox_sent`, since AppMessage permits one message in flight and its ack already provides flow control. The phone acks nothing during a healthy transfer; it only sends `TRANSFER_ACK` to request a restart when it detects a gap, because the firmware's PCM decoder can only rewind to offset zero. `TRANSFER_END` carries the byte count actually decoded, which is authoritative over the duration-derived estimate announced up front.

## Format decision

Audio is stored on the watch as Speex (~3.6 kB/s) but exported as **decoded PCM wrapped in WAV**. Speex would be 9× smaller on the link, but iOS has no Speex decoder, so it could never be previewed in the settings WebView. The cost is volume, not complexity:

| duration | on watch flash | on the BT link | base64 in localStorage |
|---|---|---|---|
| 7 s | 25 kB | 224 kB | 298 kB |
| 2 min (max) | 432 kB | 3.84 MB | 5.1 MB |

Watch storage is not the constraint — 1 MB total budget, ~4 min 51 s across all recordings, 2 min cap per recording. The constraints are the link (32 kB/s, ~1 kB per AppMessage) and the phone's `localStorage` quota, which a 1.2 MB budget with oldest-first eviction protects.

`src/pkjs/recordings.js` also contains a complete, tested VRC2→Ogg/Speex converter. It is currently **unused** — kept for the day archival of long recordings matters.

## Firmware dependency

Requires PebbleOS `audio/record-pkjs` at SDK revision ≥ 112, including [`8bee5e6`](https://github.com/julpel8/PebbleOS/commit/8bee5e6bdb794b19840b6800391be7505c39a09c). That commit fixes the bug that made this PoC produce 32 ms of silence: the `read_pcm` syscall re-checked ownership on every call, and that check re-opened the very file the export stream held open, which PFS refuses with `E_BUSY`. First call succeeded, every subsequent one returned 0. The fix caches the owner UUID at stream start; it also repairs `play` and `delete`, which shared the same check.

## Verified vs. not

Confirmed on device: the transport carries bytes intact — watch-side and phone-side peak samples matched exactly. Confirmed locally: WAV container reads correctly under ffprobe (7.00 s, 16 kHz mono s16le), 43 tests pass, both firmware slots and the app build clean.

Not yet confirmed: a **full-length export after the firmware fix**. Everything up to now truncated at 1024 bytes.

## Known gaps

1. **The settings page is amputated.** `showConfiguration` opens only the recordings page; destinations, Google OAuth, Todoist, Nextcloud and webhook are unreachable, and `webviewclosed` no longer saves config. Deliberately left as-is; recoverable from `git show 0e27f21:src/pkjs/index.js`.
2. **The recording listing shares the `E_BUSY` hazard.** `prv_read_info` opens each file, so a recording is invisible to `audio_recording_list()` while its export stream is open. Mostly latent now that exports complete, but an aborted export would make the next one fail with "Phone unavailable" until the app restarts.
3. **No API releases a stuck export stream.** It clears on read-at-offset-zero, play, new capture, delete, or app exit. Playback is the practical escape hatch.
4. **Phone-initiated delete during playback fails**, for the same ownership reason — the on-watch delete path stops playback first, the phone path doesn't.
5. **Peak sample hit full scale (32767)** in the first 512 samples of every truncated export. Could be loud audio or a decoder transient at stream start; unresolved until a full stream can be listened to.

@jplexer @gmarull  PS: this PR is enough to handle audio with the watch

neobrain:
Personally I'd more interested in the "record live audio and process raw samples" use case that #719 was primarily about rather than the "record a snippet to flash storage in a size-optimized format" use case mentioned in passing, but both are important to cover of course.

This PR seems to start with a rather heavyhanded solution of implementing *only* the latter though - is there a benefit in doing so? Forcing a roundtrip through storage makes realtime processing impossible, whereas a simpler API exposing direct access to audio samples in a RAM buffer would serve both use cases. The logic to encode+save to storage would simply be provided by a helper API provided by the SDK *on top of* the OS interfaces that expose the audio samples themselves.

(Disclaimer: I'm not very familiar with the Pebble software stack yet, but I'm assuming currently neither the OS itself nor the SDK have microphone APIs)

julpel8:
My original issue was that taking voice notes was impossible without a phone connected over Bluetooth and therefore nearby. I mentioned #719 because it was related, but I did not have any specific implementation in mind beyond solving that problem.

-------- Original Message --------

jplexer:
> I'm not very familiar with the Pebble software stack yet, but I'm assuming currently neither the OS itself nor the SDK have microphone APIs

kinda? the firmware does have a sort of microphone api, but its only used for sending speex packets to the mobile app for transcription. apps do not have access to this
also nice to see you here @neobrain, loved your talk on FEX so much I saw it 2 times

baibhavbista:
ran into another situation on another app that I am building that would benefit a lot with this API, hoping we get it or alternatives soon 🤞 😁

PR #1650, opened 2026-07-02
doc/development: add logging for firmware developer

ck-telecom:


gmarull (changes requested):
`./pbl console`, not worth documenting this stuff, also because things are changing fast at this point

PR #1652, opened 2026-07-02
Switch ALS to LUX

jplexer:
Still think it needs some more in the field testing

### DIFF
--- a/src/fw/board/board_nrf5.h
+++ b/src/fw/board/board_nrf5.h
@@ -128,6 +128,11 @@ typedef struct {
   /////////////////////////////////////////////////////////////////////////////
   const uint32_t ambient_light_dark_threshold;
   const uint32_t ambient_k_delta_threshold;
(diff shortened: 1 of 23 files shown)


src/fw/board/board_nrf5.h
```diff
@@ -128,6 +128,11 @@ typedef struct {
   /////////////////////////////////////////////////////////////////////////////
   const uint32_t ambient_light_dark_threshold;
   const uint32_t ambient_k_delta_threshold;
+  // Raw-count -> lux conversion: lux = (level - offset) * num / den.
+  // den == 0 means no conversion available for this board.
+  const uint32_t ambient_light_lux_dark_offset;
+  const uint32_t ambient_light_lux_num;
+  const uint32_t ambient_light_lux_den;
```
gmarull, line 135: we should probably calibrate asterix (with improved sampling method) at some point

gmarull (approved):

PR #1654, opened 2026-07-02
power: make idle housekeeping wakeups event-driven

Coral-coder:
## Motivation

Follows up on the observation that the watch spends ~15 ms/sec outside deep
sleep at idle. A large share of that is timer-driven housekeeping: at idle the
CPU is woken about twice per second by mechanisms whose only job is periodic
bookkeeping (DST polling, cron list polling, watchdog liveness bits, the
kernel event-loop timeout, and the 1 Hz regular_timer heartbeat itself). This
series makes those wakeups event-driven, cutting idle timer wakes from ~2.1/s
to ~0.7/s with no observable behavior change.

## Changes

- **clock**: the DST watcher polled every second, but DST transitions and the
  top of the hour both land on minute boundaries — check once a minute from
  the regular_timer minutes list instead. The hourly chime match is widened
  from `== 0` to "first minute of the hour", which is equivalent at minute
  cadence and more robust to a delayed tick.

- **cron**: the service polled the sorted job list at 1 Hz. It already knows
  exactly when the next job is due, so arm a one-shot timer for that instant,
  re-armed on every list mutation (schedule / schedule_after / unschedule /
  clear), on firing, and on clock change. Job execution timing — including
  `offset_seconds` resolution — is unchanged. The armed interval is capped at
  one hour so tick-clock drift across long sleeps stays bounded.

- **kernel (liveness bits)**: the NewTimers, KernelBG-idle and PULSE-idle
  watchdog check-ins move from 1 s to 3 s. The hardware watchdog allows 10 s
  between feeds on SF32LB52 and 8 s on nRF52, so at least ~5 s of margin
  remains. Busy tasks still check in far more often as a side effect of doing
  work; only the idle cadence changes.

- **kernel (event loop)**: an idle KernelMain sleeps 3 s between watchdog
  check-ins instead of 1 s. Events wake the loop immediately regardless of
  the timeout.

- **regular_timer**: replace the fixed 1 Hz repeating timer with a one-shot
  armed for the soonest-due callback — the minimum multisecond countdown
  across the seconds list, or the next wall-clock minute boundary when the
  minutes list is non-empty. Elapsed whole seconds are credited from the
  monotonic tick clock (immune to wall-clock changes) and are arithmetically
  identical to the old one-decrement-per-second scheme for on-time wakes.
  Arming is wall-second aligned (+2 ms bias against tick rounding) so second
  and minute boundaries are hit exactly. This also removes the per-second
  `localtime_r()` call (timezone lookup plus a year-walk loop from the epoch)
  as a side effect. The `regular_timer_fire_seconds/minutes` test hooks are
  unchanged.

- **drivers/sf32lb52**: RC10K sleep-clock calibration backs off from every
  15 s to every 60 s after eight fast calibrations (~2 minutes of thermal
  settling). On-wrist temperature moves slowly and the HAL keeps a running
  average of the reference cycle, so the sleep-time math stays accurate while
  three blocking crystal-referenced measurements per minute are removed.

## Testing

- Full unit test suite passes (`./waf test`).
- `test_clock` expectations updated for the minute-cadence chime check;
  `test_cron` runs against the one-shot timer via the unchanged
  `cron_service_wakeup()` entry point.
- Needs on-device validation for deep-sleep residency numbers; happy to
  iterate on measurements.

Authored with Fable 5

### DIFF
--- a/src/fw/console/pulse2.c
+++ b/src/fw/console/pulse2.c
@@ -263,7 +263,8 @@ static void prv_pulse_task_main(void *unused) {
   static RegularTimerInfo idle_watchdog_timer = {
     .cb = prv_pulse_task_idle_timer_callback
   };
-  regular_timer_add_seconds_callback(&idle_watchdog_timer);
+  // 3s check-in, same margin rationale as the other task liveness feeds.
+  regular_timer_add_multisecond_callback(&idle_watchdog_timer, 3);
 
   CobsDecodeContext frame_decode_ctx;
   cobs_streaming_decode_start(&frame_decode_ctx, s_current_rx_frame,
--- a/src/fw/drivers/sf32lb52/rc10k.c
+++ b/src/fw/drivers/sf32lb52/rc10k.c
@@ -8,15 +8,32 @@
 #include "bf0_hal.h"
 
 #define RC10K_DEFAULT_FREQ_HZ 10000UL
-#define RC10K_CAL_PERIOD_MS 15000U
+
+// Calibrate frequently right after boot while the oscillator settles
+// thermally, then back off: on-wrist temperature moves slowly and the HAL
+// keeps a running average, so a stale-by-a-minute calibration stays well
+// within the accuracy the sleep-time math needs.
+#define RC10K_CAL_PERIOD_FAST_MS 15000U
+#define RC10K_CAL_PERIOD_SLOW_MS 60000U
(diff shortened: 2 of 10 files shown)


gmarull:
please, split into multiple PRs, some changes require some extra verification

Coral-coder:
Seems to be an issue with diag info

PR #1656, opened 2026-07-03
fw/apps/prf: use snake grid on round displays in mfg touch test

gmarull:
Replace the concentric-ring pattern with the same boustrophedon snake grid used on rectangular displays. Each row spans the chord of the circle at its height (inset from the edge where touch is unreliable), so rows near the top and bottom hold fewer columns and only the middle runs full width. Walls between rows span their boundary chord, with a gap around the row-to-row connector.

This unifies the guide path and dot drawing code, since the round path is now also a polyline.


gmarull:
merging need for testing

PR #1659, opened 2026-07-03
⚡️cron: arm a one-shot timer for the next job instead of polling at 1 Hz

Coral-coder:

The cron service registered a per-second regular_timer callback that locked the job-list mutex and compared the head job's execute time against the clock 86400 times a day, almost always concluding there was nothing to do. Cron already knows exactly when the next job is due, so arm a one-shot new_timer for that instant instead and re-arm on every list mutation (schedule, schedule_after, unschedule, clear) and firing.

Job execution semantics are unchanged: callbacks still run on the timer task at the same second they used to, offset_seconds resolution is preserved, and cron_service_handle_clock_change() still runs due jobs and now re-arms for the new head. The armed interval is capped at one hour so any tick-clock drift across long sleeps stays bounded.

(cherry picked from commit 90a67da23149275b8cf746706e04744e321f6c7c)

### DIFF
--- a/src/fw/services/cron/service.c
+++ b/src/fw/services/cron/service.c
@@ -4,10 +4,11 @@
 #include "pbl/services/cron.h"
 #include <pebbleos/cron.h>
 
+#include "drivers/rtc.h"
 #include "pbl/os/mutex.h"
-#include "system/passert.h"
-#include "pbl/services/regular_timer.h"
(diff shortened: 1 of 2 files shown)


jplexer (approved):
(fixed an edge case)

gmarull (approved):

PR #1660, opened 2026-07-03
⚡️kernel: feed task-liveness watchdog every 3 s instead of every second

Coral-coder:

## What

Two related cadence changes, one per commit:

1. **Task-liveness watchdog feeds: 1 s → 3 s.** The hardware watchdog timeout is 8–10 s depending on platform; feeding the liveness bits every second is 3× more often than needed. At 3 s cadence a genuinely-hung task is still detected with ≥5 s of margin before the hardware timeout.
2. **KernelMain idle event timeout: 1 s → 3 s.** KernelMain's event loop woke every second just to check in with the watchdog; with feeds at 3 s it only needs to wake at the same cadence.

## Why

On an otherwise-idle watch these two are a steady 2 wakeups/second that serve no purpose beyond watchdog bookkeeping. This cuts them to ~0.7/s combined.

## Testing

Full unit suite passes; field-tested on SF32LB52 (Obelix) alongside the other idle-wakeup changes — no watchdog resets over multi-day soak, hang detection verified still working via a deliberately-stalled task.

### DIFF
--- a/src/fw/kernel/event_loop.c
+++ b/src/fw/kernel/event_loop.c
@@ -593,7 +593,10 @@ void launcher_main_loop(void) {
 
     // We make this PebbleEvent static to save stack space
     static PebbleEvent e;
-    if (event_take_timeout(&e, 1000)) {
+    // The timeout exists only to refresh the watchdog bit above; events wake
+    // the loop immediately. 3s keeps ~5s of margin against the shortest (8s)
+    // hardware watchdog while letting an idle KernelMain sleep 3x longer.
+    if (event_take_timeout(&e, 3000)) {
       const PebbleTaskBitset kernel_main_task_bit = (1 << PebbleTask_KernelMain);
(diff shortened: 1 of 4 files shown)


src/fw/kernel/event_loop.c
```diff
@@ -593,7 +593,10 @@ void launcher_main_loop(void) {
 
     // We make this PebbleEvent static to save stack space
     static PebbleEvent e;
-    if (event_take_timeout(&e, 1000)) {
+    // The timeout exists only to refresh the watchdog bit above; events wake
+    // the loop immediately. 3s keeps ~5s of margin against the shortest (8s)
+    // hardware watchdog while letting an idle KernelMain sleep 3x longer.
+    if (event_take_timeout(&e, 3000)) {
```
gmarull, line 599: let's not touch this (yet), task watchdog needs a bit of love first.

Coral-coder: It very well did, but it seems to work fine these past 2 weeks.

PR #1668, opened 2026-07-06
alarms: preview sounds on highlight

OPerepadia:
Currently, there is no way to preview alarm sounds without actually setting an alarm and waiting for it to trigger.

This PR adds sound previews to the alarm sound picker. When the user opens the picker and moves between alarm tones, the currently highlighted tone plays as a preview.

Tested on PT2.


gmarull (changes requested):
please, split changes following commit guidelines (applib, app, popups)

gmarull (approved):

PR #1682, opened 2026-07-09
⚡️ power: reduce idle draw from ALS, touch, and stationary polling

Coral-coder:
Three small idle-power reductions:

- ALS: extend the cached ambient-light level TTL from 1 s to 5 s while the backlight is off, skipping the ~200 ms I2C poll on rapid wake-check sequences.
- CST816 touch: the wedge-detection watchdog only needs to catch a controller that stopped asserting INT; check every 60 min instead of every 30 min.
- Stationary service: poll for the stationary transition every 2 min instead of every minute; entry into stationary mode is already delayed well beyond that.

(cherry picked from commit 8a32da38e670d0fb39081009fd1230be04c5032b)

### DIFF
--- a/src/fw/drivers/touch/cst816/cst816.c
+++ b/src/fw/drivers/touch/cst816/cst816.c
@@ -61,7 +61,7 @@ PBL_LOG_MODULE_DEFINE(driver_touch_cst816, CONFIG_DRIVER_TOUCH_LOG_LEVEL);
 
 /* Workaround: the CST816 occasionally wedges and stops asserting its INT line.
  * If no touch activity is seen between two watchdog checks, hard-reset it. */
-#define CST816_WATCHDOG_PERIOD_MIN    30
+#define CST816_WATCHDOG_PERIOD_MIN    60
 
 /* The chip stays awake for 2s after a wake; an interrupt seen >=2s after the
  * previous one therefore marks a fresh sleep->awake transition. */
--- a/src/fw/services/light/service.c
+++ b/src/fw/services/light/service.c
@@ -119,7 +119,7 @@ static uint32_t s_total_on_time_ms; // Total backlight on time tracked internall
 //! that follows the press within the TTL) skip the ~200 ms I2C poll.
 static uint32_t s_als_cached_level;
 static RtcTicks s_als_cached_ticks;  // 0 = invalid
-#define ALS_CACHE_TTL_TICKS (RTC_TICKS_HZ)  // 1 second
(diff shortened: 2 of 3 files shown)


src/fw/drivers/touch/cst816/cst816.c
```diff
@@ -61,7 +61,7 @@ PBL_LOG_MODULE_DEFINE(driver_touch_cst816, CONFIG_DRIVER_TOUCH_LOG_LEVEL);
 
 /* Workaround: the CST816 occasionally wedges and stops asserting its INT line.
  * If no touch activity is seen between two watchdog checks, hard-reset it. */
-#define CST816_WATCHDOG_PERIOD_MIN    30
+#define CST816_WATCHDOG_PERIOD_MIN    60
```
gmarull, line 64: the net effect of this is probably ~0, so not worth changing

src/fw/services/light/service.c
```diff
@@ -119,7 +119,7 @@ static uint32_t s_total_on_time_ms; // Total backlight on time tracked internall
 //! that follows the press within the TTL) skip the ~200 ms I2C poll.
 static uint32_t s_als_cached_level;
 static RtcTicks s_als_cached_ticks;  // 0 = invalid
-#define ALS_CACHE_TTL_TICKS (RTC_TICKS_HZ)  // 1 second
+#define ALS_CACHE_TTL_OFF_TICKS (5 * RTC_TICKS_HZ)
```
gmarull, line 122: @jplexer pls check

src/fw/services/stationary/service.c
```diff
@@ -245,7 +245,7 @@ static void prv_exit_disabled_state(void) {
 #if DEBUG_STATIONARY
   regular_timer_add_seconds_callback(&s_accel_stationary_timer_info);
 #else
-  regular_timer_add_minutes_callback(&s_accel_stationary_timer_info);
+  regular_timer_add_multiminute_callback(&s_accel_stationary_timer_info, 2);
```
gmarull, line 248: same, net effect is likely negligible, not worth changing

src/fw/services/light/service.c
jplexer: no effect, since we sample whenever we need to this is basically only "we reactvated so fast, dont check again"

gmarull:
Thanks for this PR, but as explained, it likely provides no gains. Happy to re-evaluate if, e.g., PPK2 measurements are provided to counter our arguments.

PR #1686, opened 2026-07-13
resource: add zh_CN as built-in lang

ck-telecom:


gmarull:
what is the resource delta?

PR #1688, opened 2026-07-13
Fix blank device name on PRF QR screen

gmarull:
## Summary

The device/mobile name field on the PRF getting-started (QR) screen could render blank, or show a stale string, while a comm session was up. Two independent holes caused this:

- `gap_le_connection_copy_device_name()` leaves the output buffer untouched when the remote device name has not been fetched yet (the GATT read of GAP Device Name 0x2A00 is asynchronous), so the connected code path could draw an empty/stale buffer.
- `bt_driver_store_device_name_kernelbg_cb` only emitted `PEBBLE_BLE_DEVICE_NAME_UPDATED_EVENT` when the connection had a valid bonding and the name was persisted. In PRF a comm session can exist before a bonding is finalized, so the freshly read name was never announced and the screen stayed blank until an unrelated BT connection event repainted it.

## Changes

- `comm/ble`: always emit the device-name-updated event once a name has been read; persist to storage only when a bonding exists. Also bail out early if the connection went away before the background callback ran, instead of copying from an uninitialized name buffer.
- `fw/apps/prf/recovery_first_use`: clear the name buffer before copying and fall back to the local device name when the remote name is still unknown, matching the disconnected code path.

Net behavior: watch name → (phone connects, name read in flight) still watch name → name-updated event → phone name. No blank window at any point.

## Testing

- Full PRF build for `obelix@pvt` (nimble) compiles clean.

### DIFF
--- a/src/fw/apps/prf/recovery_first_use/recovery_first_use.c
+++ b/src/fw/apps/prf/recovery_first_use/recovery_first_use.c
@@ -155,8 +155,14 @@ static void prv_update_name_text(RecoveryFUAppData *data) {
   } else if ((comm_session_get_system_session() != NULL) && (gap_conn != NULL)) {
     // If we have connected to a device and we have a connection to the mobile app, show the device
     // name (we are required to have a connection to mobile app to get the name).
+    data->name_text_buffer[0] = '\0';
     gap_le_connection_copy_device_name(gap_conn, data->name_text_buffer,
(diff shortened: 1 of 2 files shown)


jplexer (approved):

gmarull:
Superseded by #1689: the blank name field turned out to be caused by the device name read running concurrently with GATT discovery (responses cross-delivered), fixed at the root there.

PR #1697, opened 2026-07-13
apps/music: fix layout on large round displays and add screenshot tests

gmarull:
## Summary

On Getafix (260x260 round), the music app used the "large" size config that was only ever tuned for rectangular Emery-class displays:

- The cassette/pause status icon was anchored to the bottom-right corner and fell almost entirely outside the visible circle (FIRM-3278: "Pause icon cut off by screen edge").
- The first line of the track title was clipped by the circle's left edge.
- The "start playback on your phone" window artwork sat off-center.

This adds round-specific values to the large config (wider horizontal margin, fields shifted down so each lies within its chord — which also pulls the status icon inside the circle — and centered no-music artwork). Rectangular platforms get bit-identical values.

| Before (qemu_gabbro) | After (qemu_gabbro) |
| --- | --- |
| title clipped by circle, pause/cassette icon mostly outside the visible area | all content inside the circle |

Also adds screenshot tests for the music app (playing, paused, long metadata, no progress reporting, no music) with baselines for asterix (flint), obelix (emery) and gabbro, so future layout changes to this app are covered on all three platforms.

Fixes FIRM-3278

## Testing

- Verified in QEMU (qemu_gabbro) by injecting now-playing metadata: playing, paused, volume-overlay and no-music states all render inside the circle; long artist names ellipsize instead of clipping.
- `./pbl test`: full suite passes, including the new `test_music` binaries on all three platforms.


jplexer (approved):
Peak music

PR #1704, opened 2026-07-14
fw/drivers/imu: return fresh data from accel_peek while sampling

jplexer:
## Summary

Legacy tilt apps (Gyro Maze, Mr. Runner - UP, Fluid Sim, Level, ...) poll `accel_service_peek()` instead of subscribing with a batch size. While sampling is active, both native drivers returned the sample cached at the last FIFO drain from `accel_peek()`. The FIFO watermark is set by the most demanding real subscriber — with only the activity tracker subscribed (25 Hz × 125 samples on LSM6DSO) the cached sample can be up to **5 seconds** old, and peek-mode subscribers (`samples_per_update=0`) don't lower the watermark. Classic-era drivers (bmi160) read the current output registers on every peek, which is why these apps were responsive on older watches.

This is the deterministic half of FIRM-3385 (multi-second tilt lag in games); the intermittent half (stalled INT1 stream) is covered by the recent self-heal series (#1701 lineage), which is preserved here.

## Changes

- **lsm6dso**: on peek during active sampling, read the current sample from `OUTX_L_A` — the output registers update at ODR independently of the FIFO on this part, so the read has no side effects. Fall back to the cached sample if the read fails. Staleness drops from ≤5 s to ≤1 ODR period.
- **lis2dw12**: the FIFO is read *through* the output registers on this part, so a direct read would steal a sample from the subscriber stream. Drain the FIFO instead: queued samples are dispatched to subscribers, the cached sample is refreshed (≤1 ODR period old), and the stall-watchdog tick is updated.

Concurrency: peek and all driver work run under the recursive accel-manager mutex, so peek-drain cannot race INT1 servicing; if a peek-drain empties the FIFO ahead of a queued INT1 work item, the INT1 pad re-check returns harmlessly.

## Testing

- Built `obelix@pvt` and `getafix@dvt2` (normal variant), both clean.
- App-side configs verified from source/disassembly: laggy apps are all peek-pollers; Squid Jump (reported working) subscribes with 1 sample @ 10 Hz.

Fixes FIRM-3385

### DIFF
--- a/src/fw/drivers/imu/lsm6dso/lsm6dso.c
+++ b/src/fw/drivers/imu/lsm6dso/lsm6dso.c
@@ -261,13 +261,6 @@ static void prv_lsm6dso_process_samples(uint16_t num_samples, uint64_t timestamp
   };
 
   accel_cb_new_samples(&batch);
-
-  // Keep the most recent sample around for accel_peek().
-  uint8_t *last = &LSM6DSO->state->raw_sample_buf[(num_samples - 1) * LSM6DSO_FIFO_WORD_SIZE_BYTES];
-  prv_raw_to_mg(&last[1], &LSM6DSO->state->last_sample);
-  LSM6DSO->state->last_sample.timestamp_us =
-      timestamp_us + (num_samples - 1) * LSM6DSO->state->sampling_interval_us;
-  LSM6DSO->state->last_sample_valid = true;
 }
 
 static uint8_t prv_get_bdr(uint32_t sampling_interval_us) {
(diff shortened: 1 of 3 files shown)


src/fw/drivers/imu/lsm6dso/lsm6dso.c
```diff
@@ -862,6 +864,15 @@ int accel_peek(AccelDriverSample *data) {
...
     }
 
+    if (prv_lsm6dso_read(LSM6DSO_OUTX_L_A, raw, sizeof(raw))) {
+      prv_raw_to_mg(raw, &LSM6DSO->state->last_sample);
+      LSM6DSO->state->last_sample.timestamp_us = prv_get_curr_system_time_us();
+      LSM6DSO->state->last_sample_valid = true;
+    } else {
+      // Fall back to the last drained FIFO sample
+      PBL_LOG_ERR("Failed to read sample");
+    }
```
gmarull: feels like bad error handling: if we fail, we fail, otherwise we change the behavior.

gmarull (commented):
one comment, also needs verification in hw if not already done

gmarull (approved):

PR #1712, opened 2026-07-15
vibes: harden AWINIC stop path and extend Vibe Log Info coverage

jplexer:
## Context

FIRM-3361 (recurrence of the FIRM-1538 / FIRM-2319 "notification vibrates endlessly until dismissed" family, now seen on getafix/Android as well as obelix/iOS).

Root-cause analysis: the notification vibe path has no software repeat — only `phone_ui` and `alarm_popup` use repeat timers. On the AWINIC haptic drivers, `vibe_ctl(true)` starts an open-ended hardware drive (AW86225: RAM loop mode with SEQ1 loop count 0xF = infinite loop; AW8623X: CONT mode with max drive times), so every moment of silence depends on a timed STOP sequence over I2C. The AW86225 stop path discarded every I2C return value: one transient I2C failure while stopping leaves the motor buzzing indefinitely with nothing in the logs, until the next unconditional `vibe_ctl(false)` — which is exactly what dismissing the notification does (`vibes_cancel`). This matches the reported symptom precisely.

## Changes

- **aw8623x**: fix uninitialized `ret &=` in `vibe_set_strength` — a failed CONTCFG6 write could never trip the assert.
- **aw86225**: `prv_aw862xx_play_go()` now reports success; the stop path verifies the chip reached standby (including re-checking after the forced-standby fallback, which previously fired blind), retries once, and logs an error if the motor cannot be confirmed stopped.
- **vibe_pattern**: the Vibe Log Info toggle now also elevates pattern-completion and pattern-clear logs (with live strength and dropped-step count at clear time). A capture showing `trigger_start` with no `pattern complete`, then a clear reporting nonzero strength, confirms the wedge; repeated `trigger_start` lines would instead indicate a software re-trigger.

## Verification

- `getafix@dvt` (AW8623X) and `getafix@dvt2` (AW86225) build clean
- 4/4 vibe unit test suites pass (`test_vibe`, `test_vibe_score`, `test_vibe_score_info`, `test_vibe_intensity`)
- PRF variant compiles (link failure on main is pre-existing and unrelated)
- Caveat: QEMU stubs the vibe driver, so the AW86225 stop-verify behavior needs a bench check on real getafix/obelix hardware.

FIRM-3361

### DIFF
--- a/src/fw/drivers/vibe/vibe_aw86225.c
+++ b/src/fw/drivers/vibe/vibe_aw86225.c
@@ -143,40 +143,54 @@ bool prv_modify_reg(uint8_t reg_addr, uint32_t mask, uint8_t reg_data)
 	return prv_write_register(reg_addr, reg_val);
 }
 
-static void prv_aw862xx_play_go(bool flag)
+//! Start (flag=true) or stop (flag=false) playback. Returns true when the
+//! command was written successfully; for stop, additionally requires the chip
+//! to have reached standby. The RAM waveform is configured as an infinite
+//! hardware loop, so a stop that silently fails leaves the motor running.
+static bool prv_aw862xx_play_go(bool flag)
 {
-	uint8_t val;
+  uint8_t val;
 
-	if (flag) {
-		val = AW862XX_BIT_PLAYCFG4_GO_ON;
-		prv_write_register(AW862XX_REG_PLAYCFG4, val);
-	} else {
-		bool standby = false;
-		prv_modify_reg(AW862XX_REG_SYSCTRL1, AW862XX_SYSCTRL1_RAMINIT_MASK,
-		               AW862XX_SYSCTRL1_RAMINIT_ON);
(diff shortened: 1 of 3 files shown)


gmarull (commented):
will check more in detail tomorrow

src/fw/drivers/vibe/vibe_aw86225.c
```diff
@@ -90,6 +90,7 @@ PBL_LOG_MODULE_DEFINE(driver_vibe_aw86225, CONFIG_DRIVER_VIBE_LOG_LEVEL);
 #define AW862XX_PWR_OFF_TIME                            (2) /* ms */
 #define AW862XX_PWR_ON_TIME                             (8) /* ms */
 #define AW862XX_STOP_STANDBY_RETRIES                    (40)
+#define AW862XX_STOP_ATTEMPTS                           (2)
```
gmarull: Unless we are 100% sure this happens, we should not add a retry (yet). I'd for now log and see if this is observed.

gmarull: Did logs contain e.g. an i2c xfer error?

PR #1716, opened 2026-07-15
fw/drivers/rtc: bound RSF sync wait in rtc_get_time_ms

jplexer:
The wait for RTC shadow-register synchronization spun forever if RSF never set. On FIRM-3403 a wedged low-power domain left RSF stuck low, so rtc_get_time_ms hung the system task and tripped the task watchdog ("stuck task" reboot 9 s after boot on v4.23.0).

Bound the wait at 1M iterations (~40-200 ms depending on HCLK, vs ~0.5 ms genuine sync time) and fall back to the last-synced shadow values, degrading to stale time instead of a watchdog reboot loop. The timeout is logged once; the guard flag is set before logging because PBL_LOG timestamps re-enter rtc_get_time_ms.

Related to FIRM-3403.

### DIFF
--- a/src/fw/drivers/rtc/sf32lb.c
+++ b/src/fw/drivers/rtc/sf32lb.c
@@ -283,13 +283,28 @@ void rtc_set_time(time_t time) {
   prv_rtc_set_time_no_cal_reset(time);
 }
 
+// RSF sync completes within a few RTC clock cycles (~0.5 ms at RC10K); a
+// wedged low-power domain can leave it stuck low forever, so bound the wait
+// well above the genuine sync time (~40-200 ms depending on HCLK).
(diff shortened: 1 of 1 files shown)


gmarull:
@rabbitsaviola please check

gmarull:
It may be worth panicking if this happens, because a bad RTC can then cause all sorts of random bugs that are very difficult to trace. But let's wait for @rabbitsaviola feedback.

rabbitsaviola:
> It may be worth panicking if this happens, because a bad RTC can then cause all sorts of random bugs that are very difficult to trace. But let's wait for @rabbitsaviola feedback.

Agree. RSF timeout is unexpected. It's better to find the root cause.

PR #1719, opened 2026-07-16
fw/shell/normal: show a dialog when a quick launch app is not installed

ericmigi:
Quick launch preferences store the app UUID. When the phone offloads an app to stay under the locker sync limit, it deletes the app's BlobDB record from the watch, so the UUID no longer resolves to an install id. All quick launch paths treated this as "nothing assigned": a long press opened the Quick Launch setup menu with the shortcut showing as unassigned, a single click asserted in app_manager_put_launch_app_event() and rebooted the watch, and combos silently did nothing.

An enabled quick launch preference always has a valid UUID (enforced by prefs normalization), so a failed lookup on an enabled button means the app is missing from the watch, not unassigned. Show a dialog telling the user to re-add the app from the phone instead of pretending the button is unassigned.

Fixes FIRM-3347
Fixes FIRM-2825

### DIFF
--- a/src/fw/shell/normal/watchface.c
+++ b/src/fw/shell/normal/watchface.c
@@ -21,6 +21,10 @@
 #include "applib/app_timer.h"
 #include "applib/app_launch_reason.h"
 #include "applib/ui/click_internal.h"
+#include "applib/ui/dialogs/dialog.h"
+#include "applib/ui/dialogs/expandable_dialog.h"
+#include "pbl/services/i18n/i18n.h"
+#include "resource/resource_ids.auto.h"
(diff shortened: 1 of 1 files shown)


jplexer:
mhh the correct way would be for a quick launch app to not get offloaded (cc: @sjp4 )

sjp4:
> mhh the correct way would be for a quick launch app to not get offloaded (cc: @sjp4 )

i'm guessing this is most likely to happen when the user uninstalled the app. The hitting-the-locker-sync-limit case is very unlikely (unless you are eric)

ericmigi:
the dialog is still bad and could be updated easily with this PR

PR #1721, opened 2026-07-16
ci: rebase PR branches onto target branch before running

gmarull:
## Summary

- Check out the PR head SHA and rebase it onto the current tip of the target branch before any CI job runs, following the approach used by Zephyr. If the rebase does not apply cleanly, CI fails immediately with an error asking the author to rebase locally.
- This replaces building GitHub's ephemeral merge commit, which only reflects the target branch at event time, so stale branches were tested against an outdated base (and conflicting PRs failed at checkout with a cryptic "unable to resolve refs/pull/N/merge" error).
- The rebase logic lives in a reusable composite action (`.github/actions/rebase`) with a checkout-style `submodules` input (default `false`, accepts `true`/`recursive`) to re-sync submodules after the rebase so gitlinks match the rebased tree.
- Compliance diff ranges now use `origin/$GITHUB_BASE_REF` instead of the event base SHA, which would include unrelated upstream commits after the rebase.

## Test plan

- YAML validated locally.
- This PR itself exercises the new path: all PR workflows should show the "Rebase onto target branch" step rebasing onto `main` before building/testing.

### DIFF
--- a/.github/actions/rebase/action.yml
+++ b/.github/actions/rebase/action.yml
@@ -1,0 +1,46 @@
+name: Rebase onto target branch
+description: >
+  Rebase the checked-out PR branch onto the current target branch so CI
+  runs against up-to-date code. Fails if the rebase does not apply
+  cleanly. No-op on non-PR events. Requires a full checkout
+  (fetch-depth: 0) of the PR head SHA.
+
+inputs:
+  submodules:
+    description: >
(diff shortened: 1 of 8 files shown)


gmarull:
Note: this will also allow us to re-run CI on old PRs without a manual rebase and make sure everything's still green

jplexer (approved):
cool!

PR #1742, opened 2026-07-20
fw/services/activity: keep short sleep cycles inside a fragmented night

ericmigi:
Fixes FIRM-3473

## What changed

The Kraepelin sleep detector discarded any sleep cycle shorter than 60 minutes, so brief mid-night restlessness that split a night into fragments made genuinely-slept 20–59 min chunks display as "awake". Short cycles were doubly doomed: the running avg-VMC / active-pct checks include the trailing awake run, and `vmc_clip` can't stop that movement from dominating a short cycle's averages (14 clipped minutes alone push a 45-min cycle over the threshold), so they usually died before the length check even ran.

Now a short cycle is kept when it is **sandwiched inside a night**:

- Average-based rejections are deferred for a cycle that starts within 30 min of the previous cycle's end while under 60 min.
- At end-of-cycle validation, those checks re-run over the slept portion only (trailing awake run excluded). A cycle that slept ≥20 min and passes is **held back** as a pending cycle (with its deep-sleep sessions).
- The pending cycle registers only once a following cycle is kept and starts within the same 30-min gap; otherwise it is dropped. Requiring sleep on both sides keeps post-wake stillness (dozing / reading in bed) from extending the night.

## Why

Deep dive on FIRM-3473 (getafix, fw v4.27.0): overnight the firmware recorded two sleep sessions with a 71-min "awake" gap covering the disputed 3:31 AM. The logs show the algorithm *detected* sleep in that gap — including a 37-min deep-sleep run — then rejected the cycle (`Session to delete not found` at 10:59:01Z is the rejection's only trace; that warning fires exactly when a never-registered <60-min cycle is discarded). Full analysis in the Linear issue.

## How I validated

- `./waf test -M ".*kraepelin.*"` — all captured reference nights pass **unchanged** except `sleep_4`, which gains a 23-min interior cycle the old code discarded (the exact reported bug shape, on real data). The post-wake phantom-sleep guard nights (`sleep_7`, `sleep_8`, `pbl_27921`, `pbl_37734`, `watch_on_table*`) are all bit-identical to baseline.
- 5 new synthetic-night tests: mid-night short cycle kept; nightstand (flat/still) cycle still vetoed; far-gap cycle still rejected; tiny (<20 min) cycle still rejected; chained fragments kept (last fragment conservatively dropped).
- All 12 activity suites + health/sleep/alarm suites pass.
- Built and booted `qemu_emery`; firmware boots to launcher normally.
- Surveyed ~70 sleep-tracking Linear issues before designing: phantom-sleep complaints (FIRM-1533, FIRM-2500, FIRM-1121) are as common as missed-sleep ones — hence the sandwich requirement instead of a naive length relaxation, which measurably regressed labeled reference nights (+33 to +81 min phantom sleep) when tried.

## Known limitations (deliberate)

- The last short fragment of a broken night is dropped (nothing follows to prove it was sleep) — indistinguishable from post-wake dozing without HRM.
- The pending cycle registers when its follower reaches 60-min ongoing registration; a follower rejected *after* that point leaves the pending cycle registered. Undoing it would require deleting a non-ongoing session (asserts in `activity_sessions.c`) or breaking chronological session order.
- The pre-existing too-short gate still uses elapsed minutes (including the awake tail), so a ~50-min-slept cycle with an 11-min tail passes as a full cycle, as before. Left untouched for scope/regression safety.

> **Emulator-only validation.** Not yet run on hardware; the algorithm is platform-independent C exercised heavily by the host test suite, but a night of real-wrist data on getafix/obelix before release would be prudent.

### DIFF
--- a/src/fw/services/activity/kraepelin/kraepelin_algorithm.c
+++ b/src/fw/services/activity/kraepelin/kraepelin_algorithm.c
@@ -124,6 +124,8 @@ typedef struct {
 
 // ----------------------------------------------------------------------------------------
 // Sleep detection structures
+// Max number of deep sleep sessions per sleep session
+#define KALG_MAX_DEEP_SLEEP_SESSIONS 8
 // The data for each minute that we use for computing sleep
 typedef struct {
   uint16_t vmc;
@@ -138,6 +140,12 @@ typedef struct {
   uint32_t vmc_sum;
   uint16_t consecutive_sleep_minutes;
   uint16_t consecutive_awake_minutes;
+  // Contribution of the current run of awake minutes to the totals above. Reset when
+  // sleep resumes, so (total - tail) gives the slept portion of the cycle, excluding the
+  // movement that ends it -- vmc_clip alone cannot keep that movement from dominating the
+  // averages of a short cycle.
+  uint16_t num_non_zero_tail;
+  uint32_t vmc_sum_tail;
 } KAlgSleepActivityStats;
 
 typedef struct {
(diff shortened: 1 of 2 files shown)


ericmigi:
@jplexer speculative fix - i'm not really sure how to test this?

PR #1753, opened 2026-07-20
docs: promote PULSEv2 protocol docs, drop remaining legacy files

rhaist:
`docs/legacy/` was created in 3d2ff1fc as a staging area — *"Content may be eventually be brought up to the upcoming docs platform."* This PR finishes that migration.

**Promoted** (verified against the current tree): the four PULSEv2 pages → `docs/reference/pulse2/` — the protocol specs match the live implementation (`src/fw/console/pulse*`, `tools/pulse/`).

**Deleted**:
- `porting-bluetooth.md` — per review, the bt_driver layer it documents is slated for removal.
- `battery_state.dot` — describes a ten-state charge state machine; current `battery_state.c` has three connection states.
- `legacy_flash_imaging.txt` — self-declared obsolete, superseded by PULSE flash imaging.
- `docstring-help.md` — every referenced path moved to `sdk/docs/` years ago.
- `tasks.txt` — predates the `PebbleTask_*` naming and cites removed APIs.

With the directory gone, `legacy` is dropped from the Sphinx `exclude_patterns`. Sphinx builds clean with `-W -n` (the RTD strict settings).


docs/development/porting_bluetooth.md
```diff
@@ -0,0 +1,181 @@
+# Porting a new Bluetooth stack into PebbleOS
```
gmarull: not worth bringing this up, I have in the agenda a full removal of all that layer anyway

rhaist: Alright. I'll drop the whole file for now.

docs/index.md
```diff
@@ -98,6 +99,10 @@ boards/index.md
 
 ```{toctree}
 :hidden:
-:caption: 📖 Reference 
+:caption: 📖 Reference
 reference/external.md
+reference/pulse2/pulse2.md
+reference/pulse2/reliable-transport.md
+reference/pulse2/flash-imaging.md
+reference/pulse2/history.md
```
gmarull: maybe create a hierarchy here, so top level toctree includes eg reference/index, then reference index pulse2/* etc.

rhaist: Makes sense. Will also add some explanations to make it a little more user and machine friendly :)

gmarull (approved):

PR #1774, opened 2026-07-22
docs: publish Doxygen API reference via Read the Docs

rhaist:
Implements the proposal from #1751 (standalone Doxygen site on Read the Docs, no breathe), plus some docs-chain modernization picked up along the way.

## What's in here

- **Publish the Doxygen API reference**: RTD installs doxygen via apt, clones doxygen-awesome-css (pinned by commit) and runs `doxygen Doxyfile.rtd` in `pre_build`; the output ships verbatim under `/apidoc/` via `html_extra_path` and is linked from the Reference page. The PR skip-guard now also triggers on source changes so previews never show a stale API reference.
- **Doxygen input covers the new tree layout**: `include/` (public pbl headers), `lib/`, and `subsys/` in addition to `src/` — the reference would otherwise go blind as code migrates out of `src/`.
- **A `\mainpage`** so the entry page isn't blank, with starting points and a source-tree overview.
- **Tagfile** at `/apidoc/pebbleos.tag` for machine consumers.
- **Doxyfile regenerated** with `doxygen -u` (1.8.1.2-era template → current; effective config verified byte-identical via `-x_noenv` diff).
- **Sphinx toolchain bumped**: sphinx 9.1, myst-parser 5.1, book-theme/design floors raised; RTD image now `ubuntu-lts-latest`; unused sphinx-quickstart scaffolding (docs/Makefile, make.bat) dropped.

## Notes

- No graphviz: the base Doxyfile has `HAVE_DOT=NO`, so it would be dead weight.
- Output is ~110 MB / ~2900 files with the source browser off — server-side storage only; individual pages are a few KB over the wire.
- Verified locally: doxygen + strict `sphinx-build -W` pass on the new toolchain; the RTD preview on this PR should show the real thing.

## Possible follow-ups (separate PRs)

- Burn down the ~330 Doxygen warnings (mostly 2012-era `@param` mismatches) in area-sized chunks.
- Curated `@defgroup` topics over applib/services for a browsable Topics index instead of flat file lists.
- Cross-link guide pages to specific API reference pages.

Fixes #1751

PR #1775, opened 2026-07-22
doxygen: burn down all documentation warnings (758 -> 0)

rhaist:
Burns down all Doxygen warnings in one sweep: **758 → 0** (measured over `include`, `lib`, `src`, `subsys` — the input set the API reference uses since #1774 landed). This is the first follow-up teased in #1774; rebased onto main after that merge.

## What's in here

All changes are comment-only, split per area (`include:`, `applib:`, `fw:`, `lib/util:`), plus two config lines:

- `EXAMPLE_PATH = sdk/docs/external_refs` — the example HTML files that several applib headers `\include` have existed there all along but were never wired up, so the snippets were silently dropped from every build.
- `WARN_AS_ERROR = FAIL_ON_WARNINGS` in `Doxyfile.rtd` (per review) — the RTD PR build now fails when new documentation warnings are introduced, locking in the zero. `FAIL_ON_WARNINGS` reports all warnings before exiting nonzero, unlike `YES` which aborts on the first. Scoped to the RTD config so local and legacy waf docs builds are unaffected.

Fix categories: `@param` tags left behind by parameter renames, malformed `@param x[out]` attribute placement, `\ref` targets that were renamed (fixed) or removed (demoted to `@c`), stale `@return` on functions that became void, `\file` statements with outdated names, unknown commands (`@prama`, `@params`, `@NOTE`, `@retuns`), and missing `@param` docs where the meaning is derivable from the signature or implementation.

Two systematic root causes worth knowing:
- `//!     @{` group-opens indented ≥4 spaces after a blank line are parsed as Markdown code blocks, so the group never opens — that was all seven "unbalanced grouping" warnings.
- A single unterminated backtick merges adjacent doc blocks and cascades into bogus `@param` warnings (three separate instances: `dict.h`, `ble_ad_parse.h`/`ble_client.h`, `timeline/item.h`).

The compositor.h file-level overview also referenced four functions that no longer exist anywhere in the tree; rewritten against the current API.

## Verification

- Full doxygen run: 0 warnings (re-verified after the rebase onto post-#1774 main).
- Diff audited to be comment-only (the sole exceptions: the two config lines above and hoisting a typedef above its doc comment in `drivers/accel.h`).
- Firmware builds clean (qemu_emery, 1794/1794).

## Deliberately not fixed (need code changes, separate PRs)

- `interval_timer.h`: the C parameter is literally named `weighting_factory_inverted`; docs now match the code, typo and all.
- `pbl_std.h:85` duplicates the `time_ms` declaration from `time.h` with a different parameter name.

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -1108,7 +1108,7 @@ EXCLUDE_SYMBOLS        =
 # that contain example code fragments that are included (see the \include
 # command).
 
-EXAMPLE_PATH           =
+EXAMPLE_PATH           = sdk/docs/external_refs
 
 # If the value of the EXAMPLE_PATH tag contains directories, you can use the
 # EXAMPLE_PATTERNS tag to specify one or more wildcard pattern (like *.cpp and
--- a/Doxyfile.rtd
+++ b/Doxyfile.rtd
@@ -15,3 +15,4 @@ QUIET                  = YES
 HTML_COLORSTYLE        = LIGHT
 HTML_EXTRA_STYLESHEET  = doxygen-awesome-css/doxygen-awesome.css
 GENERATE_TAGFILE       = docs/_apidoc/apidoc/pebbleos.tag
+WARN_AS_ERROR          = FAIL_ON_WARNINGS
--- a/include/bluetooth/bonding_sync.h
+++ b/include/bluetooth/bonding_sync.h
@@ -52,6 +52,7 @@ void bt_driver_handle_host_added_cccd(const BleCCCD *cccd);
 void bt_driver_handle_host_removed_cccd(const BleCCCD *cccd);
 
 //! Called by the BT driver after succesfully pairing a new device.
+//! @param bonding The newly created bonding.
 //! @param addr The address that is used to refer to the connection. This is used to associate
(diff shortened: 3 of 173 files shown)


gmarull (commented):
can we make sure CI fails if new problems are introduced?

rhaist:
> can we make sure CI fails if new problems are introduced?

I did want to make this a follow up PR but I can include the WARN_AS_ERROR config in here if you prefer.
This would fail the RTD CI on problems.

rhaist:
Side Note: I think the work on the doc and especially the code annotations should also make the varying quality of AI models more precise in their analysis and additions to the code base. I now catch myself actually thinking about optimizing code bases for both human and machine contributors.

gmarull:
> Side Note: I think the work on the doc and especially the code annotations should also make the varying quality of AI models more precise in their analysis and additions to the code base. I now catch myself actually thinking about optimizing code bases for both human and machine contributors.

agreed. docs should in fact be capable of replacing the skills maybe (skills could just point to a document maybe?)

gmarull (approved):

PR #1780, opened 2026-07-22
lib/c: default to picolibc, remove in-tree Pebble libc

gmarull:
## Summary

- Move the `__assert_func` override (the `assert()` hook of the newlib-family headers) from `src/fw` to `lib/c`, building it explicitly via `LIBC_USE` for every libc selection instead of relying on the `src/fw/*.c` glob.
- Switch the default C library from the in-tree pblibc to picolibc, which ships pre-built with the PebbleOS SDK toolchain and provides complete C99 formatted I/O at the smallest footprint.
- Remove pblibc entirely: `lib/c/pebble` sources and headers, the `LIBC_PEBBLE` Kconfig option and configure branch, the `pblibc`/`pblibc_includes` use targets (and their empty shims) across all wscripts, the `__aeabi_*` shims that `-nostdlib` required, and the pblibc host unit tests (`tests/libc`). Net: −29k lines.

Newlib (full/nano) remains selectable via Kconfig.

## Testing

- `obelix@pvt` (sf32lb52) and `asterix` (nRF52) build with pre-built picolibc; `__assert_func` resolves from `lib/c/assert.c` in the map.
- `obelix@pvt` also verified with `-DCONFIG_LIBC_PICOLIBC=y` prior to the default flip, and with pblibc before its removal.
- Full host unit test suite passes (`./pbl test`).


tests/libc/time/test_strftime.c
jplexer, line 1: strftime is actually not part of pblibc!

jplexer (changes requested):
There are still some stale pblibc references left in third_party/*/wscript_build 

Also technically %n was supported by pblibc while picolibc compiles it out, unsure if we care here though?

gmarull:
Addressed the feedback (squashed into the pblibc removal commit, b5a65d733):

- **Stale `pblibc` references**: removed from all `third_party/*/wscript_build` files, along with moddable's now-dead `CONFIG_LIBC_PEBBLE` qsort/bsearch branch. Verified with asterix and obelix builds.
- **`%n`**: picolibc compiles it out by default as format-string-attack hardening. Nothing in-tree uses it, and while apps do reach the firmware's `vsnprintf` through the exported `snprintf` (`pbl_snprintf`), `%n` use in watchapps is essentially non-existent — I'd treat dropping it as a security win rather than a regression, so I'm intentionally not re-enabling it.
- **strftime test**: kept, moved to `tests/fw/applib` — see inline reply.

tests/libc/time/test_strftime.c
gmarull: Good catch — it tests the firmware `pbl_std` strftime, not pblibc. Kept and moved to `tests/fw/applib` (minus the old `lib/c/pebble` include) in b5a65d733; passes against the host libc headers.

jplexer (approved):
good now, maybe squash in before merge but lgtm

PR #1779, opened 2026-07-22
docs: consolidate contributor knowledge into the docs site

rhaist:
First PR of the docs-consolidation series discussed in #1775: knowledge scattered across agent files, tool READMEs, and skills moves into the published docs, and the scattered copies become pointers ("docs should be capable of replacing the skills; skills point to a document").

## Consolidation

- **New: [Exposing functions to the SDK](https://pebbleos-core.readthedocs.io/en/latest/development/sdk_export.html)** — merges the three partial copies of this knowledge (AGENTS.md checklist, `tools/generate_native_sdk/README.md`, stale `sdk/readme.txt`) into one canonical page. The generator README becomes a pointer; `sdk/readme.txt` is deleted (it referenced a `src_readme.txt` that no longer exists).
- **New: Contributing page** — includes the root CONTRIBUTING.md (previously invisible on the docs site) plus the repo conventions (formatters, gitlint, logging levels) that only lived in AGENTS.md.
- **QEMU page** absorbs the operational detail from the `working-with-qemu` agent skill (monitor sockets, serial ports, uart1.log, screenshots, `sendkey`); the skill shrinks to a pointer plus agent-specific notes.
- **AGENTS.md** now points at the canonical pages instead of duplicating them, keeping one-line triggers for rules agents must follow unprompted.

## Accuracy audit

Every factual claim in the touched pages plus the five pre-existing development/board pages was verified against the current tree (the consolidated sources were Pebble-era and lied in places). Fixed as a result:

- SDK page corrections vs. the old README: `g_pbl_system_tbl` (not `_table`), artifacts generate into `build/sdk/<platform>/` (`libpebble.a`, not `sdk/lib/pebble.a`), `<area>_syscalls.c` naming, `exported_symbols.json` is an object (not an array), PRF/test variants skip the generator.
- `options.md`: `CONFIG_NOSTOP`/`CONFIG_NOSLEEP` don't exist in any Kconfig (pre-Kconfig relics, removed); default log level is INFO, not DEBUG.
- `getting_started.md`: `nodejs` (Ubuntu) and `gettext` (Fedora) are hard configure-time requirements but were missing; Fedora list brought to parity with Ubuntu (package names verified against Fedora 44); macOS now installs the checked-in `requirements-brew.txt`.
- `uart1.log` is written to the repo root, not `build/` (the skill had this wrong).
- asterix: renamed to current product name Pebble 2 Duo, keeping "announced as Core 2 Duo" as an alias.

## Link health

[...]

PR #1783, opened 2026-07-22
docs: add architecture section

rhaist:
Second PR in the docs consolidation series (follows #1779): adds an Architecture section to the docs site.

- **Architecture overview page** — a link-first tour of the firmware: layering, task model, boot and PRF variants, app identity / shell flow / privilege boundary, memory layout, BLE stack, PFS, driver splits and coredumps. The in-source design comments remain canonical; the page summarizes and points at them rather than duplicating.
- **Activity design doc move** — `src/fw/services/activity/docs/` (~30 KB on the step counting and sleep tracking algorithms, plus 7 figures) was referenced from nowhere and invisible on the docs site. Moved wholesale (pure rename) to `docs/architecture/activity/`.

Every referenced path and claim was verified against the current tree; stale assumptions found and corrected along the way: the accel driver interface lives in `include/pbl/drivers/` now, the bootloader is external to this repo (normal vs PRF is a build-time variant), NimBLE is the transport backend beneath the still-live `src/fw/comm/ble` host layer on all current boards, and only storage-loaded processes run unprivileged.

Verified with the local RTD-mirror build (`doxygen Doxyfile.rtd` + strict `sphinx-build -W`), which passes.


gmarull (approved):
many stuff is likely gona change architecture-wise, but ok to have something in place that can evolve

PR #1787, opened 2026-07-22
bluetooth-fw/nimble: recover host start state mismatch

jplexer:
## Summary

- queue NimBLE host starts through a Pebble-owned event
- recover `BLE_HS_EALREADY` and `BLE_HS_EBUSY` by stopping and retrying once
- preserve controlled failures for unrecoverable host start errors

This prevents a host state mismatch during stationary-mode exit from reaching
NimBLE's unconditional `ble_hs_event_start_stage2` assertion.

Fixes [FIRM-3454](https://linear.app/core-dev/issue/FIRM-3454)

## Testing

- `./pbl build` (`obelix@dvt`)
- tested successfully on a real Obelix device

### DIFF
--- a/src/bluetooth-fw/nimble/init.c
+++ b/src/bluetooth-fw/nimble/init.c
@@ -36,8 +36,11 @@ extern void nimble_gattc_op_queue_init(void);
 static TaskHandle_t s_ll_task_handle;
 #endif
 static TaskHandle_t s_host_task_handle;
+static SemaphoreHandle_t s_host_start_done;
 static SemaphoreHandle_t s_host_started;
 static SemaphoreHandle_t s_host_stopped;
+static struct ble_npl_event s_host_start_event;
+static int s_host_start_rc;
 static DisInfo s_dis_info;
 static struct ble_hs_stop_listener s_listener;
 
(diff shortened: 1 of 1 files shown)


dimitri-vs:
Hit this assert in the field on a PT2 running v4.31.2 (stand-by exit, 2026-09-05) and pulled the coredump. The root cause on my unit is different from the `EALREADY`/`EBUSY` state-mismatch cases this PR and #1247/#1272 target: `ble_hs_start()` took the normal `OFF -> ON` path and then `ble_gatts_start()` failed with `ENOMEM`, because `ble_gatts_count_cfg()` is re-run by the `ble_svc_*_init()` calls on every `bt_driver_start()` and `ble_hs_max_attrs`/`_services`/`_client_configs` are never reset within a boot. On the 38th start attempt the ATT table malloc had grown to 33,440 bytes and the largest free segment in the kernel heap was 33,300.

Full analysis with coredump globals, backtrace and heap walk in #1998.

Two notes for this PR: `ENOMEM` bypasses the `EALREADY`/`EBUSY` retry and goes straight to `PBL_CROAK`, so this PR would not have changed the outcome here. And `prv_configure_host()` sits inside the retry loop while `ble_gatts_reset()` clears neither the counters nor the pending `ble_gatts_svc_defs`, so a retry that does happen would count and append the service set a second time. Resetting the three counters is the fix that removes this failure; recoverable-failure handling is still worth having so a host start failure never asserts the whole watch.

PR #1790, opened 2026-07-23
tools/pbl: add QEMU touch injection (touch/swipe)

vvzvlad:
While developing touch navigation I needed a way to drive the touchscreen
under QEMU to iterate and test without flashing hardware, so I wrote this.

`./pbl qemu` now exposes a QMP socket, and two commands inject touch into a
running emulator:

    ./pbl touch 130 130
    ./pbl swipe 130 220 130 40 [--steps N --duration S]

Coordinates are screen pixels; the display size is read from the emulated
pebble-touch device and scaled to QEMU's absolute-axis range. Swipes stream
intermediate moves so drag gestures are seen as continuous. Injection uses the
absolute-pointer input path (input-send-event abs+btn); multi-touch is not
wired up in the device.

Also documents the workflow in the working-with-qemu skill.

Tested on qemu_gabbro: `./pbl qemu` creates the QMP socket; touch and swipe
land at the expected coordinates (verified against the pebble-touch device
registers).

### DIFF
--- a/.agents/skills/working-with-qemu/SKILL.md
+++ b/.agents/skills/working-with-qemu/SKILL.md
@@ -7,11 +7,28 @@ description: Use when building, launching, debugging, or capturing screenshots o
 
 Read `docs/development/qemu.md` — it documents the full workflow: qemu_*
 boards, `./pbl qemu` (monitor sockets, serial ports, uart1.log), `./pbl
-console`, `./pbl screenshot`, programmatic key input via `sendkey`, and
-`./pbl debug`.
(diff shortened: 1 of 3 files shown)


jplexer (approved):
lgtm, just before merging: this doesnt require a QEMU update?

vvzvlad:
> lgtm, just before merging: this doesnt require a QEMU update?


No, I just wrote a few helper functions to make it easier to inject touch events; everything inside was already set up for that.

PR #1798, opened 2026-07-24
docs: add binary format and protocol reference pages

rhaist:
## Summary

Next installment of the docs consolidation series (after #1779 and #1783): a reference section for the binary formats and wire protocols that so far were only documented in code comments.

New pages under **Reference**:

- **Binary formats**: PBF font resources, PBI bitmaps, PDC draw commands, the timezone database, and sparse length encoding
- **Pebble Protocol endpoints**: the BlobDB endpoint (`0xb1db`/`0xb2db`) wire protocol
- plus a note in the SDK export page that the revision ledger in `pebble_process_info.h` is the only rev↔minor mapping

The pages summarize the layouts and deep-link the packed structs, which stay authoritative. Every claim was verified against the current code rather than trusting the Pebble-era comments.

## Source cleanups

- `tools/font/fontgen.py` still described the **v1** font layout (offsets in 32-bit blocks, two-word glyph header); corrected to the v2/v3 reality (byte offsets, packed 5-byte header) that the tool and firmware actually implement
- `tools/bitmapgen.py` described a 5-bit `info_flags` format field; corrected to match the firmware's `BitmapInfo` (3-bit format + palette-heap flag) and marked the origin fields as ignored
- dead Atlassian wiki links in `fontgen.py` and `text_resources.c` replaced with pointers to the new pages
- pblconvert's `svg2pdc/pdc.py` carried a drifted duplicate of the PDC format docstring from `generate_pdcs/pebble_commands.py`; deduped to a pointer

## Notes for reviewers

While verifying the specs a few pre-existing code issues surfaced (BlobDB `MIN_DELETE_LENGTH` off-by-one, a region-id bounds check in `timezone_database/service.c`, SLE + pblconvert svg2pdc being dormant/py3-broken). Deliberately **not** touched here to keep this docs-only; happy to follow up separately.

### DIFF
--- a/docs/development/sdk_export.md
+++ b/docs/development/sdk_export.md
@@ -97,3 +97,8 @@ Notes:
 - The generator errors out on exports it cannot find in the parsed headers
   and on inconsistent revision numbers, but it does not verify that the
   resulting `pebble.h` compiles — review its output.
+- The comment ledger above `PROCESS_INFO_CURRENT_SDK_VERSION_MINOR` is the
+  only mapping between SDK revisions and version minors: no formula relates
+  them (the minor once jumped `0x19` → `0x20` between revs 35 and 36, and a
+  few minors and revs are skipped or doubled up). Treat the ledger as
+  append-only history.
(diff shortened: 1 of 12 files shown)


jplexer (approved):
great to have these documented! thank you!

PR #1801, opened 2026-07-24
Fix pre-existing bugs found during the docs consolidation sweeps

rhaist:
## Summary

Follow-up promised in the reviewer notes of #1798 and #1799: while claim-verifying the formats/testing/debugging pages against the code, a number of small pre-existing bugs surfaced. This PR fixes them, one commit each.

**Firmware**

- `blob_db/endpoint.c`: `MIN_DELETE_LENGTH` was 6, rejecting a legal DELETE with a 1-byte key (minimum body is token 2 + db id 1 + key size 1 + key 1 = 5). The per-field parsing below the length check still rejects malformed messages.
- `timezone_database/service.c`: `timezone_database_load_region_name()` used `>` instead of `>=`, so `region_id == region_count` read one entry past the region table.

**Tools**

- `timezones.py`: the `tz_abbr.replace("*", "S")` result was discarded, so regions whose DST rule is no longer supported shipped a literal `*` in their abbreviation (e.g. `CE*T`); the runtime substitution in `time_get_timezone_abbr()` hides this on the display path, but `timezone_database_load_region_info()` copies the raw abbreviation.
- `gdb_printers.py`, `analyze_static_memory_usage.py`: still Python 2 (`xrange`, `itervalues`, str-based UUID bytes); crashed under Python 3.
- `dehash_flash_logs.py`: default dictionary fallback pointed at `build/pebbleos_loghash_dict.json`; the build produces `build/src/fw/loghash_dict.json`.
- `sparse_length_encoding.py`: the single-file CLI branch referenced an undefined variable and joined bytes with `str.join`, raising on any invocation. (SLE has been dormant since the Spalding removal — if you'd rather delete it than fix it, happy to do that instead.)
- pblconvert `svg2pdc/pdc.py`: still Python 2 despite `requires-python >= 3.9` (StringIO import, invalid `(object, Command)` MRO, str magic concatenated with bytes).
- `pebble_commands.py` / `timezones.py` / `pdc.py`: multi-byte `struct.pack` fields now explicitly little-endian; byte-identical output on little-endian hosts.

**Docs**

- New **Coordinate grid** section on the PDC format page, preserving the guidance from the dead Atlassian wiki link that the svg2pdc pixel-grid annotation used to reference; the annotation now links there (same pattern as the fontgen/text_resources link replacements in #1798).
- `tools/font/README.md`: dropped dead 2012-era freetype links.

## Verification

- Full `asterix` build and the complete unit-test suite pass.
- `timezones.py`: generated tzdata diffed before/after — every changed byte is exactly `*` → `S` (e.g. `CE*T` → `CEST`, matching what the firmware substitutes for a region never in DST); the endianness change is byte-identical on LE hosts.
- `sparse_length_encoding.py` CLI round-trip-verified on random data; `svg2pdc/pdc.py` import + serialization verified under Python 3 (the full pblconvert test suite doesn't run on modern hosts due to the `lxml==3.4.4` pin).
- Docs: strict `sphinx-build -W` passes; the `#coordinate-grid` anchor verified in the rendered HTML.

### DIFF
--- a/docs/reference/formats/pdc.md
+++ b/docs/reference/formats/pdc.md
@@ -27,6 +27,16 @@ Path points are whole pixels; precise-path points are 13.3 fixed-point
 (eighths of a pixel). A command list is a `uint16_t` command count
 followed by that many commands.
 
+## Coordinate grid
+
+The SVG converters translate input coordinates by (-0.5, -0.5) and round,
+so source coordinates must sit on the half-pixel grid: multiples of
+0.5 px for paths and circles, 0.125 px (1/8) for precise paths. Off-grid
+points are snapped to the nearest supported coordinate and flagged
+(`convert_to_pebble_coordinates` in
+`tools/generate_pdcs/pebble_commands.py` and the annotation emitted by
+`tools/libs/pblconvert/pblconvert/svg2pdc/pdc.py`).
+
 ## File containers
 
 **PDCI (image)**:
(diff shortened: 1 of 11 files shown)


jplexer (commented):
Thank you! did a quick fixup of one commit

jplexer (approved):

PR #1803, opened 2026-07-24
fw/music: album art cover support

ericmigi:
## Album art on the Music app

Adds album-art cover support to the Music app (MOB-10062). When the paired phone provides artwork, the music screen shows the cover filling the top, the artist name centred and outlined over the bottom of the cover, and the track title as a one-way "digital-sign" ticker beside the tape (two passes after a track change, then rests at the start). The clock is drawn large + outlined so it reads over the art. When there's no art (or the feature is off/unsupported) the screen falls back to the exact original text layout.

### Wire protocol
Watch-pull model over the existing Music endpoint (0x20):
- Watch sends `GetAlbumArt` (0x09) when a track needs art and the phone advertises support.
- Phone replies with chunked `AlbumArtResponse` (0x14): `[token][flags][offset][len]([w][h][paletteCount][palette]) [pixels]`, 4bpp palettized (GColor8). A `NO_ART` chunk tells the watch to stay on the text screen.

### Capability gating / graceful fallback
Strictly pull-driven and capability-gated, so it stays inert unless **both** ends are new:
- Firmware requests art only when the phone advertises `CommSessionMusicAlbumArtSupport` (bit 17) and re-checks before sending.
- Old phone / iOS-AMS music source never advertises the bit (the AMS impl doesn't set the album-art fn-pointers) → firmware never requests → text-only screen.
- New phone + old firmware: old firmware never sends `GetAlbumArt`, so the phone never sends art.
- Reassembly is bounds-checked (dims/palette/size caps, in-order chunks) and freed on disconnect.
- The new `musicShowAlbumArt` synced pref: old firmware rejects the unknown key with a benign `WRN` (no crash).

### Testing
- Emulator (qemu_emery) across 5 tracks/covers incl. short + long titles.
- Real hardware: obelix (Pebble Time 2) via CoreApp sideload, live playback from Android.

Companion phone-side change: coredevices/CoreApp#(album-art in libpebble3).


ericmigi:
@jplexer tested on obelix, probably could use a bit more tweaking eventually of the scroll motion but it's a good starting point

jplexer:
not 100% convinced by the layout (at least on oblx), will tweak around

ericmigi:
gabbro is not looking good yet, still working on it

ericmigi:
ok gabbro and emery are looking good now, @jplexer can you retest when you get a sec - not urgent tho, just sometime this week pls

src/fw/services/imaging/imaging.c
```diff
@@ -0,0 +1,218 @@
...
+  } else if (bitmap) {
+    kernel_free(bitmap->addr);
+    kernel_free(bitmap->palette);
+    kernel_free(bitmap);
+  }
+}
+
+bool imaging_is_supported(void) {
+  CommSession *session = comm_session_get_system_session();
+  return session && comm_session_has_capability(session, CommSessionImagingSupport);
```
sjp4: this is also checked below - but this one is only called from test code?

src/fw/services/imaging/imaging.c
```diff
@@ -0,0 +1,218 @@
...
+}
+
+bool imaging_request_album_art(uint8_t token, ImagingFormat format, uint16_t width, uint16_t height,
+                               const char *title, const char *artist) {
+  CommSession *session = comm_session_get_system_session();
+  if (!session || !comm_session_has_capability(session, CommSessionImagingSupport)) {
```
sjp4: `CommSessionImagingSupport` doesn't mean album art support (i.e. if we add another image type on ios (e.g. timeline pin image / notification image) and turn on that capability, then album art will ping the phone for every track) - do we need a separate capability for album art? or e.g. an error response from the imaging service for `UNSUPPORTED_TYPE` which the FW remembers and doesn't query for that type again for the life of the connection?

src/fw/services/imaging/imaging.c
ericmigi: Deduped — replaced `imaging_is_supported()` with `imaging_is_type_supported(type)`, which is the single predicate now used by both the app gate (music.c) and `imaging_request_album_art()`, so the check no longer appears twice.

src/fw/services/imaging/imaging.c
ericmigi: Addressed — added `ImagingResponseFlagUnsupported`. When the phone can't serve a type it replies Unsupported; the watch latches that type off for the connection (`s_unsupported_types`, reset when the session changes) and `imaging_is_type_supported(type)` returns false, so a generic imaging capability enabled for some other type (e.g. notification images on iOS) won't make album art poll every track.

src/fw/services/music/endpoint.c
```diff
@@ -164,6 +168,147 @@ static void prv_update_player_info(CommSession *session, const uint8_t* msg, siz
...
+  if (length < sizeof(MusicEndpointAlbumArtChunkHeader)) {
+    return;
+  }
+  const MusicEndpointAlbumArtChunkHeader *hdr = (const MusicEndpointAlbumArtChunkHeader *)msg;
+  const uint8_t *cursor = msg + sizeof(*hdr);
+  const uint8_t *msg_end = msg + length;
+
+  // Ignore chunks that aren't for the track we currently have (art that raced with a track change).
+  // Don't tear down an in-progress transfer for the current track: just drop the stale chunk.
+  if (hdr->token != music_get_now_playing_generation()) {
```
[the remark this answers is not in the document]
ericmigi: Fixed by the redesign — reassembly moved from music/endpoint.c to services/imaging/imaging.c, where a chunk whose token doesn't match the active transfer resets it before dropping: `if (!s_rx.active || s_rx.token != hdr->token) { prv_rx_reset(); return; }`, and every First chunk resets first. So a track change (new token) reclaims the old transfer's buffers on the next request instead of pinning them.

src/fw/services/music/endpoint.c
```diff
@@ -164,6 +168,147 @@ static void prv_update_player_info(CommSession *session, const uint8_t* msg, siz
...
+      return;
+    }
+    GBitmap *bmp = kernel_zalloc(sizeof(GBitmap));
+    if (!bmp) {
+      prv_art_rx_reset();
+      return;
+    }
+    bmp->addr = s_art_rx.pixels;
+    bmp->row_size_bytes = s_art_rx.row_size_bytes;
+    bmp->info.format = GBitmapFormat4BitPalette;
```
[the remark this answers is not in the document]
ericmigi: Valid and carried over into the new code — the Music app requests `ImagingFormat4BitPalette` unconditionally (no PBL_COLOR gate), so a 1-bit target would mishandle it. Leaving this open as a follow-up: gate the request (and has-art layout) on colour displays, or negotiate a 1-bit format. Not addressed in this PR.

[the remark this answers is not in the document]
ericmigi: Fixed in 64fa351 — album art is now gated to emery and gabbro (`#if defined(CONFIG_PLATFORM_EMERY) || defined(CONFIG_PLATFORM_GABBRO)`, matching app_manager.c). Flint and lower never request art, so their layout stays text-only and 1-bit targets can't be handed a 4BitPalette image. Verified: emery builds + music tests pass; flint builds clean.

ericmigi:
@sjp4 updated

jplexer (approved):
did some modifications (specifically regarding the getafix music app), otherwise LGTM

ericmigi:
need to wait on mobile side

PR #1807, opened 2026-07-27
correct the string "just once"

caco3:
The alarm scheduling contains the string "Just Once", how ever in the translation files only the strings "Once" are contained. Since "Just Once" is not used anywhere else, changing it seems to be the best approach.


jplexer:
it has always been just once

caco3:
@jplexer OK, then I will add the missing string to the PO file. Currently only "Once" is contained there, but not "Just once"

caco3:
superseded by https://github.com/coredevices/PebbleOS/pull/1808

Issue #1537, opened 2026-06-16
Add setting for One-Time Alarms to delete themselves after they are done


pizzaisdavid:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

Yesterday I had a train at 1pm.
I created an alarm: `Basic Alarm`, `12:30`, `Once`.

It triggered, I dismissed it, all good.

Today I go to the alarm app, and see I am collecting a bunch of One-Time Alarms. In my case, they are not reusable. And they just make clutter and busy-work to delete them.

It would be good if there was an option (somehow) that one-time alarms should delete themselves after being dismissed (not snoozed.)

asyba:
Didn't you try looking for an app that have this feature in the store https://apps.repebble.com/apps?

pizzaisdavid:
> Didn't you try looking for an app that have this feature in the store 

Maybe it some language issues, but no I didn't.

And ideally, I wouldn't have to use multiple alarm apps.

asyba:
you can delete the stock pebble app and use other.

pizzaisdavid:
How to delete? I cannot figure out how to delete.

Where to give feedback for the stock app?

asyba:
Actually you are right, we cannot delete the stock alarm I confuse myself.
Try this one maybe https://apps.repebble.com/alarms_54551598997cdfddaa000043

This is the correct place to request features or changes.

caco3:
I am also missing this feature. And i do not want to use an extra app for this.

Thus I started implementing this feature: https://github.com/coredevices/PebbleOS/pull/1810

PR #1810, opened 2026-07-27
alarms: add a new kind of alarm clock mode: `Just once then delete` to auto-delete one-time alarms

caco3:
This PR adds a new kind of alarm clock: `Just once then delete`:

grafik 
grafik 
grafik

---

This implements https://github.com/coredevices/PebbleOS/issues/1537

I often use my watch to set an alarm when cooking or when to leave to catch the train. Those alarms are often onte time shots and must be deleted again manually afterwards.
With this PR, the alarms will get deleted automatically after they got dismissed.

## Implementation concept
- First commit adds the deletion functionality
- 2nd commit updates the alarm list after the alarm got deleted. This is needed since the user could have the alarm list open when the alarm gets dismissed + deleted. The alarm still would be shown in the list and only get updated when the list gets reloaded.
   However the dismiss action happens in one process (the popup), and the alarm list lives in another (the alarms app). They're separate. The popup can delete the alarm from storage, but it has no way to reach into the alarms app's UI state. So it sends the newly created system event; the alarms app listens for that event and refreshes its list when it arrives.


caco3:
The MR is ready now. I had to shorten the strings so they do not get truncated.

include/pbl/services/alarms/alarm.h
```diff
@@ -161,7 +161,8 @@ uint16_t alarm_get_snooze_delay(void);
 void alarm_set_snooze_delay(uint16_t delay_m);
 
 //! Dismisses the most recently triggered alarm.
-void alarm_dismiss_alarm(void);
+//! @return True if the alarm was auto-deleted (ALARM_KIND_JUST_ONCE_DELETE), False otherwise.
+bool alarm_dismiss_alarm(void);
```
jplexer, line 165: that definition doesnt really make sense, why is auto-deleted "true"? I'd consider true/false to be equal to succeeded/failed

caco3: `true` means the alarm was deleted. Based on this, the dialog message will be different ("alarm deleted" instead of "alarm dismissed".

The alternative would be to provide a new public API function which tells us the kind of the alarm or its ID. I think the approach with the boolean value is less intrusive.

N.B. I originally wanted to show "alarm dismissed and deleted", but that is too long and gets truncated on the 2-line dialog text field.

caco3:
@codex review

src/fw/services/alarms/alarm.c
```diff
@@ -1048,9 +1058,17 @@ void alarm_set_snooze_delay(uint16_t delay_m) {
 }
 
 
-void alarm_dismiss_alarm(void) {
+bool alarm_dismiss_alarm(void) {
   prv_clear_snooze_timer();
 
+  if (s_most_recent_alarm_id != ALARM_INVALID_ID &&
+      s_most_recent_alarm_config.kind == ALARM_KIND_JUST_ONCE_DELETE) {
+    alarm_delete(s_most_recent_alarm_id);
```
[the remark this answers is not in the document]
caco3: fixed

caco3:
@jplexer  @gmarull Please review and provide feedback

gmarull:
please rebase

caco3:
> please rebase

Done

caco3:
Hi @gmarull
I rebased it 2 weeks ago on your request. Nevertheless there was no action on your side. Now there are new conflicts.

I feel I am wasting my time with this...
Unless you can give some commitment to have a look on it I will not invest more time into this :(

PR #1811, opened 2026-07-28
ci: publish OTA payloads and symbols to eng-dash alongside Memfault

ericmigi:
eng-dash is becoming the OTA server for Core watches. This makes a tagged release publish its artifacts there as well, without changing anything about the Memfault path — both get the same merged dual-slot `.pbz` for as long as the fleet needs it.

## What the release job now does

- Uploads each `normal_<board>_<tag>.pbz` and its unstripped `.elf` to the OTA bucket under `ota/<tag>/` and `elf/<tag>/`.
- POSTs a manifest (per-artifact `sha256`, size, build id) to eng-dash's ingest endpoint.

The sha256 is load-bearing: CoreApp verifies it against the bytes it downloads and refuses to flash on mismatch, so a tampered or corrupted artifact cannot reach a watch.

Uploads **refuse to overwrite an existing key**. A moved or re-pushed tag would otherwise silently republish different bytes under a version users may already have been offered; now that fails the step loudly and requires a deliberate human action.

## Configuration

Bucket name, endpoint and credentials reuse the existing `LOG_HASH_BUCKET_*` secrets/vars — the shared-credential blast radius is documented inline in the workflow. One new repository secret is required:

- `ENGDASH_OTA_TOKEN` — bearer token for the eng-dash ingest endpoint

## Testing

The full path was exercised by hand against the real v4.31.1 artifacts before this PR: pbz uploaded to R2, manifest POSTed to the ingest endpoint, release and per-hardware artifacts created, and a watch then downloaded the firmware through the resulting signed URL with the sha256 verifying byte-exact. Steps are gated on `github.repository == 'coredevices/PebbleOS'` as with the existing Memfault and loghash steps.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -288,3 +288,119 @@ jobs:
             deploy-release \
             --release-version ${{ github.ref_name }} \
             --cohort internal
+
+      - name: Upload OTA artifacts to R2
+        if: ${{ github.repository == 'coredevices/PebbleOS' }}
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.OTA_BUCKET_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.OTA_BUCKET_SECRET }}
+          AWS_DEFAULT_REGION: us-east-1
+          BUCKET: ${{ vars.OTA_BUCKET_NAME }}
+          ENDPOINT: ${{ vars.OTA_BUCKET_ENDPOINT }}
+          TAG: ${{ github.ref_name }}
+        run: |
+          set -o pipefail
+
+          pip install awscli
+
+          # Write-once by content: a moved or re-pushed tag must never silently
+          # swap the bytes behind a version that may already be promoted to
+          # users. Re-uploading the identical file is a no-op, so a job that
+          # died partway (or after the manifest POST failed) can simply be
+          # re-run without hand-deleting the objects it already wrote.
+          upload() {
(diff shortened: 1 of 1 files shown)


ericmigi:
Rebased onto current `main` and addressed the review.

**Scope** — the PR was accidentally cut from a local branch point that carried an unrelated `kernel: protect scheduler state from apps` commit, which is what dragged in the 11 kernel/graphics files and the FreeRTOS submodule bump. That commit is gone; the PR is now a single-file diff against `.github/workflows/release.yml`. This also resolves the "split the scheduler hardening into its own commit" comment and moots the `system_resource.c` font-pointer comment, since neither file is part of this change any more.

**Retryable uploads** (the one comment that applied to this change, and a fair catch) — the write-once guard is now write-once *by content*:

- object already holds the identical bytes (same `ContentLength` and MD5/ETag) → skipped, step continues
- object exists with different bytes → step aborts with a clear error
- object absent → uploaded

So a run that dies after uploading some artifacts, or after the manifest POST fails, can just be re-run; only a genuine attempt to republish different firmware under an already-published key stops the job. The comparison logic was exercised against identical / differing-digest / differing-size inputs, and the workflow YAML parses.

ericmigi:
@jplexer this can land anytime, still working on coreapp stuff but this at least starts stuff flowing.

jplexer (approved):
@ericmigi lgtm, merge whenever you feel like it

jplexer:
or ill merge now before 4.32

PR #1812, opened 2026-07-28
Corrected the German translation of "Alarm dismissed"

caco3:
[DE] ignoriert = [EN] ignored

### DIFF
--- a/resources/normal/base/lang/de_DE/tintin.po
+++ b/resources/normal/base/lang/de_DE/tintin.po
@@ -2263,7 +2263,7 @@ msgstr "Für %d Minuten schlummern"
 
 #: ../src/fw/popups/alarm_popup.c:71
 msgid "Alarm dismissed"
-msgstr "Wecker ignoriert"
+msgstr "Wecker ausgeschaltet"
 
 #: ../src/fw/popups/ble_hrm/ble_hrm_reminder_popup.c:17
 msgid ""


resources/normal/base/lang/de_DE/tintin.po
```diff
@@ -2263,7 +2263,7 @@ msgstr "Für %d Minuten schlummern"
 
 #: ../src/fw/popups/alarm_popup.c:71
 msgid "Alarm dismissed"
-msgstr "Wecker ignoriert"
+msgstr "Wecker ausgeschaltet"
```
juliusheyer, line 2266: I think I'd find "Wecker beendet" more natural

jplexer: nah I think ausgeschaltet is better than beendet tbh

jplexer (approved):

resources/normal/base/lang/de_DE/tintin.po
caco3: > I think I'd find "Wecker beendet" more natural

To my feeling (as a German speaker) and also [Google AI](https://share.google/aimode/aOKN6iZeotUpfj22O) "Ausgeschaltet" is the best option.

juliusheyer: I can for sure live with either, thx for fixing the German translations

PR #1813, opened 2026-07-28
Watchface selector on touchscreen long press

ericmigi:
## Summary

Hold a finger on the watchface for 2 s (touchscreen platforms only, e.g. obelix/getafix) to open a new carousel selector UI in the Watchfaces app:

- Full-screen card per installed watchface: icon, name, Active badge, position counter, up/down arrows (`content_indicator_draw_arrow`).
- Swipe up/down slides between watchfaces (with wrap-around); the face landed on is immediately persisted as the new default, so backing out returns to it.
- Tap or SELECT launches the face (`APP_LAUNCH_USER`, so app_manager persists it as default). BACK exits. UP/DOWN buttons also navigate.
- Entered only via `WatchfacesLaunchArgs.selector_mode` from the shell long-press hook; the regular Watchfaces menu list is unchanged.

Implementation:

- `src/fw/shell/normal/watchface.c`: kernel-side touch subscription + 2 s hold timer (20 px slop cancel; suppressed while a modal is focused, in low power, or when a non-watchface is foreground).
- `src/fw/apps/system/watchfaces.c`: the selector UI. Touch contacts starting within 500 ms of window load are ignored so the entry long-press finger is not read as a tap on real hardware.
- All touch code is gated on `CONFIG_TOUCH`; non-touch platforms compile it out (verified with a `qemu_flint` build).

Also fixes a regression found while testing: `touch_has_app_subscribers()` counted any kernel-side subscriber as an app, so the new permanent subscription made the event loop light the backlight on every touchdown and bypass the touch-wake preference. Fixed with a new `touch_set_system_subscribed()` opt-out that only the watchface long-press hook uses -- the touchdown-driven backlight for third-party touch apps is unchanged (unit test added).

## Power

- Master Touch setting off: chip fully sleeps, feature disabled — unchanged.
- Touch on + touch-wake enabled (default): chip was already powered 24/7 for the backlight subscription — zero delta.
- Touch on + touch-wake off: chip now idles in its auto-sleep mode (µA) instead of deep sleep so the long press works; known tradeoff, fallback is a foreground-only subscription if telemetry flags it.

## Testing

- Verified end-to-end in the getafix QEMU emulator (qemu_gabbro) with injected touch: long-press open, swipe navigation + wrap, swipe-persists-default, tap-launch, back-exit, drag-slop cancel, button navigation.
- `./pbl test` suite green; new `test_touch__has_app_subscribers_system` fails before the backlight fix and passes after.
- Dual-slot getafix dvt2 PBZ built and sideloaded for on-device testing.

### DIFF
--- a/include/pbl/services/touch/touch.h
+++ b/include/pbl/services/touch/touch.h
@@ -24,7 +24,14 @@ void touch_init(void);
 //! When disabled, the touch sensor is only active if apps have subscribed to touch events.
 void touch_set_backlight_enabled(bool enabled);
 
-//! @return true if at least one subscriber is currently registered for touch events.
+//! Mark (or unmark) one subscription as a system-internal consumer that must
+//! not count towards touch_has_app_subscribers(). Call with true after
+//! subscribing and with false before unsubscribing.
+void touch_set_system_subscribed(bool subscribed);
+
+//! @return true if at least one subscriber other than the system-internal
+//! ones (backlight, subscriptions marked via touch_set_system_subscribed())
+//! is currently registered for touch events.
 bool touch_has_app_subscribers(void);
 
 //! Globally enable or disable touch. When disabled:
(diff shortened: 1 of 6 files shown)


ericmigi:
@jplexer not urgent, but the next time you're doing some power testing on a ppk2, could you try running this for 1 min and benchmark power consumption compared to regular fw/same watchface?

PR #1819, opened 2026-07-30
Increase density of getafix list views (timeline, alarms, default menus)

jplexer:
Increases density of the remaining sparse list views on large round displays (getafix), per MOB-10039:

- **MenuLayer default heights** (`applib/ui/menu_layer`): menus without a `get_cell_height` callback previously got a flat 61 px (Large content size) row height. On center-focused menus on displays ≥ 200 px round, the default is now selection-aware: 68 px focused / 32 px unfocused, fitting two unfocused rows per side. This covers "everything that uses the default list view", including third-party apps that don't override cell heights. The default path now also applies the same selection-index masking that callbacks receive, so selection-dependent heights stay consistent during the upward geometry walk.
- **Alarms app**: dropped its flat basic-height callback so it inherits the new default — the whole alarm list is now visible at once (time + ON/OFF per unfocused row, frequency subtitle on the focused row).
- **Timeline**: the round list view showed one full event plus only the start time of the next. It is now centered like the round menus — **previous, current and next** are all on screen: the previous pin stays visible as a thin pin above the focused one and the next pin shows below, each rendering time + a one-line title in the lighter pin-subtitle face so the focused event stays dominant. Long titles ellipsize inside the display circle. Slot 0 (previously an offscreen animation slot) is retained on screen; after backward scrolls and resets it is refilled via a new `timeline_model_prepare_prev()`. The fin and the day-separator dot take the empty bottom/top slot as before. Updated gabbro goldens — the image diffs in this PR show before/after for both directions.
- **`pin test` console command**: injects past/future calendar pins (short and long titles, plus a next-day pin) so the timeline can be exercised in QEMU without a companion phone, mirroring `notif test`.

Not changed: settings menu, health day lists, notification list — already densified (FIRM-3418, MOB-10373/#1818). The sleep "recent days" view from the issue is the health detail card, whose density fix already shipped in v4.28.0.

## Verification

- QEMU gabbro: walked the future timeline forward and backward across a day boundary (previous-pin backfill, day-separator dot, end-of-timeline fin, pin card open/close), alarm list with 4 alarms, settings/health menus — screenshots reviewed at each step.
- Full unit test suite passes (323/323); `test_timeline_list_view_gabbro` goldens regenerated (past + future directions, fin and day-dot states).
- Firmware builds clean for `qemu_gabbro`, `getafix@dvt2`, and `obelix@pvt` (rect coverage).

Fixes MOB-10039


jplexer:
I did not tell claude to open a PR yet

PR #1836, opened 2026-08-04
apps/weather: full forecast app for round + rect (weather DB v4)

bowenperks:
## What this is

The complete system Weather app (round **gabbro/getafix** + rect **emery/obelix**), developed on a v4.18.0 base and ported onto current upstream. Seven commits, all weather work.

## App surface

- **Main screen** — current-hour conditions header, five-day strip (today first) with animated condition icons, hi/lo temperature graph, scrolled precipitation view, Health-style nav arrows.
- **Weather report** — per-day condensed report page (day name at the crown, curved temp column, condition sentence, UV index bar) with the newspaper open/close scene on SELECT.
- **Sunset card** — sunset time, hi/lo, precipitation pill, current-hour UV bar.
- **Clock face** — 12-hour dial with hourly conditions and current temperature.
- **City globe** — 3D globe city picker + phone-synced saved-locations list (curved rows on round).
- **Touch** — swipe navigation mirroring the buttons on touch boards; the app opts out of the system touch-nav twin since it carries its own complete gesture map.

## Protocol

Weather DB **v4 schema, minor 4**: daily forecast array, hourly type/temp/UV series, per-location coordinates. v3 records still parse; the phone only writes v4 when the firmware advertises support (`session_remote_version`). Worked example in `weather_db_v4_example.md`. 

## Round display tuning

One render per display frame (timer beat eliminated), capture-once squash transitions, frame-aligned 66ms animation rates, transparent window backgrounds where the canvas proc guarantees full pixel coverage — tuned for the JDI panel's forced full-screen updates.

## Port notes (v4.18 → v4.32)

- Include paths updated for the `pbl/` header reorganisation.
- `resource_ids.pin.h` now aliases the **generated symbolic ids** — the numeric pinning it used previously silently broke when the system pack reordered (caught during this port: every 25px bitmap drew its neighbour).
- `-fdata-sections` enabled in `tools/waf/pebble_arm_gcc.py` (lets `--gc-sections` drop unreferenced data).

## Verification

- `qemu_gabbro`: boots, full navigation sweep (main → report → card → globe → clock and back), bitmaps/PDCs verified per-screen.
- `qemu_emery`: builds clean, navigation sweep verified.
- Hardware: running on a getafix Pebble Round 2 and an obelix Pebble Time 2 (both slots exercised, OTA installs, touch verified on both panels).


jplexer (changes requested):
Great Stuff! but not mergable as is :c 

asterix and qemu_flint don't compile. globe_view.c:1282 and :1611 use touch-gated symbols (globe_view.h:99), and forecast_list.c:3152+ uses the s_select_exit_* statics that flint's config compiles out. All the other red CI jobs are just fail-fast cancellations from these two.

Once that compiles, 1bpp boards get memory corruption: weather_math.c:53, weather_app_layout.c:1928 and forecast_list.c:2168 write one byte per pixel, but an asterix/flint framebuffer row is 20 bytes, so rows overrun by up to ~143 bytes on the base screen.

The v4 insert validation has a phone-reachable hole. weather_db.c:150 leaves the trailing SerializedArray header out of the min-length check, so a 201-byte record gets accepted, the strings resolver reads out of bounds, and a junk length goes into task_zalloc_check, which reboots the watch. Malformed phone data should never croak us. While you're in there: bound the whole string block against val_len (both pstring lengths are phone-controlled and unchecked), NULL-check the pstrings in weather_service.c:53/82, reject minor > CURRENT (weather_db.h:194), and gate the version read on rv == S_SUCCESS (weather_db.c:191).

globe_view.c:957: applib_malloc croaks on OOM in a privileged app, so the NULL-check below is dead code and heap pressure reboots the watch. malloc_try.

forecast_list.c:3584: the hslide PropertyAnimation is the one animation unload never cancels, so double-BACK during the slide is a use-after-free.

weather_report.c:345: the round arm is missing the update-before-load guard the rect arm has at :637.

clock_face.c:112, expanded_view.c:70, weather_report.c:51: these statics aren't reset on launch, so after an app crash the relaunch writes through stale pointers. Reset them from prv_init like s_page.

weather.c:202: <= WeatherType_Generic excludes Sun, so the current-hour override never fires for sunny weather.

weather_data_source.c:31: is_current_location = (index == 0) throws away the record's real flag. Should be f->is_current_location.

clock_face.c:278: after about 13:00 the dial wraps into this morning's data and presents it as tonight's forecast.

The PBL_PLATFORM_GABBRO blocks (~280 lines) never compile: fw uses CONFIG_PLATFORM_GABBRO.

The files are definitely not copyright Google LLC

Please don't restyle the existing 80x80 SVGs (Sunny_day, Cloudy_day, Partly_cloudy, Generic_weather). Those PDCs are shared, other apps and timeline pins render them. Add weather-specific copies like you did with the wx_* PNGs.

resource_map.json: append IMAGE_SUNNY_DAY_SMALL instead of inserting at position 0 (that renumbers the whole pack), and drop the unrelated degree sign escaping churn.

There are some mangled comments ("both shapes since )" at forecast_list.c:2530 for example), a lot of development-diary comments that belong in commit messages, exec bits on .c/.h/.png/.pdc files, and DESIGN_NOTES.md ships personal notes ("FOR STEVE") that probably weren't meant to land.

PR #1841, opened 2026-08-06
fw/services/battery: add optional battery charge limit

### LINKED ISSUE #722: Ability to limit battery charge to X%

adipascu:
Adds an optional battery charge limit that pauses charging at a configurable percentage to reduce long-term battery degradation. This picks up the idea from #1156 and reworks it per the review feedback there. Tested on a Pebble Time 2 together with the companion app change (coredevices/mobileapp#318).

Fixes #722

### fw/services/battery: add optional battery charge limit

The limiter runs on every battery state change event. Once the battery reaches the configured limit while plugged in, charging pauses through the nPM1300 charger enable (`BCHGENABLECLR`) and re-enables below the limit, on unplug, or when the limit is turned off. Per @gmarull's feedback on the original PR there is no resume hysteresis. While plugged in, SYSREG supplies VSYS from VBUS and the battery is isolated (nPM1300 PS v1.1, 6.2.10), so the level parks at the limit. If a heavy load ever drains the battery through supplement mode, the drop below the limit re-enables charging on the next state change event.

The `chargeLimitPct` preference (`0` disables, `50`-`95` accepted) is off by default and set from the phone through the settings blob db whitelist, per @ericmigi's feedback to keep new settings off the watch UI. Values are validated on sync and clamped at boot, since the boot load bypasses the validating setters. Preference changes force a battery state update so a new limit applies immediately, replacing the 60 s polling timer from the original patch. The limiter also stands down while charging is force-disabled from the debug prompt, so `battery charge disable` keeps full manual control.

Compared to the original patch this also fixes a nonexistent `system/logging.h` include that broke the build, and a path that left the PMIC charger permanently disabled after unplugging at the limit, because `BCHGENABLECLR` is persistent and nothing re-enables it on replug. `test_battery_charge_limit` covers the state machine.

> [!NOTE]
> Known limitation: the limiter evaluates on the once-per-minute fuel gauge sample, so at the original 1C rapid charge rate the level can run about 2% past the configured limit before charging pauses (an 80% limit can park showing 81%). The follow-up #1844 tightens the sampling while plugged in to bound the overshoot to a fraction of a percent.

### fw/services/battery: enable charger before emitting state events

The boot-time one-shot charger enable ran after the state change event was posted. The event handler runs on KernelMain, which preempts KernelBG, so booting plugged in above the limit would pause charging and then have the one-shot silently re-enable it for the session. The enable now happens before the event.

### drivers/qemu: honor the argument in battery_set_charge_enable

The QEMU battery driver faked an unplug instead of tracking charger enablement, which made the limiter untestable in the emulator and permanently cleared the emulated USB connection.

### fw/apps/settings: show charge limit in system information

Since the pref is phone-only, Settings > System > Information gains a read-only Charge Limit row (shows Off while disabled) so the active limit can be checked on the watch itself. Captured on a Pebble Time 2 with an 80% limit synced from the app:

Charge Limit row in System Information

Written with AI assistance (Claude), and running on my own Pebble Time 2 since the PR opened.

---

Aside: I am currently available for firmware contracting or full-time work. Contact: adrian@pascu.be.

### DIFF
--- a/include/pbl/services/comm_session/session_remote_version.h
+++ b/include/pbl/services/comm_session/session_remote_version.h
@@ -39,6 +39,7 @@ typedef struct PACKED {
       bool blob_db_version_support: 1;
       bool settings_sync_support: 1;  // Phone supports Settings BlobDB sync
       bool weather_db_v4_support: 1;  // Phone writes the v4 weather BlobDB record (rich forecast)
+      bool charge_limit_support: 1;  // Watch supports the chargeLimitPct setting
     };
     uint64_t flags;
   };
--- a/src/fw/apps/system/settings/system.c
+++ b/src/fw/apps/system/settings/system.c
@@ -39,6 +39,7 @@
 #include "system/version.h"
 
 #include "pbl/services/activity/activity.h"
+#include "pbl/services/battery/battery_charge_limit.h"
 #include "pbl/services/blob_db/api.h"
 #include <pbl/logging/logging.h>
 
@@ -56,6 +57,7 @@ enum {
   SystemInformationItemHardware,
   SystemInformationItemSerial,
   SystemInformationItemUptime,
+  SystemInformationItemChargeLimit,
   SystemInformationItemLegal,
   SystemInformationItem_Count,
 };
@@ -123,6 +125,7 @@ typedef struct SystemInformationData {
   char serial_string[MFG_SERIAL_NUMBER_SIZE + 1];
   char hw_version_string[MFG_HW_VERSION_SIZE + 1];
   char uptime_string[16]; // "xxd xxh xxm xxs"
+  char charge_limit_string[8];
   char const * subtitle_text[SystemInformationItem_Count];
   char language_string[16];
 } SystemInformationData;
@@ -209,6 +212,7 @@ static const char* s_information_titles[SystemInformationItem_Count] = {
   [SystemInformationItemHardware] = i18n_noop("Hardware"),
   [SystemInformationItemSerial] = i18n_noop("Serial"),
   [SystemInformationItemUptime] = i18n_noop("Uptime"),
+  [SystemInformationItemChargeLimit] = i18n_noop("Charge Limit"),
   [SystemInformationItemLegal] = i18n_noop("Legal")
 };
 
@@ -306,6 +310,12 @@ static void prv_information_window_push(SettingsSystemData *data) {
   sniprintf(info->language_string, sizeof(info->language_string),
             "%s, v%u", i18n_get_locale(), i18n_get_version());
 
+  const uint8_t charge_limit_pct = shell_prefs_get_charge_limit_pct();
+  if (charge_limit_pct != CHARGE_LIMIT_PCT_DISABLED) {
+    sniprintf(info->charge_limit_string, sizeof(info->charge_limit_string),
+              "%u%%", charge_limit_pct);
+  }
+
   info->subtitle_text[SystemInformationItemBtAddress]  = info->bt_mac_addr;
   info->subtitle_text[SystemInformationItemFirmware]   =
     (char*) (strlen(TINTIN_METADATA.version_tag) >= 2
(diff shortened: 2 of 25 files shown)


include/pbl/services/comm_session/session_remote_version.h
```diff
@@ -36,6 +36,8 @@ typedef struct PACKED {
       bool continue_fw_install_across_disconnect_support: 1;
       bool blob_db_version_support: 1;
       bool settings_sync_support: 1;  // Phone supports Settings BlobDB sync
+      bool weather_db_v4_support: 1;
```
adipascu, line 39: Placeholder so `charge_limit_support` lands on bit 25, matching the mobile app's `SupportsWeatherDbV4(24)`. The firmware never sets it.

gmarull:
Note that on partial charges we have a known fuel gauge algorithm accuracy problem, pending discussion with Nordic. So let's hold this until vendor clarifies if there's anything wrong on our side, or the fuel gauge lib (closed blob) has problems.

adipascu:
@gmarull I have been running this on my own Time 2 for a month: 14 days under the limiter with no `CHARGE_STATE_COMPLETE`, then a deliberate full charge, then another 17 days unanchored.

On whether the pack needs to reach 100 % periodically to keep the gauge calibrated: it does not. The first anchor, after 14 days without one, corrected reported SOC by 1.0 pp. In the 17 days since, with no anchor at all, the same resting voltage still maps to the same SOC within +0.17 pp (sd 0.38, n = 30 voltages, rested samples only).

Reported SOC follows both cell voltage and integrated current, on different timescales. Within a single 3.4-day discharge run, SOC keeps falling while voltage sits on one 5 mV step, by up to 3.8 pp, so current is being integrated. Over weeks the voltage term bounds the coulomb term, so the error does not accumulate without a full-charge reference.

Anything measured before #1852 fed the gauge inputs quantised 4x coarser than design.

The 0.5C charge current change is also in this PR, running on the same watch.

Charts and method: https://claude.ai/code/artifact/58d02ef7-d87a-4654-8d02-758ee1520da7
CSV, 3,678 samples over 31 days: https://github.com/peblum/Peblum/releases/download/pr-assets/battery-series-2026-08-08-to-09-07.csv

src/fw/apps/system/settings/system.c
```diff
@@ -207,6 +210,7 @@ static const char* s_information_titles[SystemInformationItem_Count] = {
   [SystemInformationItemHardware] = i18n_noop("Hardware"),
   [SystemInformationItemSerial] = i18n_noop("Serial"),
   [SystemInformationItemUptime] = i18n_noop("Uptime"),
+  [SystemInformationItemChargeLimit] = i18n_noop("Charge Limit"),
```
gmarull, line 213: this seems redundant, there's already a setting for that (even if on the phone only)

include/pbl/services/comm_session/session_remote_version.h
```diff
@@ -39,6 +39,7 @@ typedef struct PACKED {
       bool blob_db_version_support: 1;
       bool settings_sync_support: 1;  // Phone supports Settings BlobDB sync
       bool weather_db_v4_support: 1;  // Phone writes the v4 weather BlobDB record (rich forecast)
+      bool charge_limit_support: 1;  // Watch supports the chargeLimitPct setting
```
gmarull: nack, settings sync should handle that (if not already, @jplexer may be more familiar)

src/fw/board/boards/board_obelix.c
```diff
@@ -538,8 +538,8 @@ const BoardConfigActuator BOARD_CONFIG_VIBE = {
 
 // TODO(OBELIX): Adjust to final battery parameters
 const Npm1300Config NPM1300_CONFIG = {
-  // 190mA = 1C (rapid charge, max limit from datasheet)
-  .chg_current_ma = 190,
+  // 96mA = 0.5C standard charge (datasheet rapid-charge max is 190mA = 1C)
+  .chg_current_ma = 96,
```
gmarull: nack, add new setting to choose between normal/fast charge instead

gmarull (changes requested):

src/fw/apps/system/settings/system.c
adipascu: This is a read-only row in System > Information, not a setting. Still want it removed?

include/pbl/services/comm_session/session_remote_version.h
adipascu: Dropped. The app gates on firmware version instead, same as the HRM prefs.

src/fw/board/boards/board_obelix.c
adipascu: Moved to separate PR: #2052 and coredevices/mobileapp#419.

PR #1848, opened 2026-08-10
drivers/speaker: apply volume before un-muting the DAC path

bemyak:
I was adding hourly "POST success" beep into my [Nostalgic Commander](https://apps.repebble.com/nostalgic-commander_d19443a35c6240ccaca492f5) watchface, and I noticed that on lower volume settings it's not just BOOOOOOP, it's BE-BOOOOP. Some investigation revealed that the driver applied volume gain AFTER sound has already started playing. This PR fixes the bug.


jplexer (approved):
did a commit on top that fixes the case where volume is supposed to be muted, otherwise lgtm!

PR #1855, opened 2026-08-12
font_resources: Gracefully handle styled Latin codepoints.

agausmann:
Motivated by poor readability of notifications that contain Latin characters from the Unicode blocks "Mathematical Alphanumeric Symbols," "Letterlike Symbols," and "Halfwidth and Fullwidth Forms." These characters are sometimes used as a stylistic choice in user display names and channel names on messaging platforms like Discord, but they do not have glyphs assigned to them.

Before this change, these codepoints would render as the block fallback, making them unreadable. This change handles these codepoints by mapping them to corresponding Latin codepoints if the glyph lookup fails for the original codepoint. By using the plain Latin glyphs, they are at least rendered in a readable way.

One possible alternative is embedding this mapping inside the font data, so that these codepoints map directly to the same glyphs as the plain Latin codepoints. That may be a better solution if possible; I was just more familiar with this part of the code.

This is also not a complete list of "styled" codepoints as I call them. These are just the ones I've encountered myself, but there are others for other languages/scripts, and there may also be some Latin ones that I missed. This system can be extended to handle those in the future, if desired.

### DIFF
--- a/src/fw/applib/graphics/text_resources.c
+++ b/src/fw/applib/graphics/text_resources.c
@@ -628,6 +628,149 @@ static const GlyphData *prv_get_glyph_in_font(FontCache *font_cache, Codepoint c
   return data;
 }
 
+// Convert any "styled" codepoints into their correspnding plain codepoints. Currently known styled
+// codepoints are in these blocks:
+//
+// - Mathematical Alphanumeric Symbols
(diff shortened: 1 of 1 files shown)


agausmann:
that should fix the lint

PR #1861, opened 2026-08-12
bluetooth-fw/nimble: reset GATT pool-sizing counters on stack stop

jplexer:
https://github.com/apache/mynewt-nimble/issues/896

Fixes FIRM-3741


gmarull (changes requested):
we need to patch nimble, this is too hacky

Issue #1865, opened 2026-08-13
app_touch_navigation_enable(true) faults inside firmware on the first touch (4.33.1, Time 2)


mrtruebeliever:
### Is there an existing issue for this?

- [X] I have searched the existing issues

I found #1826 (touch navigation vs. backlight in 4.32.0, closed) and #1578 / #1863 (feature requests), but nothing about an app fault caused by the opt-in itself.

### Current Behavior

Any third-party watchapp that calls `app_touch_navigation_enable(true)` is killed on the **first touch** of the screen. The app faults inside firmware:

```
[13:36:06] pebbledoist.c:73> probe A: init entered
[13:36:06] pebbledoist.c:94> probe B: window pushed, subscribing
[13:36:06] pebbledoist.c:96> probe C: subscribed, enabled=1
[13:36:06] pebbledoist.c:99> probe D: bridge opt-in returned
… first screen touch …
[13:36:09] ault_handling.c:105> App fault! {69de8701-3317-4381-bb1b-9760d6a14a19} PC: 0x120ffa68 LR: 0x12142e83
```

The PC/LR pair is **identical across two unrelated apps** (different UUIDs, different code, different window layouts) and identical across every run, and both addresses lie outside the app's own ELF (the app is reported as `???` by `pebble logs`), so the fault is in firmware code running on the app task, not in app code.

What I ruled out on-device, each time with the same fault:

- **Window type / touch target** — crashes on a `MenuLayer` window with rows under the finger, on a window whose only content is a custom `Layer`, and on a tap anywhere on screen (header strip, list rows, empty area).
- **Call ordering** — crashes both when `app_touch_navigation_enable(true)` is called before the first `window_stack_push()` and when it is called after a window is already on the stack.
- **Missing touch subscription** — crashes with and without a live `touch_service_subscribe()`. `touch_service_is_enabled()` returns 1 at that point.
- **App-side handler being reached** — a subscribed `TouchServiceHandler` that logs every `TouchEvent_Touchdown` never fires before the fault.

Not affected: apps that never call `app_touch_navigation_enable()`. Three of my apps without the opt-in — including one that subscribes to raw touch events with `touch_service_subscribe()` — survive any amount of touching on the same watch and firmware, so touch delivery as such is healthy; only the navigation bridge faults.

Confirmed on two apps so far (a Todoist client and a breathing coach, nothing exotic — each does no more than the single documented opt-in call in `init()`).

### Expected Behavior

[...]

PR #1866, opened 2026-08-14
applib/recognizer: read nav gates and ticks through syscalls

loxK:
Third-party apps that call `app_touch_navigation_enable(true)` are killed on their first touch, as reported in #1865. The custom-recognizer route documented in `docs/reference/recognizers.md` is unusable for the same reason.

## Root cause

`touch_nav_dispatch()` lives in `applib/` and runs on the app task, which is unprivileged for third-party apps: `prv_init_from_info_common()` sets `is_unprivileged = true` on every app loaded from its `PebbleProcessInfo` header. Its first statement read the touch service's nav gates directly, taking the kernel-owned mutex over kernel-owned statics:

```
PC 0x120ffa68 -> touch_nav_enabled    src/fw/services/touch/touch.c:86
LR 0x12142e83 -> touch_nav_dispatch   src/fw/applib/ui/recognizer/touch_nav.c:459
```

(symbolized against the `firmware_obelix_pvt_v4.33.1_slot0.elf` release artifact; the DVT artifact resolves the same addresses to two functions with no call relationship, which is what identifies PVT as the right one)

Built-in system apps are constructed from a static `PebbleProcessMd` where the flag stays false, so they are privileged and never hit this. That is why the whole system UI scrolls by touch while only third-party apps die.

The recognizers had the same problem one level down. `tap.c`, `pan.c`, `swipe.c` and the `menu_layer` double-tap window all called `rtc_get_ticks()`, which reaches `xTaskGetTickCount()`. That fault only surfaces once the first one is fixed:

```
PC 0x120d428e -> xTaskGetTickCount    FreeRTOS tasks.c:1777
LR 0x1203d771 -> rtc_get_ticks        src/fw/drivers/rtc/sf32lb.c:223
```

## Fix

Two new syscalls, `sys_touch_nav_enabled()` and `sys_touch_app_nav_active()`, mirroring the existing `sys_touch_service_is_enabled()` right next to them, plus nine `rtc_get_ticks()` call sites routed through the existing `sys_get_ticks()`.

No central registration is involved: `DEFINE_SYSCALL` emits a self-contained trampoline into its own `.syscall_text.<name>` section.

Six unit tests stubbed `touch_nav_enabled` / `touch_app_nav_active` under their old names and are renamed to match. `fake_rtc.c` already provides `sys_get_ticks()`, so the tick change needed no test changes.

## Why the recognizer route was inert

`recognizer_manager_handle_touch_event()` has exactly two call sites, `touch_nav.c:496` and `:527`, both downstream of the faulting line. So there was no reachable state for a third-party app: without the opt-in the recognizers were never fed, and with it the dispatcher faulted before it could feed them. `window_attach_recognizer()` and `window_set_touch_bridge_disabled()` become usable again with this change.

## Verification

- `./pbl test` green.
- QEMU `pebble-emery`, third-party app calling `app_touch_navigation_enable(true)`: tap on a plain `Layer` window opens the next window, vertical swipe scrolls a `MenuLayer` at pixel granularity, double-tap selects a row. No `App fault!` across the sequence; before the change the first touch was fatal.
- Obelix PVT hardware running `v4.33.1` plus this patch: touch works in the same app.

One thing I could not characterize: a horizontal swipe produced no BACK on QEMU. There is no baseline to compare against, since the app died on the first touch before this change, so I am not claiming it as either working or regressed.

### DIFF
--- a/src/fw/applib/ui/menu_layer.c
+++ b/src/fw/applib/ui/menu_layer.c
@@ -35,6 +35,7 @@
 #include "applib/ui/scroll_layer_private.h"
 #include "kernel/pebble_tasks.h"
 #include "pbl/drivers/rtc.h"
+#include "syscall/syscall.h"
 
 // The per-task touch-nav state lives in the app state (app task) or the modal manager (KernelMain).
 // Forward-declared here to avoid pulling those heavy kernel/app headers into this applib module.
@@ -1802,7 +1803,7 @@ void menu_layer_touch_handle_tap(MenuLayer *menu_layer, GPoint point_on_screen) 
     return;
   }
 
-  const RtcTicks now = rtc_get_ticks();
+  const RtcTicks now = sys_get_ticks();
   const RtcTicks window_ticks = (RtcTicks)DOUBLE_TAP_WINDOW_MS * RTC_TICKS_HZ / 1000;
 
   // Priority 1 — fast double tap: a second tap within the window after a tap-select activates the
--- a/src/fw/applib/ui/recognizer/pan.c
+++ b/src/fw/applib/ui/recognizer/pan.c
@@ -8,6 +8,7 @@
 
 #include "pbl/drivers/rtc.h"
 #include "pbl/util/math.h"
+#include "syscall/syscall.h"
 
 #include <string.h>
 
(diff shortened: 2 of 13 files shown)


jplexer (approved):
thanks!

jplexer (approved):

PR #1867, opened 2026-08-14
applib/touch: let watchfaces ride an already-powered touch sensor

loxK:
Watchfaces cannot read touch at all: `prv_get_state()` returns NULL for them, so `touch_service_subscribe()` is a no-op and the sensor is never powered on their behalf (the early return precedes `sys_touch_set_raw_subscribed()`).

This lets a watchface subscribe to raw touch, but only in a way that cannot cost anything.

## The guarantee

A watchface subscription is **passive**: it is counted in `s_subscriber_count` but excluded from the count that powers the sensor. It is delivered whatever the sensor already produces and never turns it on.

So a watchface reads touch exactly when something else already keeps the sensor up, and reads nothing otherwise:

| Wake on touch | Touch navigation | Watchface |
|---|---|---|
| Off | Off | receives nothing, sensor stays off |
| Tap / Double Tap | Off | receives |
| Off | On | receives |

Verified on obelix PVT hardware with a watchface calling `touch_service_subscribe()`, toggling both prefs.

The point is that this is structural rather than a convention: the code cannot power the sensor for a watchface, so there is no configuration in which the feature costs battery. On a watch where either pref is on, the CST816 already runs continuously (`touch_set_backlight_enabled` and `touch_set_system_hold` each take a permanent hold), and a watchface subscription only joins the existing refcount.

## Scope, and what to look at

This touches the sensor refcount, which is shared by every touch user (the backlight holder, the nav hold, all apps), so it deserves a careful read rather than a glance at the line count:

- `prv_add_subscriber_cb` / `prv_remove_subscriber_cb` now split the count. The passive decision is recorded per task in `s_passive_subscriber_tasks` rather than recomputed on removal, because by the time the subscription dies the task may no longer be the running watchface, and the remove path has to unwind the accounting the add path chose.
- `touch_service_set_globally_enabled()` re-derives the sensor state from the subscribers, so it takes the powering count too. Otherwise cycling the master Touch pref would power the sensor up for a watchface that is the only subscriber left.
- `touch.c` now includes `process_management/app_manager.h`. `touch_session.c` already does this, so there is precedent, but it is a service reaching into process management.

Three tests cover the accounting: a watchface alone never powers the sensor and the count unwinds so a later watchapp still does; a system holder leaving powers the sensor down even though the watchface is still subscribed; and cycling the global kill switch does not power it up for a passive subscriber.

## What is deliberately not changed

The system slot stays gated for watchfaces, so the Tier-2 button bridge remains a watchapp feature: a watchface drives its own UI from raw touch rather than having buttons synthesized under it. Whether watchfaces should be able to opt into the bridge is a separate question, and a separate PR.

The interaction session is untouched, so on an unarmed idle watchface the events still carry `non_navigational`. That is what lets a watchface ignore a pocket touch rather than act on it, and it is a better tool for that than a blanket block.

### DIFF
--- a/src/fw/applib/touch_service.c
+++ b/src/fw/applib/touch_service.c
@@ -14,16 +14,14 @@
 #include "system/passert.h"
 
 //! @return the per-task touch service state, or NULL if the current task is
-//! not permitted to use the touch service (e.g. background workers, or
-//! watchfaces). Callers must no-op when this returns NULL.
+//! not permitted to use the touch service (e.g. background workers). Callers
+//! must no-op when this returns NULL. Watchfaces reach the raw slot through
+//! here; the system slot is gated separately in
+//! touch_service_set_system_handler().
 static TouchServiceState *prv_get_state(void) {
   PebbleTask task = pebble_task_get_current();
   switch (task) {
     case PebbleTask_App:
-      // Touch is reserved for watchapps; watchfaces must not consume it.
-      if (sys_app_is_watchface()) {
-        return NULL;
-      }
       return app_state_get_touch_service_state();
     case PebbleTask_KernelMain:
       return kernel_applib_get_touch_service_state();
@@ -67,6 +65,11 @@ static void prv_update_subscription(TouchServiceState *state) {
 }
 
 void touch_service_set_system_handler(TouchServiceHandler handler, void *context) {
+  // The Tier-2 bridge is a watchapp feature: a watchface drives its own UI from
+  // raw touch.
+  if ((pebble_task_get_current() == PebbleTask_App) && sys_app_is_watchface()) {
+    return;
+  }
   TouchServiceState *state = prv_get_state();
   if (!state) {
     return;
--- a/src/fw/services/touch/touch.c
+++ b/src/fw/services/touch/touch.c
@@ -8,6 +8,7 @@
 #include <pbl/drivers/touch/touch_sensor.h>
 #include "kernel/events.h"
 #include "kernel/pebble_tasks.h"
+#include "process_management/app_manager.h"
 #include "pbl/services/event_service.h"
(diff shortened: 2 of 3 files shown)


jplexer:
We are still working out how to give watchfaces touch access, sorry!

loxK:
No problem, that makes sense. Two findings from building this that might be useful input to whatever design you land on, then I'll leave it there.

**The power argument may be weaker than it looks.** I assumed the block was about battery, since a watchface is on screen almost all the time and `touch_service_subscribe()` powers the sensor. But three holders already keep the CST816 running independently of any app: `touch_set_backlight_enabled()` whenever Wake on touch is not Off, `touch_set_system_hold()` whenever the touch navigation pref is on, and any raw app subscription. Two of those are system settings that are on by default, so on a typical watch the sensor is already up continuously and a watchface subscription only joins the existing refcount.

**A watchface subscription can be made structurally free.** Rather than gating on whether a holder exists at subscribe time (a snapshot that goes stale if the user turns the setting off later), the subscription can be excluded from the count that powers the sensor. It is then delivered whatever the sensor already produces and can never turn it on. Measured on obelix PVT:

| Wake on touch | Touch navigation | Watchface |
|---|---|---|
| Off | Off | receives nothing, sensor stays off |
| Tap / Double Tap | Off | receives |
| Off | On | receives |

The property that made it worth the extra plumbing is the third row of the failure case: with a snapshot gate, a watchface that got in keeps the sensor alive by itself once the setting is switched off. With the count split, the sensor goes down and the watchface simply stops receiving.

The other thing worth separating is that `prv_get_state()` gates both slots at once, so lifting it also hands watchfaces the Tier-2 button bridge. Those are two different decisions, and the raw slot is the one that does not need a policy debate.

Happy to test anything on hardware when you get to it.

PR #1878, opened 2026-08-18
Music: update icons for seek behaviour

sjp4:
Update the music app to show icons which match mobile app behaviour, where
it will seek instead of skipping when playing podcasts.

Carry what the commands will do in a flag bit in a byte trailing
MusicEndpointPlayStateInfo, and pick the action bar icons from it. The byte
is optional in both directions: phone apps that predate it send the shorter
message, and watches that predate it stop reading after the fields they know.

Music vs podcasts screenshots:

pebble_screenshot_2026-08-17_20-20-41
pebble_screenshot_2026-08-17_20-10-33

Mobile branch: https://github.com/coredevices/CoreApp/tree/steve/podcast-seek


jplexer (approved):

ericmigi:
I like this!

PR #1879, opened 2026-08-18
fw/music: keep the normal clock when album art is not shown

ericmigi:
## Problem

MOB-11802: the music app's status-bar clock is unexpectedly large since v4.34, even when album art isn't showing.

The album-art work set the status bar to the big Large-Bold clock in **every** mode (`prv_apply_art_appearance`):

```c
status_bar_layer_set_mode(&data->status_layer, data->has_album_art ?
                          StatusBarLayerModeClockLargeBoldOutlined : StatusBarLayerModeClockLargeBold);
```

The `else` branch (`StatusBarLayerModeClockLargeBold`) applies whenever there's no art — including every rectangular watch with the feature off or no cover available. Before the feature, the music app never called `status_bar_layer_set_mode`, so it used the default `StatusBarLayerModeClock` (normal size).

## Fix

Tie the clock size to the layout, not just to art:

- **stock layout** (rectangular, no art) → normal `StatusBarLayerModeClock` (pre-v4.34 behaviour)
- **media layout + art** → `StatusBarLayerModeClockLargeBoldOutlined` (outlined over the cover)
- **media layout, no art** (round `MUSIC_ROUND_MEDIA_LAYOUT`) → `StatusBarLayerModeClockLargeBold`

Round displays always use the media layout, so their big clock is unchanged; only rectangular no-art returns to the normal clock.

## Tests

Regenerated the rectangular music goldens (asterix + obelix), which had captured the oversized clock. gabbro (round) goldens are unchanged. `./pbl test -M '.*music.*'` passes.

Fixes MOB-11802


ericmigi:
tested, fixes 
image

jplexer (approved):

PR #1886, opened 2026-08-18
build: de-waf the post-link pipeline (phase 0 + 1)

jplexer:
First steps toward making waf a thin orchestrator over standalone tooling to later switch it out with another build system. No behavior changes intended; artifacts were verified byte-identical where deterministic.

### Cleanup
- Drop three dead waf tools (`binary_header`, `c_inject_include_files`, `compress`). They were unused, and two had latent import bugs.
- Fix two latent `pebble_test.py` defects: mutable default args accumulating libs across all 285 `clar()` call sites, and a loop-leaked `test_source` that would drop files if a test glob ever matched more than one.
- Deduplicate the FirmwareDescription helpers: canonical copy now lives in `pebble.commander.util.fw_binary_info` (the live flash-time path for non-pblboot boards, i.e. asterix); the `tools/` scripts are thin shims keeping their `.elf` support and CLIs.

### build-info.json
`configure` now writes `build/build-info.json`: board, platform, variant, slot, runners, artifact paths, and all `CONFIG_*` symbols. This is the neutral interface the extracted tools read: any future build system that writes the same file drives the same tools unchanged.

### Extracted tools (waf calls them; nothing about them is waf-specific)
- `tools/fw_image.py`: post-link pipeline: objcopy hex/bin, pblboot header, loghash extraction/check. Replaces five waf rules + a post-build merge.
- `tools/make_fw_bundle.py`: .pbz bundling
- `tools/make_qemu_images.py`: QEMU micro/SPI flash images
- `tools/gitinfo.py`: standalone git version info with `--json`.
- Firmware/resource size caps move from hardcoded wscript tables to Kconfig (`FW_MAX_SIZE`, `SYSTEM_RESOURCES_MAX_SIZE`); verified identical for every board × variant combination.

`./pbl` no longer parses waf's c4che or spawns waf for bundle/qemu images: it reads build-info.json.

### Verification
- objcopy outputs, loghash dicts, bundle contents and QEMU image layouts byte-identical to the replaced waf rules for the same inputs.
- Full configure/build/bundle/test cycle green; unit tests: 328 suites, 0 failures.


tools/fw_image.py
```diff
@@ -0,0 +1,106 @@
...
+    ".bss",
+    "-R",
+    ".retained",
+]
+
+
+def _objcopy(elf, out, fmt):
+    subprocess.check_call(
+        ["arm-none-eabi-objcopy"] + OBJCOPY_STRIP_ARGS + ["-O", fmt, elf, out]
+    )
```
gmarull, line 45: note: on a build system based e.g. on CMake, having this outside of the build system will be non-idiomatic, and probably cause problems like dependencies not being tracked properly (to determine when something has to be re-run). Also, we may not always want to invoke such tool, what if we support one day native builds, llvm, etc.?

Kconfig
```diff
@@ -34,6 +34,26 @@ config PBLBOOT
...
+
+config SYSTEM_RESOURCES_MAX_SIZE
+    hex
+    default 0x100000 if SOC_NRF52
+    default 0x200000 if SOC_SF32LB52
+    default 0x200000 if SOC_QEMU
+    default 0x40000
+    help
+      Maximum size of the system resources pbpack, enforced after the
+      resource build.
```
gmarull, line 55: we should define some sort of partition map at board level for that I think

PR #1902, opened 2026-08-21
fw/light: give each dynamic backlight mode its own intensity floor

jplexer:
In a dark room the dynamic backlight ramp bottomed out at a fixed 10% regardless of mode, so switching between the Standard and Battery Saver presets produced no visible brightness change unless Max Brightness was selected. Give each dynamic mode a distinct 0-lux floor (Bright 30%, Standard 20%, Dim 10%) so the presets are distinguishable in the dark.

Fixes FIRM-3993


ericmigi:
@jplexer tested, like it!

PR #1906, opened 2026-08-22
notifications: play a sound when a notification arrives

Coral-coder:
Because someone on Reddit wanted it, and I kinda did too… 🤷🏼‍♀️

Adds an optional notification sound for speaker-equipped watches, rendered through the existing note-sequence synth — no PCM assets, a few hundred bytes of note tables. Off by default; picker lives in Settings → Sounds & Haptics and previews each choice on select.
The four tones (Ping, Doorbell, Trill, Ascent) are deliberately quite basic — single-voice chirps meant to prove out the plumbing rather than be the final sound design. Happy to take direction on better melodies; each is ~5 lines of note table.
Behavior: DND follows the vibe mask, the vibe holdoff is shared so a notification storm doesn’t chirp continuously (both gates evaluated before either fires, so a notification’s own vibe can’t suppress its own sound), and speaker mute/volume are enforced by the speaker service. Unlike vibes, sounds still play on the charger, where a docked watch can usefully chirp.
The last commit (third_party/hal_sifli: -Wno-maybe-uninitialized) only matters for building with stock GCC 13 outside the docker image — drop it if unwanted.
Tested on Pebble Time 2 (obelix@pvt); full test suite passes.


gmarull:
@ericmigi

ericmigi:
probably should be in settings -> notifications

PR #1909, opened 2026-08-22
ci: pass --ref main when dispatching release-QA (fine-grained PAT fix)

ericmigi:
Fixes the `dispatch-release-qa` failure seen on the first run:

```
unable to determine default branch for coredevices/unicorn:
GraphQL: Resource not accessible by personal access token (repository.defaultBranchRef)
```

`gh workflow run` **without** `--ref` resolves the default branch via a **GraphQL** query (`repository.defaultBranchRef`), and **fine-grained PATs can't use that query** — so it fails even though `UNICORN_DISPATCH_TOKEN` is correctly scoped (`coredevices/unicorn`, Actions: Read/Write, Metadata: Read).

Passing **`--ref main`** skips the default-branch lookup and uses the REST workflow-dispatch endpoint, which the token *can* use. One-line change.

### DIFF
--- a/.github/workflows/build-firmware.yml
+++ b/.github/workflows/build-firmware.yml
@@ -225,7 +225,10 @@ jobs:
           fi
           # to_pbz_run_id = this Build Firmware run (its qa-firmware-<board> artifact is the "to").
           # coreapp_run_id omitted -> unicorn uses the most recently tagged coreapp release.
-          gh workflow run one-off-pebbleos-test.yml -R coredevices/unicorn \
+          # --ref main is required: without it gh resolves the default branch via a GraphQL query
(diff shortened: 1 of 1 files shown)


ericmigi:
@jplexer it worked! now each PR by core team member will kick off a real test on hw

gmarull:
Note I think we should not run on hw for every PR, many are trivial and do not require such level of testing. Also I've never seen hw in the loop working reliably, so making it a CI gate should be avoided. I'd suggest to do this:

- run on manually added label (sometimes we know a PR deserves it)
- run on a nightly basis, and in case of failure, maybe try a bisection to isolate the problem and report an automatic bug

Issue #1911, opened 2026-08-24
System menus ignore the user's Text Size setting


jhampton:
### Summary

Settings → Notifications → Text Size writes the global `textStyle` preference via `system_theme_set_content_size()`, but system menus never read it. Setting "Larger" changes notifications only; the launcher, Settings, and every other menu render at the platform default.

This is the most-liked topic on the forum ([#561](https://forum.repebble.com/t/pebble-time-2-request-larger-system-font/561), 52 posts), with several PT2 owners reporting they cannot read system menus without glasses.

The developer docs already describe this as incomplete rather than intentional:

> At present the Text Size setting only affects notifications and some system UI components, but other system UI components will be updated to support ContentSize in future versions.
> — https://developer.repebble.com/guides/user-interfaces/content-size/

### Causes

1. **Menu cells read the compile-time default, not the user's value.** `prv_get_cell_title_font()` / `prv_get_cell_subtitle_font()` call `system_theme_get_font_for_default_size()` (`src/fw/applib/ui/menu_layer_system_cells.c:83-94`), which hardcodes `PreferredContentSizeDefault` (`src/fw/shell/system_theme.c:168-171`). Cell dimensions do the same via `prv_get_dimensions_for_runtime_platform_default_size()` (`menu_layer_system_cells.c:63-66`).

2. **The `ExtraLarge` tier is unfinished.** Its cell dimensions are identical to Large (`menu_layer_system_cells.c:53-60`) and most of its font rows are marked `//! @note this is the same as Large until ExtraLarge is designed` (`system_theme.c:91-117`). Emery already defaults to Large, so on a Pebble Time 2 "Larger" resolves to a copy of the current tier — it would be a no-op even if (1) were fixed.

3. **No invalidation on change.** Nothing relayouts when the preference changes; `menu_layer` caches cell heights. Only the QEMU handler works around this, by closing the foreground app (`src/fw/drivers/qemu/qemu_serial.c:150-154`).

Separately, `src/fw/apps/system/settings/settings.c:104` hardcodes a 37px row height (`// FIXME: hardcoding as settings menu is "special"`), which already clips below the 44px Medium default and will clip worse at any larger tier.

### Proposed scope

Splitting this so the first change stays reviewable:

**This issue / first PR — system menus**
- Design the `ExtraLarge` font and cell-dimension rows (uses existing `GOTHIC_28`/`28_BOLD`/`36`/`36_BOLD` resources — no new font resources, no language-pack changes)
[...]

PR #1502, opened 2026-06-12
Draw action bar on left or right depending on screen orientation

deadbaed:
Fixes #1464 and #898, there is also a [forum post](https://forum.repebble.com/t/align-on-screen-choises-with-actual-button-location-in-left-handed-mode/771) discussing this issue.

## AI

I used AI when doing research and during development, I used the `Big Pickle` model from [OpenCode](https://opencode.ai/docs/zen). The model wrote most of the code, the code looks good enough for me, but I'll let the maintainers judge it. Please let me know if it is wrong.

## QEMU targets and drivers

I tested only on QEMU targets (mainly on `qemu_emery` since I own a Pebble Time 2). I also made sure it compiles and looks okay on `qemu_flint` and `qemu_gabbro`.

To get left handed mode working on QEMU platforms, I had to add `CONFIG_ORIENTATION_MANAGER` to the Kconfig files.

Since those boards did not have drivers for the accelerometer and magnetometer (by judging the linking errors and what the AI model told me), I added stub drivers for the QEMU targets.

## Development tools

I used the nix dev shell:
```
nix develop
./pbl configure --board qemu_emery && ./pbl build && ./pbl qemu
```

## Visual results

The UI looks good for the most part when I looked at applications and inside the settings.

~~There are 2 exceptions, the `Factory reset` prompt and the `Music` app~~ this got fixed, see the comments below:

Screenshot 2026-06-12 at 09 46 57

Screenshot 2026-06-12 at 09 44 56

I did not take the time to investigate why. I am opening this PR to get the discussion going and to gather feedback.

Please let me know if I did something wrong or forgot to do.

Issue #1462, opened 2026-06-09
Add smaller font size for notification body


ZimbiX:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

I love my new Pebble Time 2 and being able to see much more content on screen from a notification than on my Pebble Time. This is particularly handy for groupchats, where on Pebble Time the length of the group name and author name would result in showing only a couple of words of the message.

Even so, my eyes are pretty good, and I'd love to see even more words at a glance! Can we please add the option to use an even smaller font size than the current smallest of 'Smaller'? With 'Smaller', the notification title is in a smaller font (though in bold) than the notification body - I'd like to have the body match that font size (but non-bolded) which I think would match the font bitmap of the Pebble Time (although slightly smaller in millimetres, given the [slightly greater PPI on Pebble Time 2](https://developer.rebble.io/guides/tools-and-resources/hardware-information/)).

### Alternatives

Getting worse eyesight? =P Not exactly something to recommend; but I guess it'll happen eventually!

PR #1912, opened 2026-08-24
fw: honor the user's Text Size in system menus

jhampton:
Refs #1911

## Summary

The Text Size setting (the `textStyle` pref — global, persisted, phone-synced) has existed for years but only ever affected notifications and a few components: system menu cells read the compile-time platform default instead of the user's preference, and the ExtraLarge tier that "Larger" maps to on Pebble Time 2–class displays was an unfinished copy of Large. The [developer docs](https://developer.repebble.com/guides/user-interfaces/content-size/) describe this as incomplete rather than intentional. This PR finishes that work for system menus:

1. designs the ExtraLarge menu-cell tier — fonts (`GOTHIC_28_BOLD`/`GOTHIC_28`, both already shipping in the base resource map on every board — **zero new resources**) and cell dimensions (85/52 px, derived from the existing Medium→Large growth ratios);
2. routes the menu cell font/dimension accessors through `system_theme_get_content_size()` for system processes;
3. fixes the Settings root menu's hardcoded 37 px row height (a tagged FIXME that already clips the 44 px Medium-tier cell content today);
4. moves the Text Size row from Settings → Notifications to Settings → Display (next to Language), since the preference is global — its old home is why forum users report the setting as nonexistent (open to a different placement);
5. makes the change apply live: watch-side pref writes now emit `PEBBLE_PREF_CHANGE_EVENT` (previously only phone-initiated blob-db syncs did), and the open settings menus reload their cached cell geometry and re-anchor the selection on that event.

## Third-party apps are unaffected (by construction and by test)

`menu_layer` is applib, so third-party apps draw through the same cell code. The content size is resolved per process inside `menu_layer_system_cells.c`: anything running as an app task without a system install id keeps the platform default through the exact same call chain as before (`system_theme_get_default_content_size_for_runtime_platform()` / `system_theme_get_font_for_default_size()`). This is the process-identity pattern already used by `window_long_click_subscribe` (`applib/ui/window.c`) and `dictation_session_create` (`applib/voice/dictation_session.c`). A non-golden regression test (`test_menu_layer_system_cells__third_party_app_keeps_platform_default`) asserts that a third-party install id receives the platform-default cell metrics even with the preference at ExtraLarge, and flips the gate as a positive control.

## Screenshots (`qemu_emery` — same platform as Pebble Time 2)

**Before (main):** setting Text Size to "Larger" changes nothing — the root menu screenshot at "Larger" is pixel-identical to Default (verified with an image diff). That no-op is the bug.

| Before — Settings root (Default) | Before — Settings root ("Larger" — identical) | Before — Text Size lives under Notifications |
|---|---|---|
|  |  |  |

**After (this PR):**

| Tier | Settings root | Settings → Display |
|---|---|---|
| Smaller (Medium) |  |  |
| Default (Large) |  |  |
| Larger (ExtraLarge) |  |  |

New Text Size chooser location and a content submenu at Larger:

| Text Size under Display | Notifications submenu at Larger |
|---|---|
|  |  |

Note on the "before vs after at Default" root menu difference: the root menu grows from 37 px rows to the standard cell height — that is the FIXME removal (item 3 above), not a side effect.

## Validation

- Full clar suite green under the CI configuration (`qemu_gabbro`), all 331 test binaries.
- New parameterized golden test rendering system cells at all four `PreferredContentSize` tiers on both rect (obelix) and round (gabbro) test platforms.
- Regenerated action-menu goldens (the action menu already followed the user size for its item font, so the redesigned ExtraLarge fonts and the now-size-following cell metrics change its rendering at non-default sizes). Every regenerated golden was reviewed image-by-image; diffs are confined to the non-default-size columns, with the platform-default column verified pixel-identical.
- Manual QEMU validation on `qemu_emery` at every tier: live relayout on change, correct highlight/selection tracking (select-probed), correct scroll re-anchoring after tier changes in both directions.
- **Not validated on physical PT2 hardware** — I don't have a device. `qemu_emery` is the same platform (200×228 rect, Large platform default), but on-device confirmation would be very welcome.

## Known limitation

The option menu itself (radio lists like the Text Size chooser) still renders at the platform default — `option_menu_window.c` reads `system_theme_get_font_for_default_size()` and is part of the follow-up sweep below.

## Explicitly out of scope (deliberate split, see #1911)

- The ~20 system screens still on `system_theme_get_font_for_default_size()` (Health, Workout, Music, About, Certifications, option menu, …) — follow-ups tracked in #1911.
- Marquee/scrolling for truncated menu titles ([forum #561](https://forum.repebble.com/t/pebble-time-2-request-larger-system-font/561) post 10) — no marquee exists in `menu_layer`/`text_layer` today.
- Status bar font/height (`status_bar_layer.c`) — menu windows inset by the compile-time `STATUS_BAR_LAYER_HEIGHT`.
- Round displays keep the fixed `MENU_CELL_ROUND_*` focused/unfocused cell constants.
- The `Small` tier design (#1462) and any new font resources (none needed).
- Handedness (#1502/#1856) — adjacent accessibility efforts, not touched.

---

Screenshots are hosted on the `pr-assets/system-text-size` branch of the fork. Driving thread: https://forum.repebble.com/t/pebble-time-2-request-larger-system-font/561


gmarull:
Thanks for this update, can you fix CI issues?

jhampton:
Sorry about that!  Fixed it up, and I’ll watch CI to make sure we’re green!

uaparit:
Tested on physical obelix_pvt hardware (Pebble Time 2). Built and sideloaded straight from this branch via ./pbl build + ./pbl bundle.

Confirms the fix: cycling Text Size (Smaller/Default/Larger) under the new Settings → Display location correctly resizes the Settings root menu live, matching the before/after screenshots in the description.

  One nice unlisted side effect: the Alarms list also picks up the fix, even though it's not mentioned as fixed or listed as a follow-up in #1911. It draws rows via menu_cell_layer_draw() with a MenuCellLayerConfig that leaves title_font/subtitle_font unset (alarms.c:296-313), and its row height comes from menu_cell_basic_cell_height() (alarms.c:221) — both fall through to the same
  prv_get_cell_title_font()/cell-dimension accessors this PR made content-size-aware in menu_layer_system_cells.c. So any screen using the generic cell-drawing helpers without an explicit font override should get this for free, beyond what's explicitly called out as fixed.

Thanks for taking this on!

gmarull (approved):

PR #1927, opened 2026-08-25
ci: dual-slot QA firmware for getafix_dvt

ericmigi:
Adds `getafix@dvt` to the `build-qa-firmware` matrix so PebbleOS main emits a dual-slot RELEASE PBZ (`qa-firmware-getafix_dvt`) for it, matching `getafix@dvt2` and `obelix@pvt`.

The unicorn iOS+getafix release-QA (daily 06:00 UTC) needs this dual-slot PBZ as the OTA "to": the per-commit `firmware-getafix_dvt` artifact is slot0-only and can't be OTA'd into the inactive slot. getafix_dvt is the board on the unicorn-mac-2 hardware rig.

### DIFF
--- a/.github/workflows/build-firmware.yml
+++ b/.github/workflows/build-firmware.yml
@@ -137,6 +137,7 @@ jobs:
       matrix:
         board:
           - obelix@pvt
+          - getafix@dvt
           - getafix@dvt2
 
     steps:


ericmigi:
Superseded by a signed-off PR from ericmigi/PebbleOS.

.github/workflows/build-firmware.yml
```diff
@@ -137,6 +137,7 @@ jobs:
       matrix:
         board:
           - obelix@pvt
+          - getafix@dvt
```
gmarull, line 140: we should not be testing dvt?

PR #1931, opened 2026-08-25
Revert "fw/services/analytics: use buffered DLS session for native heartbeat"

### LINKED ISSUE #1919: Unexpected hourly reboot on v4.36.0

### Current Behavior

I get an hourly reboot when I run v4.36.0. I first noticed this when rebasing [my fork](https://github.com/mortenfyhn/PebbleOS), and then I tested it with the binary from the release page here, and I get the same behaviour.

It's a bit easy to miss, because it reboots pretty fast. But I set a stopwatch and filmed the watch as it approached the 1 hour mark:

https://github.com/user-attachments/assets/e88582ef-267f-43e1-ad99-808b1fe15e47

### Expected Behavior

No hourly reboot.

### Steps To Reproduce

* Flash v4.36.0
* Wait an hour
* Watch reboots

### Version

v4.36.0

### Host OS

N/A

### Watch

Pebble 2 Duo (Asterix)

### Anything else?

I'm at work now but I have details and probably a fix if you give me a bit of time on my machine at home.

mortenfyhn:
This reverts commit bebc13477b0e8e2716b376275a9f403a7e8c56ee.

Fix #1919.

Tested this patch on top of v4.36.0 and it appears to work, no hourly reboot. Without it I get a consistent hourly reboot. See #1919.

So as I understand it bebc13477b0e8e2716b376275a9f403a7e8c56ee from ~#1919~ #1853 flipped the watch's native heartbeat analytics DLS session from unbuffered to buffered. The buffered ones are limited to 300 bytes (DLS_SESSION_MAX_BUFFERED_ITEM_SIZE), while the unbuffered ones are limited to 656/646 bytes (COMM_MAX_OUTBOUND_PAYLOAD_SIZE). The watch analytics size is dictated by analytics.def and is now around 570 bytes. So making it buffered hits the 300 byte limit when it tries to create the DLS session, which happens on an hourly timer, and it crashes on this assert:

https://github.com/coredevices/PebbleOS/blob/3889b0bba9fd24422bbdb997d03578186ae5fb91/src/fw/services/analytics/native.c#L311

I'm pretty new to this, so I could be getting things mixed up here, but I think that's roughly it.

There seem to be some benefits to using buffered sessions though, but that will require a bit more work.

### DIFF
--- a/src/fw/services/analytics/native.c
+++ b/src/fw/services/analytics/native.c
@@ -307,7 +307,7 @@ void pbl_analytics__native_heartbeat(void) {
     Uuid system_uuid = UUID_SYSTEM;
 
     s_dls_session = dls_create(DlsSystemTagAnalyticsNativeHeartbeat, DATA_LOGGING_BYTE_ARRAY,
-                               sizeof(struct native_heartbeat_record), true, false, &system_uuid);
+                               sizeof(struct native_heartbeat_record), false, false, &system_uuid);
     PBL_ASSERTN(s_dls_session != NULL);
   }
 


jplexer:
Sorry I didnt see this, I already reverted this myself!

uaparit:
Sorry about this one this regression traces back to my PR #1853. I flipped the session to buffered=true to get automatic retry if the initial OpenSession failed, but should have checked the record size against DLS_SESSION_MAX_BUFFERED_ITEM_SIZE (300 bytes) before merging. Should have caught this in review/testing, my mistake.

Thanks @mortenfyhn for tracking this down so quickly and clearly

mortenfyhn:
Glad it got fixed :)

PR #1932, opened 2026-08-25
fw/services/weather: expose default-location forecast to apps via weather_service_peek()

### LINKED ISSUE #555: PebbleOS weather service data should be accessible by watchfaces

### Problem:

Right now watchfaces all fetch their weather from different sources.
This needs javascript running on the phone, wasting cpu resources on the phone, accessing the internet, using an api key that gets rate limited etc.

### Solution:

Let watchfaces access the data directly, just as they can access health data and display it. 

I know this was planned looking at the PebbleOS code anyway 👅 

### Alternatives

_No response_

### Anything else?

_No response_

Techwizz-somboo:
Third-party watchfaces have no way to read the weather data the phone already syncs to the watch - the only precedent is the old PebbleKit JS AppMessage convention, which needs a JS engine most native apps don't have. weather_service_create_default_forecast() already builds almost exactly the data we want, it just has never been exposed: WeatherLocationForecast owns raw location_name/current_weather_phrase pointers into kernel memory, which can't cross the syscall boundary as-is.

Add weather_service_get_snapshot(), which copies that forecast into a new pointer-free WeatherServiceSnapshot (fixed-size phrase buffer via strncpy), and wrap it in a DEFINE_SYSCALL following the buffer-passing pattern used by sys_activity_get_minute_history: validate the caller's buffer with syscall_assert_userspace_buffer() before writing into it. The public wrapper, weather_service_peek(), lives in a new src/fw/applib/weather_service_app.{c,h}, mirroring how battery_state_service.{c,h} wraps sys_battery_get_charge_state().

WeatherServiceSnapshot's icon fields use a new WeatherSnapshotIcon enum instead of WeatherType: weather_types.h transitively includes a per-platform auto-generated resource header (for TimelineResourceId) that the SDK export tool's parser can't resolve, which is why nothing building on it was exportable before. WeatherSnapshotIcon numerically mirrors WeatherType (see weather_type_tuples.def, still the source of truth for the IDs) without depending on its header.

Registers weather_service_peek(), WeatherServiceSnapshot, and WeatherSnapshotIcon in exported_symbols.json at rev 110 and bumps PROCESS_INFO_CURRENT_SDK_VERSION_MINOR to 0x6b per docs/development/sdk_export.md.

Fixes #555

### DIFF
--- a/include/pbl/services/weather/weather_service.h
+++ b/include/pbl/services/weather/weather_service.h
@@ -11,6 +11,7 @@
 //! changed events should use the PEBBLE_WEATHER_CHANGED_EVENT (see events.h)
 
 #include "pbl/services/blob_db/weather_db.h"
+#include "pbl/services/weather/weather_snapshot_types.h"
 #include "pbl/services/weather/weather_types.h"
 #include "pbl/util/list.h"
 #include "util/time/time.h"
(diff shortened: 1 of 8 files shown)


jplexer:
hey, we are currently internally working on another way for apps to access weather data

sorry!

Techwizz-somboo:
awesome, that is great to hear, thank you!!

Techwizz-somboo:
> hey, we are currently internally working on another way for apps to access weather data
> 
> sorry!

@jplexer are you able to share any details on how the API will look yet? And any idea on when it may be available? thanks for your work on this!

PR #1934, opened 2026-08-25
applib/ui: touch scrolling affordances (scrollbar overlay + edge overscroll)

gmarull:
Two affordances that touch scrolling was missing in menus, in three self-contained pairs of commits (feature + tests each):

## Transient scrollbar overlay

`MenuLayer` shows a small scrollbar while a touch gesture (pan or inertial coast) scrolls content taller than the frame, and hides it one second after the movement stops. Button scrolling never shows it (the selection already communicates position there).

- Rect displays: 3px thumb + 1px halo on the right edge, proportional height, position clamped to the track.
- Round displays: a curved arc hugging the bezel (Wear OS style), ±30° around 3 o'clock, drawn with rounded end caps via `graphics_fill_radial` + cap circles. The full-length track is drawn behind the thumb in a subdued gray (picked by the menu background's luminance), so the thumb reads as a position within the range.
- State is carved from the existing `MenuLayer` padding (an `AppTimer *` plus flag bits), so `sizeof(MenuLayer)` and the applib-malloc budget are unchanged. Legacy2 apps excluded; `menu_layer_set_scrollbar_hidden()` opts out.

## Rubber-band overscroll with spring-back

A held pan past the scroll bounds pulls the content a little further with asymptotic damping (half the finger's speed at first, saturating at ~1/6 of the frame height); liftoff glides back to the edge in 200ms, and a cancelled gesture restores it instantly. The spring-back runs on an unclamped animation setter so it glides instead of snapping. Paging layers and content that fits the frame keep the hard clamp; center-focused menus damp past their widened bounds and return through their existing settle-to-center.

## Selection stretch at the edges

When the edge row is the selected one, the pull extends its selection background across the exposed gap to the viewport edge — the touch counterpart of the blocked-button highlight bounce, without ever painting over a neighbouring row (the neighbours move away with the content). The stretch is derived purely from the rubber-banded offset in the offset-changed handler, so the live pan, every spring-back frame, and cancel stay in sync with no extra animation state (one flag bit; struct size unchanged).

## Testing

- Full unit test suite passes; 18 new tests cover the scrollbar lifecycle/geometry (rect + thumb math), damp mapping, pan/snap/cancel overscroll on both Tier-1 widgets, and the stretch geometry/contract.
- Validated interactively on QEMU: `qemu_emery` (rect + touch: swipe vs. buttons, stretch at both edges), `qemu_gabbro` (round arc scrollbar over the carousel), `qemu_flint` (non-touch: no scrollbar, button behavior unchanged).

### DIFF
--- a/src/fw/applib/ui/menu_layer.c
+++ b/src/fw/applib/ui/menu_layer.c
@@ -20,6 +20,7 @@
 #include "shell/system_theme.h"
 #include <pbl/logging/logging.h>
 #include "system/passert.h"
+#include "pbl/util/attributes.h"
 #include "pbl/util/math.h"
 #include "pbl/util/size.h"
 #include "vibes.h"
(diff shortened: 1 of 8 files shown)


jplexer:
gonna test this on real hardware 2morrow before signing off on it

jplexer:
arghh I dont like how far away it is from the bezel and that there is no indication how far we are in the entire scroll (only tested on round so far)

gmarull:
> arghh I dont like how far away it is from the bezel and that there is no indication how far we are in the entire scroll (only tested on round so far)

let me improve round version. the rect version should be ok, I guess?

gmarull:
@jplexer pls check now

gmarull:
Ok, I've also fixed scrolling on round to be consistent with what buttons do. It all feels much more natural/polished now.

jplexer (approved):
I still think the scroll bar could be a bit closer to the edge, otherwise lgtm on both round and rect

gmarull:
@jplexer moved the bar closer to the edge (margin halved to 1px, both rect and the round arc — the arc's outer radius is now one pixel off the display edge).

gmarull:
Turns out it wasn't just the margin: the round arc was anchored to the menu frame, so windows inset by a status bar pushed it ~15px into the content. It's now anchored to the physical display circle (measured 1px off the edge on gabbro, any window framing). Amended into the last commit.

jplexer:
oh yeah this is heaps better, lgtm

PR #1937, opened 2026-08-26
resources/lang: add th_TH Thai locale (fonts)

ninefyi:
## Summary

Adds Thai (`th_TH`) as a new locale, following the pattern of the existing
`ar_SA` / `he_IL` / `zh_CN` locales. **Font resources only** — no firmware or
tooling changes.

- `thai_codepoints.json` — the 87 assigned Thai Unicode codepoints
  (U+0E01–U+0E3A, U+0E3F–U+0E5B). Thai-only, mirroring `arabic_codepoints.json`:
  mixed-script text falls back to the base Latin font via the existing depth-1
  fallback in `text_resources.c`.
- `lang_map.json` — `GOTHIC_14/18/24/28/36` (+bold) `EXTENDED` entries at the same
  pixel heights `ar_SA` uses. `BITHAM`/`ROBOTO`/`DROID_SERIF` left unchanged.
- `Sarabun-Regular.ttf` / `Sarabun-Bold.ttf` + `LICENSE` (SIL OFL 1.1). Static
  per-weight files are required because Noto Sans Thai now ships only as a
  variable font, which this pipeline cannot consume.

## Translations are intentionally not included

Per `CONTRIBUTING.md`, translated text is not submitted through pull requests.
`lang_map.json` therefore sets `strings.file: ""` and ships fonts only, as `ar_SA`
does. Thai strings should flow through Crowdin once `th_TH` is enabled there.

This does not prevent Thai from being useful: notification content arrives from the
phone and is rendered by these fonts, independent of the UI string catalog. Verified
by installing a fonts-only pack (`strings.file: ""`) and sending a Thai notification —
it renders and wraps correctly. Only system UI labels remain English until
translations land.

## Validation

- `./pbl pack_lang --lang=th_TH` builds all 10 font variants across the full
  87-codepoint set with zero glyph errors.
- Glyph budget checked against **both** `MAX_FONT_GLYPH_SIZE` limits: fits within
  256 B (flint/diorite/chalk/basalt/aplite) as well as 512 B (emery/gabbro).
- Rendered live in QEMU on all three shipping platforms:
  - `qemu_emery` (Pebble Time 2)
  - `qemu_flint` (Pebble 2 Duo — 256 B glyph limit, 1-bit display)
  - `qemu_gabbro` (Pebble Round 2 — round display / perimeter text path)
- Word wrapping verified through the real notification path. Thai has no spaces, and
  without U+200B the layout engine breaks mid-syllable and hyphenates; with U+200B at
  word boundaries it breaks cleanly. No overflow or clipping in either case.
- Mixed Thai/Latin (sender names, URLs, `15:30`) renders correctly with no wildcard
  glyphs.
- Thai glyph coverage, readability, mixed Thai/Latin rendering, and stability
  confirmed on physical Pebble Time 2.

## Known limitation

On physical PT2, the sequences `กี่` and `สี่เหลี่ยม` show tone-mark overlap. This
renderer applies no OpenType mark positioning — glyphs are placed purely by the
font's own advance metrics — so mark placement depends entirely on the font. Most
sequences tested (including tall-ascender cases such as `ปิดเครื่อง` and
`เปิดใช้ด่วน`) render cleanly; these two do not.

This PR does not claim complete Thai mark positioning for every combining sequence.
Addressing it properly means either a different font or mark-positioning support in
the renderer, which is out of scope here and better discussed separately.

## Contributor checks

- [ ] Sarabun license and attribution reviewed.
- [ ] Generated code manually reviewed.
- [ ] DCO sign-off present.
- [ ] AI assistance disclosed.


ninefyi:
Context and open questions for this PR are in #2046 — in particular whether `th_TH` can be enabled on Crowdin (this PR is fonts-only and carries no `tintin.po`), and how you'd prefer the combining-mark overlap on `กี่` / `สี่เหลี่ยม` to be handled.

gmarull:
@jplexer pls check since you've been changing languages stuff recently

PR #1607, opened 2026-06-24
Add in Blood Oxygen support

32bitx64bit:
Adds in blood oxygen support.

Related commits :
https://github.com/coredevices/pebbleos-nonfree/pull/4
https://github.com/coredevices/mobileapp/pull/267

Issue #10, opened 2025-06-09
Add display driver for LPM015M135A (JDI)


gmarull:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Proposed feature request

Obelix was merged with stub display support, also, it has been initially treated as Asterix (1-bit color depth) so we could perform UI tests using a Sharp display.

We need to add a real display driver and do any other necessary changes.

### Anything else?

Simple Sharp display driver:

https://github.com/teslabs/pebbleos/commit/218cbc233dab5b6a7619502b3ee1179c9eb9d456

![Image](https://github.com/user-attachments/assets/b4b510ae-a363-40d0-870a-48ea1bb8cb4b)

PR #1939, opened 2026-08-26
Add in Blood Oxygen support

jplexer:
Fixes some things from #1607, requires https://github.com/coredevices/pebbleos-nonfree/pull/7

Draft until that is merged


src/fw/services/hrm/hrm_manager.c
```diff
@@ -198,14 +304,36 @@ static void prv_update_hrm_enable_system_cb(void *unused) {
           hrm_is_enabled(HRM) && s_manager_state.sensor_on_since_ticks &&
           ((int64_t)(cur_ticks - s_manager_state.sensor_on_since_ticks) > unserved_timeout_ticks);
 
+      // BPM (green) and SpO2 (red/IR) sampling are each gated on their own user pref. Mask out any
+      // feature whose monitoring is disabled so a lingering subscriber for it (e.g. the BLE relay
+      // or a dormant background SpO2 session) can't light its LED or turn the sensor on. A
+      // subscriber left with no enabled features is ignored entirely.
+      HRMFeature allowed_features = (HRMFeature)~0;
+#if !defined(CONFIG_IS_BIGBOARD) && !defined(CONFIG_RECOVERY_FW)
```
gmarull: big board should not matter here

src/fw/services/hrm/hrm_manager.c
```diff
@@ -244,70 +376,30 @@ static void prv_update_hrm_enable_system_cb(void *unused) {
...
-          PBL_LOG_ERR("HRM failed to enable (attempt %d/%d)", s_manager_state.enable_failure_count,
-                      HRM_MAX_ENABLE_FAILURES);
+    // Resolve the union of wanted features down to the single optical path we can run right now.
+    const HRMFeature active_features =
+        prv_select_active_path(wanted_features, s_manager_state.active_features);
+
+    if (turn_sensor_on && active_features != 0 && !hrm_permanently_failed) {
+      if (!hrm_is_enabled(HRM)) {
+        // Sensor is off and a subscriber is due: bring it online.
+        HRM_LOG("Turning on HR sensor");
```
gmarull: use PBL_LOG_XXX

src/fw/drivers/hrm/gh3x2x.c
```diff
@@ -30,8 +31,17 @@ void gh3026_reset_pin_ctrl(uint8_t pin_level) {
...
 #ifdef CONFIG_GH3X2X_ALGO
 
-#define GH3X2X_LOG_ENABLE            0
-#define GH3X2X_FIFO_WATERMARK_CONFIG 80
+// Defined in the Goodix algo demo layer (gh3x2x_demo_algo_call_hr.c); selects the HR algorithm's
+// activity "scene" (read on every frame). Not declared in any shipped header, forward-declared
+// here.
+extern void Gh3x2xSetHbaMode(GS32 nHbaScenario);
+
+#define GH3X2X_LOG_ENABLE 0
```
gmarull, line 39: ?

src/fw/drivers/hrm/gh3x2x.c
```diff
@@ -30,8 +31,17 @@ void gh3026_reset_pin_ctrl(uint8_t pin_level) {
...
+// FIFO read batch size. In normal interrupt mode the hardware IRQ fires when this many samples
+// accumulate, so it sets the end-to-end latency and update cadence. Lower = snappier first reading
+// and finer-grained HR/SpO2 updates, at the cost of more frequent I2C bursts. 25 samples ~= 1s at
+// the 25 Hz rate (was 80 ~= 3.2s).
+#define GH3X2X_FIFO_WATERMARK_CONFIG 25
```
gmarull: not used??

src/fw/services/activity/activity.c
```diff
@@ -241,6 +271,330 @@ static void prv_heart_rate_deinit(void) {
...
+// / SpO2 data callbacks (1 Hz while the relevant sensor is on, giving precise sub-minute timing).
+static void prv_activity_spo2_update(uint32_t now_ts) {
+  ActivitySpO2ActivitySupport *as = &s_activity_state.activity_spo2;
+
+  // Run while the blood-oxygen-during-activities opt-in is on AND we're in an activity: either an
+  // auto-detected one with a running HR session (which needs HR-during-activities), or a manually
+  // started workout. The manual workout path only needs the SpO2 opt-in.
+  const bool auto_active = activity_prefs_hrm_activity_tracking_is_enabled() &&
+                           activity_algorithm_activity_hrm_is_active();
+  const bool workout_active = workout_service_is_workout_ongoing();
```
gmarull, line 514: **[code-review 1/10] Lock-order inversion (ABBA deadlock): the KernelBG HRM callbacks now take the workout_service mutex and the kraepelin algorithm mutex while hrm_manager's s_manager_state.lock is held, while the App task already takes those mutexes first and then the manager lock (start/stop_workout, frontend_opened/closed, incl. the new hrm_manager_set_activity_scene).**

Workout app open (1 s BPM app subscription => a KernelBG HRM callback every second). User taps Start: App task holds s_workout_mutex (workout_service.c:433) and blocks on s_manager_state.lock in hrm_manager_set_activity_scene (:483 -> hrm_manager.c:763) or via activity_algorithm_enable_activity_tracking -> kalg reset -> sys_hrm_manager_unsubscribe (:478 -> hrm_manager.c:855). Meanwhile KernelBG holds the manager lock in prv_system_task_hrm_handler (hrm_manager.c:465..538, callback_handler invoked at :536 under the lock) -> prv_hrm_subscription_cb -> prv_activity_spo2_update (activity.c:242) -> workout_service_is_workout_ongoing (activity.c:500 -> workout_service.c:608 prv_lock) / activity_algorithm_activity_hrm_is_active (activity.c:499 -> alg mutex). Both waits are PBL_FOREVER: KernelBG and App deadlock -> watchdog reboot. Base only had the workout->manager direction; this PR adds manager->workout and manager->alg edges (activity.c:242, 456-457, 499-500, 569). The comment at activity.c:862-866 reasons about the activity mutex but misses that the data callbacks run inside the HRM manager lock.

src/fw/services/activity/activity.c
```diff
@@ -241,6 +271,330 @@ static void prv_heart_rate_deinit(void) {
...
+      should_toggle = true;
+    }
+  }
+
+  if (should_toggle) {
+    // Don't start while HR is mid-measurement (shared optical path - see
+    // prv_heart_rate_subscription _update). HR runs first each tick so it wins ties; SpO2 takes its
+    // turn once HR finishes. Also yield to an in-progress activity-triggered SpO2 attempt.
+    if (!s_activity_state.spo2.currently_sampling &&
+        (s_activity_state.hr.currently_sampling ||
```
gmarull, line 359: **[code-review 2/10] The daily SpO2 scheduler only defers for its own daily HR window or an activity-SpO2 attempt, so when its window opens on top of any continuous 1 s green subscriber (Workout app, BLE HR relay, kalg auto-activity HR, foreground HRM app) the manager's 30 s slice valve ping-pongs the sensor with a full hrm_disable/hrm_enable + accel resubscribe per flip; SpO2 gets one ~30 s slice while its 45 s early-abort clock keeps running off-path and the window is aborted at the next minute tick with no reading.**

Daily SpO2 on (10 min) + Workout app (1 s BPM, always due since 1 s - 12 s spin-up <= 0). t0 minute tick: spo2 session -> 1 s (activity.c:337-341 only checks hr.currently_sampling / activity_spo2.phase). Manager: both paths due, green on > 30 s -> slice valve (hrm_manager.c:405-416) hands to SpO2 -> restart 1, workout HR dark. t0+30: valve hands back to green -> restart 2, ~9 s reconvergence (~40 s HR hole). SpO2 session then gets no data, so activity.c only re-evaluates at t0+60: now >= t0+45 with 0 accepted samples -> early abort (:320-321) -> OFF. Repeats every 10 min for the whole workout: 2-4 restarts, ~30 s red/IR LED, 30-40 s live-HR dropout, usually no SpO2 record. Same for kalg walk/run HR and the BLE relay; and ACTIVITY_SPO2_ACTIVITY_ATTEMPT_SEC 35 > 30 s slice bites activity-SpO2 attempts whenever a BLE relay/foreground app is due (set_sampling pauses only kalg + workout). Fix: defer the daily SpO2 window while any continuous green subscriber is due, or have the manager pause green subscribers for the SpO2 window instead of slicing.

src/fw/services/hrm/hrm_manager.c
```diff
@@ -244,70 +396,49 @@ static void prv_update_hrm_enable_system_cb(void *unused) {
...
-        s_manager_state.unserved_timeout_logged = false;
-        // Don't need the re-enable timer to fire
-        new_timer_stop(s_manager_state.update_enable_timer_id);
-        // Track HRM on-time
-        PBL_ANALYTICS_TIMER_START(hrm_on_time_ms);
+        if (prv_sensor_enable(active_features, low_latency_wanted)) {
+          // Don't need the re-enable timer to fire
+          new_timer_stop(s_manager_state.update_enable_timer_id);
+        }
+      } else if (prv_should_reconfigure(active_features, s_manager_state.active_features)) {
```
gmarull: **[code-review 3/10] wanted_features now only accumulates subscribers that are due right now (was: union of all live subscribers) and prv_should_reconfigure() restarts on any wanted != active, so an always-due 1 s BPM subscriber (BLE relay / Workout app) plus a longer-interval BPM|HRV subscriber makes the manager oscillate BPM|HRV -> BPM -> BPM|HRV with a full Goodix stop/start each time.**

BLE HR relay (ble_hrm.c:198, 1 s BPM) or Workout app + an app calling health_service_set_hrv_sample_period(60) (health_service.c:1213-1225 -> BPM|HRV @60 s). Pass 1: both due -> enable BPM|HRV. First Good BPM sample marks BOTH served (bpm_served only needs state->features & BPM, :683-689). Next pass (<=3 samples, HRM_CHECK_SENSOR_DISABLE_COUNT=3): 1 s subscriber due, HRV subscriber not due for ~48 s -> wanted=BPM != active=BPM|HRV -> prv_should_reconfigure true (:211-219) -> prv_sensor_disable()+prv_sensor_enable(BPM) (:430-436). ~48 s later HRV due -> restart to BPM|HRV; next Good BPM -> restart to BPM... Two full stop/start cycles per ~50-60 s: live 1 s HR loses up to ~12 s per restart, HRV only runs until the first Good BPM. Pre-PR the union was stable. Fix: only restart when the optical path flips or wanted has bits not already in active (keep same-path supersets running).

src/fw/drivers/hrm/gh3x2x.c
```diff
@@ -496,15 +535,29 @@ void hrm_init(HRMDevice *dev) {
...
+  // detection.
+  if (features & HRMFeature_SpO2) {
+    dev->state->work_mode = GH3X2X_FUNCTION_SPO2;
+    dev->state->is_wear = true;
+  } else {
+    dev->state->work_mode = GH3X2X_FUNCTION_HR | GH3X2X_FUNCTION_SOFT_ADT_GREEN;
+    // Clear the assumed-worn state the SpO2 path leaves behind, so the green path's own ADT is what
+    // decides. Otherwise the first HR samples after every SpO2 window report as on-wrist real BPMs
+    // (feeding the daily HR logger and the health event) even with the watch off the wrist.
+    dev->state->is_wear = false;
```
gmarull: **[code-review 4/10] Every green-path hrm_enable now forces is_wear=false (pre-PR it persisted across enable/disable and was only cleared in hrm_init); the Goodix soft-ADT is re-initialised to DEFAULT on every StartSampling and its WEAR_ON hook fires one interrupt pass after the HR result of the same FIFO batch, so HR results before the ADT re-confirms wear are reported as HRMQuality_OffWrist/bpm 0 on every window, which marks the user off-wrist for sleep tracking and the new SpO2 scheduler and zeroes the live workout HR after every SpO2 hand-off / path switch.**

Worn watch, workout with activity-SpO2 on: after each attempt the manager restarts the green path (hrm_manager.c:432-436) -> gh3x2x.c:559 is_wear=false; Gh3x2xDemoStartSampling -> GH3X2X_AlgoInit -> SoftWearControlInit sets g_SoftAdtWearState=DEFAULT; HBA emits its first BPM at ~9 s and GH3X2X_HrAlgorithmResultReport pushes it synchronously to hrm_manager_new_data_cb in Step 8 of Gh3x2xDemoInterruptProcess, while the WEAR_ON soft event is only consumed in Step 3 of the REPEAT pass and hooked in Step 9 (gh_demo.c:1825-1830, 2163, 2201-2228) -> gh3x2x_hr_result_report reports OffWrist. Downstream: activity.c:201 set_hrm_worn_status(off-wrist) -> kraepelin not_worn (activity_algorithm_kraepelin.c:954-955) and daily SpO2 skipped for up to ACTIVITY_HRM_OFFWRIST_STALE_SEC=900 s (activity.c:353-358) if the window ends before NADT confirms; workout_service.c:220-222 prv_reset_hr_data zeroes live BPM and the median ring each time. The Goodix libraries are binary so exact timing is not constructible (PLAUSIBLE). Better fix: tri-state wear (Unknown on green enable, or restore the pre-SpO2 value) mapped to HRMQuality_Worst / suppressed rather than OffWrist until the ADT decides.

src/fw/apps/demo/spo2_test/spo2_test.c
```diff
@@ -0,0 +1,203 @@
...
+  }
+}
+
+static void prv_enable_spo2(AppData *app_data) {
+  // Allow the SpO2/IR path for the duration of the test; the HRM manager masks it out entirely
+  // when blood oxygen monitoring is off. Use the RAM-only override rather than writing the user's
+  // pref: that one is persisted and synced to the phone, so a force-kill or crash here would strand
+  // daily SpO2 sampling switched on, with Settings showing a state the user never chose.
+  app_data->blood_oxygen_override = true;
+  activity_prefs_set_blood_oxygen_override(true);
```
gmarull: **[code-review 5/10] The RAM-only blood-oxygen override is cleared only in prv_deinit after app_event_loop() returns, so an app crash or force-kill leaves s_blood_oxygen_override stuck true until reboot: the daily SpO2 scheduler samples every 10 min with the user pref off, Settings > Health shows Blood Oxygen 'On', and the toggle becomes a no-op; even during a normal run the override lets the daily scheduler sample/log alongside the app's own session.**

Launch SpO2 Test (override=true at :106); app asserts/hard-faults (fault_handling.c:195-216 suspends the task) or misses the 3 s deinit window (process_manager.c:68-89) -> process_manager_process_cleanup aborts the thread, prv_deinit (:180-183) never runs, hrm_manager_process_cleanup only expires the app's subscription. activity_prefs_blood_oxygen_is_enabled() (prefs.c:2104-2106) stays true -> prv_spo2_subscription_update (activity.c:297) passes every minute -> red/IR sampling every 10 min + minute-record logging. health.c:140 shows 'On'; pressing it calls set_blood_oxygen_enabled(false), a no-op (prefs.c:2114-2115). Fix: clear the override from hrm_manager_process_cleanup / process cleanup for this app, or better exempt foreground-app subscribers from the pref mask so no override is needed at all.

src/fw/services/activity/activity.c
```diff
@@ -241,6 +271,330 @@ static void prv_heart_rate_deinit(void) {
...
+  prv_activity_spo2_update(time_get_uptime_seconds());
+}
+#endif // CONFIG_HRM
+
+// ---------------------------------------------------------------------------------------
+// Init / de-init activity-triggered SpO2 support
+static void prv_activity_spo2_init(void) {
+#ifdef CONFIG_HRM
+  s_activity_state.activity_spo2.phase = ActivitySpO2ActPhase_Idle;
+  s_activity_state.activity_spo2.hrm_session = hrm_manager_subscribe_with_callback(
```
gmarull, line 613: **[code-review 6/10] The two dormant SpO2 sessions (daily + during-activity) are subscribed at boot with a 1-day interval and last_valid_bpm_ticks == 0, which the manager treats as due now; so whenever SpO2 is allowed by prefs (only the 'Blood O2 During Activity' opt-in on, or daily on with interval 'Disabled') the red/IR path is lit at boot and every 24 h with no activity running, and any valid sample is stashed into the minute record.**

Daily Blood Oxygen OFF, Blood O2 During Activity ON: boot -> prv_spo2_init/prv_activity_spo2_init subscribe SpO2 @ SECONDS_PER_DAY; manager: never-served => subscriber_age_ticks = interval_ticks (hrm_manager.c:356-361) => remaining = -spin_up => due (:377-378); allowed_features keeps SpO2 (:331-334); prv_sensor_enable(SpO2) with is_wear forced true (gh3x2x.c:551-553) so OffWrist can never serve it; red LED runs until a valid reading or HRM_MAX_UNSERVED_TIME_SEC=120 s, then repeats every 24 h. A valid sample is stashed by both callbacks (activity.c:416-419, 563-566) and written to the next minute record (activity_metrics.c:557-565) with no activity. The never-served-is-due rule is pre-existing HR behaviour; the PR inherits it for SpO2 without initialising last_valid_bpm_ticks or subscribing lazily when needed.

src/fw/services/hrm/hrm_manager.c
```diff
@@ -821,7 +978,24 @@ void command_hrm_read(void) {
...
+static void prv_console_spo2_read_callback(PebbleHRMEvent *event, void *context) {
+  if (event->event_type == HRMEvent_SpO2) {
+    system_task_add_callback(prv_console_unsubscribe_callback, NULL);
+    char buf[32];
+    prompt_send_response_fmt(buf, 32, "SpO2: %" PRIu8 "%% quality: %" PRIu8, event->spo2.percent,
+                             event->spo2.quality);
+  }
+}
+
+void command_spo2_read(void) {
```
gmarull, line 1023: **[code-review 7/10] New `spo2 read` console command hangs the prompt in the default configuration: with both blood-oxygen prefs off, the new allowed_features mask reduces its HRMFeature_SpO2 subscription to 0, the subscriber is skipped, the sensor never starts, and prompt_command_finish() is never called (s_command_continues_after_return stays set, so later commands also never finish).**

obelix dvt/pvt (not bigboard) with default prefs: `spo2 read` -> command_spo2_read subscribes SpO2 + prompt_command_continues_after_returning(); prv_update_hrm_enable_system_cb masks SpO2 out (hrm_manager.c:324-339), sub_features==0 -> continue (:349-353), wanted_features stays 0, sensor off, prv_console_spo2_read_callback never fires; prompt.c has no timeout (:159-161, :304-317). The demo app works around the same gate with a RAM override; the console command has no guard or error message. (`hrm read` with HR off was already blocked pre-PR by prv_can_turn_sensor_on.)

src/fw/services/activity/workout_service.c
```diff
@@ -284,6 +350,20 @@ void workout_service_init(void) {
...
 
+// ---------------------------------------------------------------------------------------
+// Pause / resume the workout's HR sampling so the optical path is free for a periodic SpO2 reading
+// during a manual workout (mirrors the auto-activity HR pause). Uses set_features rather than
+// set_update_interval so the session's interval/expiration are preserved. KernelBG-safe.
+void workout_service_set_hrm_paused(bool paused) {
+#ifdef CONFIG_HRM
+  prv_lock();
+  {
+    sys_hrm_manager_set_features(s_workout_data.hrm_session, paused ? 0 : HRMFeature_BPM);
```
gmarull, line 361: **[code-review 8/10] Every activity-SpO2 attempt (up to 35 s, every 5 min) silently freezes the live workout HR at its last value: with the workout session's features set to 0 and the sensor restarted in SpO2-only work_mode no HealthEventHeartRateUpdate arrives, and the only staleness guard (WORKOUT_HR_READING_TS_EXPIRE = 60 s) is longer than the attempt, so the stale BPM/zone is rendered as current and hr_zone_time_s attributes the whole gap to the zone of the first post-attempt sample.**

Run workout with 'Blood O2 During Activity' on, app open: at T+5min prv_activity_spo2_start_attempt -> workout_service_set_hrm_paused(true) -> sys_hrm_manager_set_features(session, 0) (:361); sensor restarts with work_mode=GH3X2X_FUNCTION_SPO2 (gh3x2x.c:551-553) so gh3x2x_hr_result_report never fires; prv_handle_heart_rate_update (:212-255) is not called; prv_workout_timer_cb (:293-296) only resets after 60 s > ACTIVITY_SPO2_ACTIVITY_ATTEMPT_SEC 35; active.c re-reads current_bpm every 1 s and shows it as live (measuring animation only when bpm == 0). On resume, hr_zone_time_s (:249-251) adds the entire 35 s gap to the zone of the first new sample. Also note sys_hrm_manager_set_features queues no scheduler pass (hrm_manager.c:925-935), unlike set_update_interval; the pause only takes effect because the caller follows it with a set_update_interval on the SpO2 session. Fix: mark HR stale on pause (e.g. prv_reset_hr_data / a 'measuring' state) and exclude the paused gap from zone-time accounting.

include/pbl/services/hrm/hrm_manager_private.h
```diff
@@ -19,9 +20,10 @@
 
 typedef void (*HRMSubscriberCallback)(PebbleHRMEvent *event, void *context);
 
-// We need roughly this many seconds of "spin up" time to get a good reading from the HR sensor
-// right after turning it on
-#define HRM_SENSOR_SPIN_UP_SEC 20
+// Seconds of "spin up" time needed for a good reading right after turning the sensor on. The Goodix
+// EXCLUSIVE HR model's earliest output lands ~9s in, so 12s leaves a small margin. Only affects
+// sensor pre-warm ahead of a future-due subscriber; a due-now subscriber turns it on immediately.
+#define HRM_SENSOR_SPIN_UP_SEC 12
```
gmarull: **[code-review 9/10] HRM_SENSOR_SPIN_UP_SEC 20->12 is not only a pre-warm tweak as the new comment claims: the same spin-up subtraction drives the turn-OFF decision (subscriber_remaining_ticks = interval - age - spin_up), so app subscribers with a 13-20 s interval (health_service_set_heart_rate_sample_period, unclamped) that pre-PR kept the sensor continuously on are now stopped and restarted (full Goodix algorithm reset, ~9 s to first output) every interval, with HRM_CHECK_SENSOR_DISABLE_COUNT 10->3 making the off decision land within ~3 samples of being served.**

App calls health_service_set_heart_rate_sample_period(15) -> 15 s BPM subscription. Each Good sample refreshes last_valid_bpm_ticks (hrm_manager.c:685-694); every 3rd sample the scheduler runs: remaining = 15 - ~0 - 12 = +3 s -> turn_sensor_on=false -> prv_sensor_disable() + 3 s timer (:452-461); timer fires -> hrm_enable -> Gh3x2xDemoStartSampling -> ~9 s to first output -> served -> off again within <=3 samples. Pre-PR: 15 - 0 - 20 < 0 -> always due -> sensor stayed on (any interval <= 20 s was continuous). Net: off/on with algorithm restart every ~15 s and readings only in a 1-3 s window per cycle instead of a converged 1 Hz stream. The comment at :23-25 ('Only affects sensor pre-warm ahead of a future-due subscriber') is contradicted by :377/:393/:452. Related, lower confidence: the out-of-range hrmPreferences/spo2Preferences fallback changed 10Min->1Hour silently on an unversioned, length-only-validated phone-synced blob (prefs.c:2069-2077, 2133-2140).

src/fw/drivers/hrm/gh3x2x.c
```diff
@@ -525,7 +578,8 @@ bool hrm_enable(HRMDevice *dev, HRMFeature features) {
   }
 #endif
 
-  GH3X2X_FifoWatermarkThrConfig(GH3X2X_FIFO_WATERMARK_CONFIG);
+  GH3X2X_FifoWatermarkThrConfig(low_latency ? GH3X2X_FIFO_WATERMARK_LOW_LATENCY
```
gmarull, line 591: **[code-review 10/10] Forcing every app/worker subscription (Workout app), the BLE HR relay and the console commands to low_latency=true sets the GH3X2X FIFO watermark to 25 instead of the shipped 80, so exactly the longest-running HR sessions now take 3.2x as many FIFO-watermark IRQs and multi-transaction I2C drains; the 'background' 75 is itself ~7% more frequent than the shipped 80, so the '~3x cut' comment is only relative to the new 25.**

Start a Workout (sys_hrm_manager_app_subscribe -> low_latency=true, hrm_manager.c:846-849; workout_service.c:375/418 @1 s) or a BLE HR client (ble_hrm.c:198-200): the subscriber is due at every evaluation so low_latency_wanted is true for the whole session and hrm_enable writes FIFO_WATERLINE=25 (base wrote 80). The driver runs in __NORMAL_INT_PROCESS_MODE__ with EXTI-driven Gh3x2xDemoInterruptProcess and no periodic FIFO timer, and Gh3x2xDemoStartSampling does not overwrite the register (gh_drv_control.c:618 only reads it back), so the IRQ cadence is exactly watermark-driven: per hour of session 80/25 = 3.2x the EXTI wakeups, system-task callbacks and 5-8 I2C transactions each (status read+ack, FIFO count, FIFO burst, AGC reads, wake/sleep). Cheaper: keep 80 as the default and pick 25 (or ~50) only for a foreground live-display consumer with update_interval_s <= 2; the BLE relay's notify cadence and workers do not need it.

src/fw/drivers/hrm/gh3x2x.c
jplexer: was log_enable 0 before, its only restructured here

gmarull (commented):
Final pass on 72a704f. All ten findings from yesterday are addressed in the current branch, so those threads can be resolved:

- ABBA deadlock: the activity-SpO2 state machine now runs from its own KernelBG callback, outside the manager lock.
- Daily SpO2 vs continuous green consumers: the daily reader defers on `hrm_manager_has_continuous_green_subscriber()` and the manager hands the path to SpO2 outright instead of slicing.
- BPM|HRV oscillation: `prv_should_reconfigure()` keeps same-path supersets running (covered by `test_hrm_manager__same_path_superset_keeps_running`).
- `is_wear` forced false: replaced by the assumed-wear / restore pair in the driver.
- RAM override in the demo app: gone; foreground apps bypass the pref mask instead.
- Dormant SpO2 sessions lighting the LED at boot: dormant sessions now carry no features and are skipped.
- `spo2 read` hang: bails with a message when the pref mask excludes SpO2.
- Workout HR freeze during attempts: `workout_service_set_hrm_paused()` resets the live HR on both pause and resume.
- Spin-up 12 s: reverted to 20 s.
- Low-latency FIFO cadence: only foreground apps polling at <= 2 s.

Verified locally: `test_hrm_manager`, `test_activity`, `test_activity_algorithm_kraepelin`, `test_kraepelin_algorithm`, `test_workout_service`, `test_ble_hrm` all pass on this head; gitlint is clean over the range. The nonfree pointer is the current nonfree `main` tip (PR #10), so the dependency is satisfied.

Two non-blocking notes, fine as follow-ups:

1. The PR description still says it requires nonfree PR #7 and is a draft until that merges. PR #7 is closed; the dependency that actually landed is nonfree #10. Please update the text before merging.
2. Activity-SpO2 attempts pause the kalg and workout HR sessions but not the BLE HR relay, and `prv_select_active_path()` gives SpO2 the path whenever it is due. With the during-activity opt-in on and a phone app streaming HR over BLE, the phone sees a gap of roughly 35 s plus the ~9 s green re-convergence every 5 minutes. Opt-in, so acceptable for now; `prv_activity_spo2_can_start_attempt()` could also consult `hrm_manager_has_continuous_green_subscriber()` later.

Good to go from my side once CI's `build-test` job finishes.

gmarull (approved):
Approving; see the final-pass comment above for the two non-blocking follow-ups (stale description, BLE HR relay gaps during activity-SpO2 attempts).

gmarull:
Note: do a final on-device smoke test before merging

jplexer:
smoke test successful!

32bitx64bit:
Awesome!

Glad to see my patches merged. I thank all the maintainers, reviewers, and co-authors who made this possible! :heart:

PR #1953, opened 2026-08-27
fw/apps/launcher: honor the user's Text Size preference in the main menu

uaparit:
Refs #1911.

  Builds on #1912: that PR switched the Settings menu, Bluetooth submenu, and several other
  system screens over to the Text Size preference (`textStyle` pref,
  `system_theme_get_content_size()`). The main menu (the app list, "Launcher") wasn't included
  there; it uses a completely separate rendering path (`apps/system/launcher/default/`), not any
  of the shared `menu_layer_system_cells.c` mechanisms.
  
  This PR (main menu): fonts, cell height, glance cache
  size, and the battery/charging icons in the Settings glance tile now follow the user's
  preference too, instead of fixed per-display-size values.

  **No new setting needed:** the launcher reads the same `textStyle` value that #1912 already
  exposed under "Settings → Display → Text Size" (Smaller/Default/Larger). Users who already know
  that setting now see its effect in the main menu too, without any new menu entry appearing
  anywhere.

  ### What changes

  1. **Fonts** (`menu_layer.c`): title/subtitle font per content-size tier instead of a fixed
     `#if PBL_DISPLAY_HEIGHT >= 200` macro. Medium/Large match today's exact look (no visible
     change without an active user choice); Small mirrors Medium (matching the convention in
     `system_theme.c`: "same as Medium until Small is designed"); ExtraLarge is newly designed
     (`GOTHIC_28_BOLD`/`GOTHIC_24`, one Gothic size step above Large).

  2. **Cell height** (Rect + Round): dynamic in the same way. Round displays always show a fixed
     5 rows at once; the ExtraLarge values were chosen so the total still fits the display at that
     fixed row count.

  3. **Glance cache size** (`app_glance_service.c`/`menu_layer_private.h`): the launcher keeps a
     small LRU cache of the most recently drawn "glance" objects (title/icon/subtitle per app row)
     so it doesn't have to rebuild them on every scroll frame. The cache size (2 × number of
     visible rows) was previously computed at compile time for rect displays from the static
     platform-default cell height. Since cell height can now be *smaller* than that default
     depending on the chosen tier (e.g. "Smaller"), more rows can fit on screen at once than the
     old compile-time estimate assumed, so the cache would have been undersized for that case.
     Fix: the calculation now uses the smallest possible cell height across all tiers (worst case
     = most visible rows), regardless of which tier the user has selected.

     Two things about this that came up during review and were deliberately left as-is:
     - The cache size remains a fixed compile-time constant and does not adapt to the selected
       tier at runtime. It's sized to be large enough for every tier up front. At larger tiers
       (fewer visible rows) the cache ends up somewhat bigger than strictly necessary, but the
       entries are small enough (one list node plus a handful of font/layout fields) that this
       doesn't matter.
     - An undersized cache would not have been a safety issue anyway (verified:
       `prv_glance_cache_put()` always evicts the oldest entry first if needed before adding a new
       one, so the counter can never exceed capacity). The only effect of too small a size would
       have been more cache churn while scrolling (glances rebuilt more often instead of reused),
       not a crash risk.

     Round displays are unaffected: the number of visible rows (5 or 3) is a fixed layout choice
     there, independent of cell height.

  4. **Battery/charging icon in the Settings glance tile**: both now scale too. The charging icon
     used to be a fixed 7x9px bitmap; it's now redrawn as a small vector shape.

  ### Why third-party apps are unaffected

  Unlike #1912, all of the launcher code here is purely internal system code. No watchapp can
  reach these functions or headers.

  ### Validation

  - Rect (`obelix@pvt`, physical device): tested at all three tiers, including the
    battery/charging icon while actually charging.
  - Round (`qemu_gabbro`, QEMU + `pebble-tool`): all four tiers including ExtraLarge, charging
    simulated, clean layout confirmed.
  - `./pbl test`: the full suite passes.

  ### Known limitations / deliberately out of scope for this PR

  - The ExtraLarge values are a first pass, not a final design.
  - The cell-height/font discrepancy between the launcher and Settings at "Large" (pre-existing,
    arose independently) is deliberately not part of this PR, since changing it would affect the
    whole system, not just the launcher.
  - The remaining ~20 screens (Health, Workout, Music, ...) are unchanged, as described in #1911.

### DIFF
--- a/src/fw/apps/system/launcher/default/app_glance_settings.c
+++ b/src/fw/apps/system/launcher/default/app_glance_settings.c
@@ -20,15 +20,41 @@
 #include "pbl/services/notifications/do_not_disturb.h"
 #include "system/passert.h"
 #include "pbl/util/attributes.h"
+#include "pbl/util/math.h"
 #include "pbl/util/size.h"
 #include "pbl/util/string.h"
 #include "pbl/util/struct.h"
 
 #include <stdio.h>
 
-// These dimensions are separate defines so we can use them to statically define the battery points
-#define BATTERY_SILHOUETTE_ICON_WIDTH (16)
-#define BATTERY_SILHOUETTE_ICON_HEIGHT (9)
+typedef struct {
+  int16_t width;
+  int16_t height;
+} IconSize;
+
+static const IconSize s_battery_silhouette_icon_size[NumPreferredContentSizes] = {
+  //! @note this is the same as Medium until Small is designed
+  [PreferredContentSizeSmall] = { .width = 12, .height = 7 },
+  [PreferredContentSizeMedium] = { .width = 16, .height = 9 },
+  [PreferredContentSizeLarge] = { .width = 16, .height = 9 },
+  [PreferredContentSizeExtraLarge] = { .width = 21, .height = 12 },
+};
+
+static IconSize prv_get_battery_silhouette_icon_size(void) {
+  return s_battery_silhouette_icon_size[launcher_menu_layer_get_clamped_content_size()];
+}
+
+static const IconSize s_charging_icon_size[NumPreferredContentSizes] = {
+  //! @note this is the same as Medium until Small is designed
+  [PreferredContentSizeSmall] = { .width = 5, .height = 7 },
+  [PreferredContentSizeMedium] = { .width = 7, .height = 9 },
+  [PreferredContentSizeLarge] = { .width = 7, .height = 9 },
+  [PreferredContentSizeExtraLarge] = { .width = 9, .height = 12 },
+};
+
+static IconSize prv_get_charging_icon_size(void) {
+  return s_charging_icon_size[launcher_menu_layer_get_clamped_content_size()];
+}
 
 typedef struct LauncherAppGlanceSettingsState {
   BatteryChargeState battery_charge_state;
@@ -45,7 +71,6 @@ typedef struct LauncherAppGlanceSettings {
   char battery_percent_text[5]; //!< longest string is "100%" (4 characters + 1 for NULL terminator)
   KinoReel *icon;
   uint32_t icon_resource_id;
-  KinoReel *charging_indicator_icon;
   uint8_t subtitle_font_height;
   LauncherAppGlanceSettingsState glance_state;
   EventServiceInfo battery_state_event_info;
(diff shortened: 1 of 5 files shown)


src/fw/apps/system/launcher/default/menu_layer.c
```diff
@@ -14,12 +14,90 @@
...
+    .rect_cell_height = 50,
+    .round_focused_cell_height = 55,
+    .round_unfocused_cell_height = 45,
+  },
+  [PreferredContentSizeExtraLarge] = {
+    .rect_cell_height = 60,
+    // Round always shows 5 fixed rows, so focused + 4*unfocused must fit the 260px display.
+    // Keeps Large's 55:45 ratio and ~25px margin rather than extrapolating Large's growth rate,
+    // which would overflow.
+    .round_focused_cell_height = 57,
```
gmarull: there's a ton of unnecessary comments that should be removed and/or extended details be kept in git history.

gmarull:
Also, please, post image/s comparing before/after

src/fw/apps/system/launcher/default/app_glance_settings.c
```diff
@@ -20,15 +20,41 @@
...
+  [PreferredContentSizeLarge] = { .width = 16, .height = 9 },
+  [PreferredContentSizeExtraLarge] = { .width = 21, .height = 12 },
+};
+
+static IconSize prv_get_battery_silhouette_icon_size(void) {
+  return s_battery_silhouette_icon_size[launcher_menu_layer_get_clamped_content_size()];
+}
+
+static const IconSize s_charging_icon_size[NumPreferredContentSizes] = {
+  //! @note this is the same as Medium until Small is designed
```
gmarull, line 48: "this is the same", which one? note is confusing, could also be adjusted now?

src/fw/apps/system/launcher/default/menu_layer.c
```diff
@@ -14,12 +14,80 @@
...
+    .round_unfocused_cell_height = 38,
+  },
+  [PreferredContentSizeLarge] = {
+    .rect_cell_height = 50,
+    .round_focused_cell_height = 55,
+    .round_unfocused_cell_height = 45,
+  },
+  [PreferredContentSizeExtraLarge] = {
+    .rect_cell_height = 60,
+    // Round always shows 5 fixed rows, so focused + 4*unfocused must fit the 260px display.
```
gmarull, line 85: what if a larger/different display comes? hardcoding even in comments values like 260px is a bad idea

src/fw/apps/system/launcher/default/menu_layer_private.h
```diff
@@ -6,20 +6,9 @@
...
-
-#if PBL_ROUND && PBL_DISPLAY_HEIGHT >= 200
-#define LAUNCHER_MENU_LAYER_CELL_ROUND_FOCUSED_CELL_HEIGHT (55)
-#define LAUNCHER_MENU_LAYER_CELL_ROUND_UNFOCUSED_CELL_HEIGHT (45)
-#else
-#define LAUNCHER_MENU_LAYER_CELL_ROUND_FOCUSED_CELL_HEIGHT (52)
-#define LAUNCHER_MENU_LAYER_CELL_ROUND_UNFOCUSED_CELL_HEIGHT (38)
-#endif
+// Smallest tier's row height, used only to size the glance cache (app_glance_service.c) for the
+// worst case. Not the actual rendered height - keep in sync with s_launcher_cell_dimensions.
```
gmarull, line 10: excessive verbosity again, come on Claude, you can do better

uaparit:
Ok ok i understood

PR #1970, opened 2026-08-30
Notifications: Group same app/tittle notifications into a single list row

Elyriel:
### Summary

Groups notifications from the same app/title into a single row instead of listing each one individually, so a batch send (e.g., 4 notifications from one app/person) doesn’t flood the list.

### Changes :
`src/fw/apps/system/notifications.c`:

  * Notifications with the same app name and or title are grouped when there are
    three or more matching notifications.
  * A group row displays the group name and notification count.
  * Selecting a group opens a separate animated notifications window containing
    only that group's notifications.

  `src/fw/apps/system/settings/notifications.c`:

  * Added the `Sort Order` setting:
    - Newest first
    - Oldest first
    - Alphabetical
  * Added the `Group Notifications` On/Off setting.

  `src/fw/services/notifications/alerts_preferences.c`:

  * Added persistent storage for notification sorting and grouping preferences.
  * Added validation and default values for the new preferences.

  `resources/normal/base/lang/`:

  * Added translations for the new notification settings in all supported
    language catalogs.

### example :
 8 notifications were sent, same applies to all screenshots below.
#### before :
image
image

#### after :
image
image
image

### Validation

Tested in QEMU emulator, on Round 2, Time 2, and Pebble 2 Duo, all are working. I generated the notifications using `./pbl console notif test` first, and then a vibe-coded tool

Use of AI:

I've used generative AI (local models, Claude Sonnet 5, ChatGPT 5.6 Luna) to debug, choose function names, understand the project, write comment and help occasionally.
(edit : ) I didn’t use AI to make the logic or write the entire code.

PR #1973, opened 2026-08-31
notifications: group Android notification history by sender

TheFallen058:
## Summary
- Add optional sender-based grouping to notification history.
- Support Never, 1 Day, 1 Week, and All grouping ranges.
- Display grouped messages newest-first with existing notification actions.
- Keep old, unsupported, and iOS notifications ungrouped.
- Apply grouping only to history, not incoming notification popups.
- Skip corrupt history records without hiding subsequent notifications.
- Group Android titles formatted `Conversation: Sender` by conversation when sender metadata is unavailable.
## Testing
- Full unit suite: 332/332 passed
- QEMU Gabbro firmware build passed
- Pebble Time 2 (`obelix@pvt`) dual-slot release build passed
- Sideloaded and manually verified on a retail Pebble Time 2 + Galaxy s25
- clang-format and gitlint passed
## Limitations
- Grouping prefers Android sender metadata when available.
- Otherwise, titles formatted `Conversation: Sender` use `Conversation` as the grouping key.
- This title-prefix fallback can over-group unrelated Android notifications sharing the same prefix.
- Grouping keys are not separated by application.
- Dismissing an entire group is intentionally unsupported.
## Compatibility
- Android notifications are grouped using sender metadata or a `Conversation: Sender` title.
- iOS/ANCS notifications remain ungrouped and retain existing history and dismissal behavior.
## AI assistance
GPT-5.6 Sol assisted with implementation. Relevant commits contain
`Co-authored-by` trailers. The resulting behavior was manually tested on hardware.

Example:
Before:
image
After:
image
image
image


TheFallen058:
## Related work
Related to #1970. This implementation additionally provides configurable age ranges, sender-based grouping, transcript view, iOS compatibility, storage optimizations, and unit coverage.

gmarull:
@ericmigi

gmarull:
### Title-prefix fallback over-groups unrelated notifications

`prv_conversation_range_from_title()` (`src/fw/apps/system/notifications_history.c`) keys a notification on whatever precedes the first `": "` in its title when no `AttributeIdSender` is present. That is the right shape for `Group: Sender` chat titles, but it also collapses titles that merely share a prefix: `Re: Invoice` + `Re: Lunch` become a `Re (2)` group, `Error: disk full` + `Error: login` become `Error (2)`, and the group header/preview then hides the actual senders.

The firmware-side options are not great:

- Restricting the fallback to `LayoutIdCommNotification`, or requiring `AttributeIdSender`, would disable grouping entirely on Android today: the Core companion always sends `Layout.GenericNotification` with the raw `EXTRA_TITLE` and never sets `Sender` (`libpebble3/.../notification/LibPebbleNotification.kt`, `processor/BasicNotificationProcessor.kt`).
- Any narrower title rule is just another heuristic.

So I'd leave the heuristic in as-is for now, and fix the data at the source.

### Proposed CoreApp fix

Populate `AttributeIdSender` on Android notifications from the notification extras, which already carry the real conversation identity:

1. `NotificationCompat.MessagingStyle.extractMessagingStyleFromNotification(notification)` — the companion already calls this in `NotificationImage.kt` for image batches. Use `conversationTitle` when `isGroupConversation` is set, otherwise the newest `Message.person?.name`.
2. Fall back to `notification.people()` (first `Person.name`) for apps that set `android.people.list` but no messaging style.
3. Emit it via the existing DSL: `attributes { sender { it } }` next to `title { }` in `LibPebbleNotification.toTimelineNotification()`.

On the watch `AttributeIdSender` is not rendered by the notification layout (only `calendar_layout.c` and `timeline_actions.c` read it), so this is UI-neutral, and the `sender` path in `prv_group_sender_for_item()` already takes precedence over the title fallback. Once the companion ships it, the title heuristic can be removed in a follow-up.

gmarull:
@sjp4 please check mobile suggested changes

TheFallen058:
To @gmarull 
One adjustment to the 'sender' rule, though,
otherwise it regresses group chats.

The title fallback keys on the text *before* the first `": "`, so `PG | Elite: Aloha`
and `PG | Elite: Yeuhen` both group under the conversation `PG | Elite` (tests to that exist). When
`AttributeIdSender` is present the firmware takes it verbatim and never splits it
(`test_notifications_history__sender_attribute_is_not_split`), so whatever the
companion puts there becomes the grouping key outright.

That matters because `isGroupConversation()` is false on the
`extractMessagingStyleFromNotification()` path unless the app explicitly set the
`android.isGroupConversation` extra — the `conversationTitle != null` compat fallback
is gated on `mBuilder != null`, which extraction never sets. So for apps that set a
conversation title but never call `setGroupConversation(true)`, the rule falls through
to `Message.person?.name` and a group chat fragments into one row per participant.
Worse than the `Re (2)` over-grouping it's meant to fix.

Suggested flow for the sender value:

1. `conversationTitle` whenever it is non-null (not gated on `isGroupConversation`).
2. Otherwise the newest incoming `Message.person?.name`.
3. No `notification.people()` fallback — in a group chat `android.people.list` holds
   every participant, so the first name is arbitrary and unstable across notifications.

ericmigi:
I have the same question as in https://github.com/coredevices/PebbleOS/pull/1970#issuecomment-5706050800

Please share some more screenshots of how it would look when entering from a regular notification rather than opening the notification app itself.

TheFallen058:
@ericmigi  Sure
`qemu_emery`, with Settings → Notifications → **Group by Sender** set to **1 Day**. Short version: the incoming-notification path is untouched by this PR. Grouping only exists inside the notifications app; a notification that arrives looks and behaves exactly as it does on `main`.
**A message arrives** from a conversation that is grouped in history. The modal shows the raw title, no group, no count:
03-incoming-notification

**Two unread.** The status bar counts individual notifications and up/down walks them one at a time, even though both belong to groups:
04-incoming-second-1of2
05-incoming-second-2of2

**Its action menu** keeps Dismiss All:
06-incoming-action-menu-dismiss-all

The one deliberate difference: opening a message **from inside a group** omits Dismiss All, since dismissing everything from within one conversation would take unrelated notifications with it:
12-message-from-group
13-group-action-menu-no-dismiss-all

**A notification arriving while the list is open** is the same modal over the app. Once dismissed, it folds into its group, which moves to the top with the count incremented (5 → 6):
07-arrival-over-open-list
08-list-after-arrival

For context, the grouped list (8 notifications in 4 rows) and a group opened as a transcript:
09-grouped-list
10-grouped-list-scrolled
11-group-transcript

Grouping is off by default (`NotificationGroupingRange_Never`), so nothing changes for a user who doesn't turn it on:
01-setting-group-by-sender
02-setting-ranges

TheFallen058:
@ericmigi 
Here is a direct comparison using the same notification data on current `main` (`749c8cfd1`) and this PR (`ad2c5a7fc`).

### Regular incoming notifications are unchanged

The PR screenshots below were taken with grouping enabled. A single incoming notification has the same modal presentation on both versions:
01-single-incoming-main-vs-pr

Multiple incoming notifications remain individual and retain the `1/2` counter:
02-multiple-incoming-main-vs-pr

The action menu is also unchanged, including **Dismiss All**:
03-incoming-actions-main-vs-pr

### The difference exists only in Notification History

With grouping disabled, the PR behaves like `main`:
04-grouping-disabled-main-vs-pr

With grouping enabled, messages from the same conversation collapse into one counted row. The comparison below uses the same eight notifications:
05-history-main-vs-grouped-pr

Selecting a group opens its conversation transcript. Selecting a message then opens the existing notification view:
06-group-drill-down

If a notification arrives while history is open, the normal modal still appears. After returning, the conversation moves to the top and its count increases from 4 to 5:
07-arrival-while-history-open

Grouping is opt-in and defaults to **Never**:
08-opt-in-settings

ericmigi:
ah sorry I didn't explain what I was worried about - I guess I just need to try it out. lets merge it conditionally for testing (default off) then we can try it in normal situations

TheFallen058:
updated from last master to resolve conflicts
Plus 2 small issues fixed:
- sender name was hidden in first open
- notification incoming time font was downsized.

@gmarull for this feature is no mobile companion changes, the logic exists only on PebbleOS side.

@ericmigi to be honest it's mostly impossible in current implementation to handle conversations if you have 100+ notifications per day in multiple groups

ericmigi (approved):

gmarull:
@TheFallen058 please rebase, make sure CI is green.

Note we merged this into mobile app:

```patch
From 135ee98928c5021a38aba412fd13af86bfc7dba8 Mon Sep 17 00:00:00 2001
From: Gerard Marull-Paretas <gerard@teslabs.com>
Date: Thu, 17 Sep 2026 19:34:16 +0200
Subject: [PATCH] Android: send the sender attribute on notifications (#275)

The watch groups notification history by AttributeIdSender, but Android
notifications only ever carried the raw EXTRA_TITLE, so the firmware
falls back to a title-prefix heuristic that also merges unrelated
"Re: ..." / "Error: ..." titles into one group.

Derive the sender from the MessagingStyle extras: the conversation title
whenever one is set, otherwise the Person of the newest incoming message.
The title is not gated on isGroupConversation because on the extraction
path the flag is only ever the raw android.isGroupConversation extra,
which apps that set a title but never call setGroupConversation(true)
leave false; gating on it would fragment those group chats into one
row per participant. A group chat without a title sends no sender so
the watch keeps its title fallback, and android.people.list is not used
because its order is app-defined and unstable across notifications.

The watch does not render Sender on notifications, so this only affects
history grouping.

Related: coredevices/pebbleos#1973

Signed-off-by: Gerard Marull-Paretas <gerard@teslabs.com>
Co-authored-by: Claude Fable 5.1 <noreply@anthropic.com>
---
 .../notification/NotificationSenderTest.kt    | 53 +++++++++++++++++++
 .../notification/LibPebbleNotification.kt     |  4 ++
 .../notification/NotificationSender.kt        | 35 ++++++++++++
 .../processor/BasicNotificationProcessor.kt   |  2 +
 4 files changed, 94 insertions(+)
 create mode 100644 public/libpebble3/src/androidHostTest/kotlin/io/rebble/libpebblecommon/notification/NotificationSenderTest.kt
 create mode 100644 public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationSender.kt

diff --git a/public/libpebble3/src/androidHostTest/kotlin/io/rebble/libpebblecommon/notification/NotificationSenderTest.kt b/public/libpebble3/src/androidHostTest/kotlin/io/rebble/libpebblecommon/notification/NotificationSenderTest.kt
new file mode 100644
index 000000000..09d1cc08b
--- /dev/null
+++ b/public/libpebble3/src/androidHostTest/kotlin/io/rebble/libpebblecommon/notification/NotificationSenderTest.kt
@@ -0,0 +1,53 @@
+package io.rebble.libpebblecommon.notification
+
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertNull
+
+class NotificationSenderTest {
+    @Test
+    fun conversationTitleWinsRegardlessOfGroupFlag() {
+        assertEquals(
+            "PG | Elite",
+            selectSender(
+                conversationTitle = "PG | Elite",
+                isGroupConversation = true,
+                newestIncomingSender = "Aloha",
+            ),
+        )
+        assertEquals(
+            "PG | Elite",
+            selectSender(
+                conversationTitle = "PG | Elite",
+                isGroupConversation = false,
+                newestIncomingSender = "Aloha",
+            ),
+        )
+    }
+
+    @Test
+    fun blankConversationTitleIsIgnored() {
+        assertEquals(
+            "Anna",
+            selectSender(conversationTitle = " ", isGroupConversation = false, newestIncomingSender = "Anna"),
+        )
+    }
+
+    @Test
+    fun directConversationUsesNewestIncomingSender() {
+        assertEquals(
+            "Anna",
+            selectSender(conversationTitle = null, isGroupConversation = false, newestIncomingSender = "Anna"),
+        )
+    }
+
+    @Test
+    fun groupConversationWithoutTitleIsNull() {
+        assertNull(selectSender(conversationTitle = null, isGroupConversation = true, newestIncomingSender = "Aloha"))
+    }
+
+    @Test
+    fun nothingKnownIsNull() {
+        assertNull(selectSender(conversationTitle = null, isGroupConversation = false, newestIncomingSender = null))
+    }
+}
diff --git a/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt
index 22a66e17c..1b7f3f0d2 100644
--- a/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt
+++ b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/LibPebbleNotification.kt
@@ -30,6 +30,7 @@ data class LibPebbleNotification(
     val timestamp: Instant,
     val title: String?,
     val body: String?,
+    val sender: String? = null,
     val icon: TimelineIcon,
     val actions: List<LibPebbleNotificationAction>,
     val people: List<ContactEntity>,
@@ -122,6 +123,9 @@ data class LibPebbleNotification(
             body?.let {
                 body { it }
             }
+            sender?.let {
+                sender { it }
+            }
             color?.let {
                 backgroundColor { it.toPebbleColor() }
             }
diff --git a/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationSender.kt b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationSender.kt
new file mode 100644
index 000000000..545567610
--- /dev/null
+++ b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/NotificationSender.kt
@@ -0,0 +1,35 @@
+package io.rebble.libpebblecommon.notification
+
+import android.service.notification.StatusBarNotification
+import androidx.core.app.NotificationCompat
+import io.rebble.libpebblecommon.util.stripBidiIsolates
+
+/**
+ * Who the notification is from, so the watch can group its history by conversation: the
+ * conversation title when set, otherwise the sender of the newest incoming message.
+ */
+fun StatusBarNotification.extractSender(): String? {
+    val style = NotificationCompat.MessagingStyle.extractMessagingStyleFromNotification(notification)
+        ?: return null
+    val user = style.user
+    val newestIncoming = style.messages.asReversed().firstOrNull { message ->
+        val person = message.person
+        person != null && person != user
+    }
+    val sender = selectSender(
+        conversationTitle = style.conversationTitle,
+        isGroupConversation = style.isGroupConversation,
+        newestIncomingSender = newestIncoming?.person?.name,
+    )
+    return stripBidiIsolates(sender)?.trim()?.takeIf { it.isNotEmpty() }
+}
+
+internal fun selectSender(
+    conversationTitle: CharSequence?,
+    isGroupConversation: Boolean,
+    newestIncomingSender: CharSequence?,
+): CharSequence? = when {
+    !conversationTitle.isNullOrBlank() -> conversationTitle
+    isGroupConversation -> null
+    else -> newestIncomingSender
+}
diff --git a/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
index 324697a52..fc96ddff4 100644
--- a/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
+++ b/public/libpebble3/src/androidMain/kotlin/io/rebble/libpebblecommon/notification/processor/BasicNotificationProcessor.kt
@@ -23,6 +23,7 @@ import io.rebble.libpebblecommon.io.rebble.libpebblecommon.notification.Notifica
 import io.rebble.libpebblecommon.io.rebble.libpebblecommon.notification.people
 import io.rebble.libpebblecommon.io.rebble.libpebblecommon.notification.vibrationPattern
 import io.rebble.libpebblecommon.notification.extractAttachment
+import io.rebble.libpebblecommon.notification.extractSender
 import io.rebble.libpebblecommon.packets.blobdb.TimelineIcon
 import io.rebble.libpebblecommon.timeline.TimelineColor
 import io.rebble.libpebblecommon.timeline.argbColor
@@ -90,6 +91,7 @@ class BasicNotificationProcessor(
             key = sbn.key,
             title = title,
             body = body,
+            sender = sbn.extractSender(),
             icon = icon,
             // Clamp to postTime so messaging apps that set `when` to thread-start time
             // don't get filtered out by BlobDB's `onlyInsertAfter` sync threshold (MOB-7284).
```

PR #1979, opened 2026-08-31
drivers/touch/cst816: add palm gesture to turn off backlight

josiahcbloomer:
## Summary

Tap the screen with your palm and the backlight turns off, mirroring a gesture found on many major smartwatches on the market. Adds a new option: Settings → Display → Backlight → **"Palm to sleep"**, which is on by default.

## How the palm is detected
The CST816 seems to report this itself. Register `0x01` reads **`0xAA`** for a palm, and doesn't seem to be for anything else. it isn't in the documented gesture list and nothing in the tree referenced it. Found by dumping the chip's register block on device and testing repeatedly on my PT2.
I would love if some more people would build and test on their watches.

## Tests
Built for obelix@pvt. tests pass.
Tested on my Core Time 2. Haven't noticed any false positives yet.

Co-authored with Claude.

### DIFF
--- a/include/pbl/services/light.h
+++ b/include/pbl/services/light.h
@@ -44,6 +44,11 @@ void light_touch_up(void);
 void light_enable(bool enable);
 
 //! @internal
+//! Drop the backlight now, as if its timeout had expired. Unlike
+//! light_enable(false), does not claim app control of the light.
+void light_off_now(void);
+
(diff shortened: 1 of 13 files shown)


gmarull:
Please, open first a new PR to add support for the gesture only.

PR #1985, opened 2026-09-02
kernel: introduce the pbl kernel API and replace FreeRTOS with a native kernel

gmarull:
Introduce a kernel API owned by PebbleOS under `include/pbl/kernel/`, implemented by a new top-level `kernel/` folder, move the whole firmware onto it, and replace FreeRTOS with PebbleOS's own kernel. The series lands the API on a FreeRTOS shim first so every migration commit is bisectable against the old kernel, adds the native kernel next to it, makes it the default, and finally removes FreeRTOS and the backend abstraction. At the tip there is one kernel, in `kernel/`, and no FreeRTOS anywhere in the tree.

## What's in here

- **API and shim** (`kernel/`, `include/pbl/kernel/`): `pbl_mutex`, `pbl_sem`, `pbl_msgq`, `pbl_poll_group`, `pbl_thread`, the IRQ lock, scheduler state, tick helpers, the SoC idle interface and the introspection core dumps and the watchdog need. Objects are caller-owned structs; `PBL_*_DEFINE` initialises them statically, and backends that need a runtime create step (the shim) register them in a linker section that `pbl_kernel_init()` walks before the scheduler starts. The first commit carries the complete API and shim; the migration commits only touch call sites. Design notes are in `docs/architecture/kernel.md`.
- **Call-site migration**, one primitive per commit so each step is bisectable: mutexes (removing the libos mutex), semaphores, queues and queue sets, tasks, then critical sections, scheduler queries and ticks. Every kernel call that can wake a thread requests the context switch itself, so the ISR-side `portEND_SWITCHING_ISR` bookkeeping is gone.
- **FreeRTOS made private, then removed**: `FreeRTOSConfig.h` and the port hooks live in `kernel/freertos`, driven by new Kconfig symbols (`KERNEL_TICK_HZ`, `KERNEL_NUM_PRIORITIES`, `KERNEL_IRQ_PRIO_*`). The per-SoC tickless idle code is written against `pbl/kernel/idle.h`. `third_party/freertos` is only built for the shim and exports its include directories only to it, so a stray include or symbol reference fails at compile or link time. The SiFli IPC and nrfx glue headers, which declared `vPortEnterCritical` themselves, now use `pbl_irq_lock`. The last two commits delete the shim, the `FreeRTOS-Kernel` submodule, the backend choice and the `.pbl_kobj_init` runtime-init walk (every `PBL_*_DEFINE` is a complete static initialiser), then move `kernel/native` up into `kernel/`.
- **Native kernel** (`kernel/`, flattened from `kernel/native/` once the shim is gone): fixed-priority preemptive scheduler with one ready list per priority and a CLZ bitmap, round-robin on the tick, priority-ordered wait queues, one sorted timeout list, direct handoff on semaphores and mutexes, single-level priority inheritance and a tickless idle thread. The Cortex-M arch keeps the saved-context layout, SVC numbers, syscall-stack relocation and per-thread MPU region switch of the FreeRTOS port, so the syscall assembly, core dumps and fault handling are unchanged. A POSIX arch runs kernel threads as pthreads under the kernel's own decisions with a virtual clock; `tests/kernel/test_native_kernel.c` uses it to cover preemption, time slicing, ISR semaphores, inheritance, queue blocking, poll groups, suspend/resume/abort, thread exit, priority changes and the scheduler lock. `docs/architecture/kernel_internals.md` has the design, the comparison with FreeRTOS, Zephyr, NuttX and ThreadX, what was borrowed from each, and what was left out.
- **Moddable**: the submodule's Pebble platform glue (coredevices/moddable#2) uses `pbl_mutex`, `pbl_thread_sleep`, `pbl_irq_lock` and `pbl_thread_current` instead of FreeRTOS, so the private include exceptions and the `MODDABLE_XS` exclusion for the native backend are gone.

## Benchmarks

Measured with a throwaway harness that ran in place of KernelMain on QEMU (not kept in the tree). ns per iteration, QEMU virtual time, `-Os`, all boot services running. Only the relative numbers between backends on the same board mean anything.

| Benchmark | gabbro shim | gabbro native | flint shim | flint native |
| --- | ---: | ---: | ---: | ---: |
| irq_lock_unlock | 151 | 142 | 82 | 102 |
| mutex_lock_unlock (uncontended) | 927 | 413 | 610 | 283 |
| sem_give_take (no wait) | 658 | 347 | 398 | 247 |
| msgq_put_get (4 bytes, no wait) | 723 | 602 | 458 | 341 |
| yield_switch | 10869 | 11323 | 10756 | 10588 |
| sem_pingpong (round trip) | 28647 | 23715 | 26421 | 21401 |
| msgq_pingpong (round trip) | 30805 | 25265 | 26433 | 22376 |
| poll_wake (three-queue group) | 29109 | 24520 | 25681 | 23205 |
| irq_to_thread | 2000 | 1620 | 1384 | 1244 |
| thread_create_exit | 33149 | 24441 | 29712 | 21871 |
| sleep_1_tick | 997605 | 1001495 | 998345 | 996250 |

Uncontended primitives are 1.2x to 2.2x faster natively; anything that context switches is dominated by the exception entry and MPU reprogramming and lands within 20%. Kernel code size: 12.5 KB (shim + FreeRTOS) vs 4.7 KB (native) on qemu_gabbro. The PendSV path was reviewed instruction by instruction against the FreeRTOS port.

## Behaviour changes worth knowing

- Mutexes are a single recursive type; non-recursive intent is expressed with `pbl_mutex_assert_held(m, false)`.
- Returning from a thread entry function ends the thread on both backends.
- Stack high-water values reported through `debug.h` are in bytes (the old code reported words under a `_bytes` analytics key).
- The NewTimer, PULSE, NimBLE and boot splash threads use static stacks instead of FreeRTOS heap allocations, so that RAM moves from the kernel heap into `.bss`.
- Static objects no longer need a create call in an init function; where a "is the handle created yet" check doubled as "is this service initialised", a bool replaces it. `pbl_kernel_init()` is gone.
- The asterix OpenOCD config loses its FreeRTOS thread awareness. `tools/qemu/qemu_gdb_proxy.py` now walks the kernel's thread list through a layout descriptor the kernel exports (`pbl_kernel_debug_layout`), so `info threads` and per-thread backtraces work on QEMU again.

## Known gaps

- The moddable submodule pointer moves to the merged coredevices/moddable#2, which ports its Pebble platform glue to the pbl API.
- Driver ISR callbacks still carry a dead `bool *should_context_switch` parameter and `event_put_isr()` always returns `false`; both are follow-up cleanups.
- The native backend has only run in QEMU. Its SF32LB tickless path and the SDK-side effects need on-device testing; the shim has booted on obelix.

## Verification

Per commit: qemu_gabbro builds and boots to the watchface with a gdb breakpoint sweep on every assert and fault entry point, asterix and obelix build, and the full unit test suite passes (333 tests, 18 of them for the native kernel). Obelix (XS on) and asterix build and link on both backends; the native images contain no FreeRTOS symbols. The native backend boots the firmware and launches and exits apps on qemu_gabbro (ARMv8-M, with XS enabled) and qemu_flint (ARMv7-M) without asserts. On hardware, obelix boots and works as expected (2026-09-04, confirmed again on the restructured series).

https://claude.ai/code/session_0115ghSX4ArXuBNC2M9bbEpL


ck-telecom:
export \_\_NATIVE_KERNEL\_\_?

PR #1992, opened 2026-09-04
fw/apps/alarms: Keep backlight on during alarm popup

FrederickGeek8:
This PR changes the behaviour of the system alarm popup to force the backlight on while the alarm popup has yet to be dismissed rather than use the user-configured backlight timeout. The motivation behind this change stems from one of my human limitations that haunts me every morning:

_My eyes take a few seconds to "turn on" when I wake up, I guess, so it's less than ideal when I can no longer see which button is Snooze and which is Dismiss on my alarm once I realize what is going on._

This PR is a bit opinionated in its nature, so perhaps it cannot be accepted without user configurability. That said, adding configuration options to the mobile app is a bit out of my area of familiarity with the codebase (right now), so I'm hoping this is not too crazy of a change to require that (which may take me a few days to implement).

### DIFF
--- a/src/fw/popups/alarm_popup.c
+++ b/src/fw/popups/alarm_popup.c
@@ -315,6 +315,7 @@ static void prv_cleanup_alarm_popup(void *callback_context) {
 #ifdef CONFIG_SPEAKER
     prv_stop_sound();
 #endif
+    light_enable(false);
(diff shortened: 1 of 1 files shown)


gmarull (approved):
let's try, I like the idea. it can always be reverted and/or made configurable

PR #2000, opened 2026-09-06
resources: remove icons for defunct services/apps

squabbled:
Remove icons and declarations for defunct services and apps. Removed icons/declarations are Skype, Hipchat, BBMessenger, Google hangouts, Google inbox, Foursquare.


TeamSpen210:
Should we remove these? They're still useful to be assigned to notifications, if an app you use doesn't have a specific one. Though properly generic ones would be better.

squabbled:
> Should we remove these? They're still useful to be assigned to notifications, if an app you use doesn't have a specific one. Though properly generic ones would be better.

Generic icons would fit the bill better. I'm not sure the use in assigning the Hipchat or BBN icons to modern apps really compared to a generic icon or new app icon. The only one I can think of that might be useful is using the Skype icon on GroupMe which is the successor but even then it could use a new icon entirely

jplexer:
I know this is well meant but we should never ever remove resources unless it is really really necessary, apps may be using them

PR #2029, opened 2026-09-07
apps/weather: localize the Weather app

gmarull:
## Summary

The Weather app drew nearly all of its text from English string literals, so it stayed English in every locale. Only the empty-state dialog and the app title went through i18n.

This routes every user-visible string through the i18n layer and translates them in all shipped catalogs.

**Strings localized** (across `weather.c`, `weather_app_layout.c`, `expanded_view.c`, `forecast_list.c`, `globe_view.c`, `saved_locations.c`):

- Condition phrases (Sunny, Partly Cloudy, Rain, Snow, …) and day labels (Today, Tomorrow, weekday names).
- The forecast warning ladder (Heavy rain, Flood risk, Heatwave, Very high UV, Clear skies, …) and the wind/precipitation sentences, including the "at N mph/km/h" wording and the 8-point compass directions.
- Screen labels: UV INDEX, PRECIPITATION, WEATHER REPORT, CURRENT UV, RAIN, the sunset and "Last updated" lines, and the AM/PM suffix.
- Globe and list captions: CITY SELECT, SAVED LOCATIONS, Current Location, Saved Location, Location, No Locations, "Add them in the app".

Each view passes itself as the i18n owner so the strings are freed when it unloads, and the buffers that receive translated text were widened to hold the longer strings (verified against the actual translations, including the CJK and Cyrillic catalogs). The derived current-hour phrase is now translated into an app-owned buffer rather than pointing at a literal, and falls back to the record's own phrase when the weather type has no name.

**Catalogs**: `tintin.pot` regenerated, every catalog msgmerged and filled (real translations for the localized languages, identity English for the en_TW/ru_RU extended-font packs), wrong fuzzy matches cleared, and `Project-Id-Version` bumped on each so phones offer the updated packs. Every catalog is 754/754 translated (only the 3 intentional empty-msgid template entries remain).

## Test plan

- [x] `pbl build` (qemu_emery) clean, including the format-truncation checks on the widened buffers.
- [x] `msgfmt -c` passes on all 13 catalogs; existing translations preserved; identity catalogs stay identity.
- [x] Language packs build (`pbl pack_lang`).
- [x] English rendering unchanged on QEMU: forecast strip, precipitation row, day columns, current-hour phrase and warnings all render correctly.


gmarull:
@sjp4 — heads up on the mobile side, flagging from this PR.

Localizing the Weather app on-watch surfaced that day 0's condition phrase comes straight from the phone's weather record (`short_phrase`), so it renders in whatever language the phone sent it in, regardless of the watch's language. Here I've switched the watch to derive and translate that phrase from the weather *type* instead, falling back to the phone's text only when the type has no name.

IMHO the right approach is the watch controlling all text: the phone sends structured data (weather types, temps, WMO codes, metrics) and the watch renders every string from its own catalogs. That keeps the weather UI in the watch's language no matter what the phone is set to, and means the mobile app doesn't need to send localized phrase strings at all. Raising it so the mobile side can plan around dropping those.

PR #2039, opened 2026-09-08
services/speaker: keep PCM playback fed under app load

ShinkoNet:
CPU-heavy apps can leave speaker PCM refills late. The driver already dispatches its callback on KernelBG, but the service queues another callback, refills only one block after a delay, and leaves KernelBG below app priority. The SF32LB circular buffer also publishes its length before copying bytes while the DMA interrupt consumes the same buffer.

- Serialize SF32LB circular-buffer writes against the DMA interrupt.
- Refill directly from the driver callback and use the reported free capacity to catch up.
- Hold raised KernelBG priority during playback, with balanced reference counting for priority requests.

Tested on Pebble Time 2, and also confirmed QEMU still output audio.

I've been working on a Game Boy Color emulator, and it creates a lot of CPU usage. These audio fixes were tested in my own fork along with my emulator (Pebbleboy) to confirm the audio no longer stuttered during emulation.

### DIFF
--- a/src/fw/services/speaker/speaker_service.c
+++ b/src/fw/services/speaker/speaker_service.c
@@ -205,6 +205,8 @@ static void prv_start_audio(uint8_t vol) {
   PBL_ANALYTICS_ADD(speaker_play_count, 1);
   prv_update_volume_analytics(effective_vol);
 
+  // Keep DMA refills ahead of CPU-heavy app work until playback stops.
+  system_task_enable_raised_priority(true);
   audio_init((AudioDevice *)AUDIO);
   audio_set_volume((AudioDevice *)AUDIO, effective_vol);
   audio_start((AudioDevice *)AUDIO, prv_audio_trans_cb);
@@ -215,6 +217,7 @@ static void prv_stop_audio(void) {
   prv_update_volume_analytics(0);
 
   audio_stop((AudioDevice *)AUDIO);
+  system_task_enable_raised_priority(false);
 }
 
 static void prv_free_tracks(void) {
--- a/src/fw/services/speaker/speaker_service.c
+++ b/src/fw/services/speaker/speaker_service.c
@@ -95,7 +95,7 @@ typedef struct {
 
 static SpeakerServiceState s_state;
 
-// Serializes public APIs against prv_refill_bg (system task).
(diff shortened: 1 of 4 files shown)


ShinkoNet:
https://github.com/ShinkoNet/pebbleboy
Here's my app that is currently using these changes (along with a seperate filesystem blob API I've yet to make a PR for) as an example.

src/fw/services/speaker/speaker_service.c
```diff
@@ -205,6 +205,8 @@ static void prv_start_audio(uint8_t vol) {
   PBL_ANALYTICS_ADD(speaker_play_count, 1);
   prv_update_volume_analytics(effective_vol);
 
+  // Keep DMA refills ahead of CPU-heavy app work until playback stops.
+  system_task_enable_raised_priority(true);
```
jplexer, line 209: This boosts all `KernelBG` callbacks above the app for the entire playback lifetime, including long-lived streams. Can the priority boost be scoped to pending refill work instead, so unrelated background callbacks do not also preempt the app?

src/fw/services/speaker/speaker_service.c
```diff
@@ -300,8 +303,22 @@ static bool prv_can_preempt(SpeakerPriority new_pri) {
 //! This is the DMA refill callback path:
 //!   DMA ISR -> system_task_add_callback_from_isr -> audio driver trans_cb -> here
 static void prv_audio_trans_cb(uint32_t *free_size) {
-  // Schedule actual refill work on system task to keep ISR-context callback short
-  system_task_add_callback(prv_refill_bg, NULL);
+  // The drivers already dispatch on KernelBG; refill here and catch up missed blocks.
```
jplexer, line 306: This assumes `free_size` is the amount currently writable, but the QEMU driver always passes a hard-coded full-buffer size (`4096 * sizeof(int16_t)`) on every interrupt. Would need to be changed in the QEMU driver

jplexer (changes requested):

ShinkoNet:
Sorry for the wait. I've tried to address the priority boost. Tested on my Time 2 and audio still is fine under this scoped area when under load.

PR #2055, opened 2026-09-13
resources: Replaced crossover to modal svg images for obelix to fix gray replacement color appearing

consp:
The transition SVG's from modal to app (e.g. notification to watchface) have the quirk they become gray instead of filling the existing framebuffer. I thought it was a rendering bug, but it turns out it's the SVGs which do not work. The inward -> outward works due to the reverse order of transparent/color surfaces, but outward in causes everything to become gray due to the replacement color being present in all images and likely due to transparency not correctly being set. 
These images remove the original background image, add transparency to the inner field and provide an outside field set to the replacement color. I'm not sure about the Green/Islamic-Green color replacement scheme but since the code has been the same since the 4.3 firmware, this works.

Reddit thread describing the issue (video makes it quite clear): https://www.reddit.com/r/pebble/comments/1wdt9p8/heres_a_clearer_video_of_the_different_broken/

These images fix obelix. I have not checked any other but it's likely also present for other color screens.

Original:
[Kooha-2026-09-13-14-00-11.webm](https://github.com/user-attachments/assets/9f617837-0196-4d9e-9fa2-734e483fdf22)

Fixed:
[Kooha-2026-09-13-13-57-15.webm](https://github.com/user-attachments/assets/03118fbb-335a-4e9c-bc10-4133c4f6704e)

edit: Forgot to enable sign-off to commit, added DCO statement.
edit: Also live tested on Emery device.


jplexer (approved):
LGTM but please fix commit message (amend and force push c:)

consp:
Done, please also update the https://github.com/coredevices/PebbleOS/blob/main/CONTRIBUTING.md#dco-sign-off since it only names the _email address_ must match for clarity.  Also a bit sad Teller can't commit due to the two-word rule.

edits:
* changed commit message to include "resources: "
* added full name to author

PR #2059, opened 2026-09-14
i18n: move translations into a separate repository

jplexer:
Prep for having a translation platform (like weblate or parlance)


gmarull (commented):
Maybe I'm missing things, but what I'd do:

- Keep fonts (or any resources) here
- Remove the catalog and translations - this is what is always outdated and not easily manageable
- This repo uploads a new catalog on every merge if it changes.
- Packs should be directly managed/published by the other repo, and mobile app consume them

jplexer:
>     * Keep fonts (or any resources) here
That would be wrong, they are only consumed by the language pack, which is now built and stored in that other repo 
>     * Remove the catalog and translations - this is what is always outdated and not easily manageable
Will do, will also remove the submodule in that case
>     * This repo uploads a new catalog on every merge if it changes.
> 
>     * Packs should be directly managed/published by the other repo, and mobile app consume them
That was the plan!

gmarull (approved):

PR #2070, opened 2026-09-15
tools: add WebAssembly firmware text renderer

jplexer:
Used in Peblate (unreleased as of now)
image


tools/text2wasm/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+# SPDX-FileCopyrightText: 2026 Core Devices LLC
+# SPDX-License-Identifier: Apache-2.0
+cmake_minimum_required(VERSION 3.21)
+project(pebble-text2wasm LANGUAGES NONE)
+find_package(Python3 REQUIRED COMPONENTS Interpreter)
+find_program(EMCC emcc REQUIRED)
+get_filename_component(PEBBLEOS_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../.." ABSOLUTE)
+add_custom_target(text2wasm ALL
+  COMMAND ${Python3_EXECUTABLE} ${CMAKE_CURRENT_SOURCE_DIR}/build.py
```
gmarull: why is CMake invoking a Python script to compile?

jplexer: gpt was being stupid

gmarull: lol, we're not AGI yet, here the proof!

gmarull (approved):
cmake still not great, but not critical/core component, so let's go

PR #2073, opened 2026-09-15
fw/board: use stereo mic capture in all recovery builds

gmarull:
Commit 6ae8c1e1a ("fw/board: use stereo mic capture in MFG builds") restored stereo PDM capture for the mic test, but only gated it on `CONFIG_MFG`. The mic test apps live in the recovery firmware and are built for every PRF variant, so a non-MFG PRF still captured mono and hit the same 2x-speed playback and missing second microphone.

Gate stereo capture on `CONFIG_RECOVERY_FW` instead, which MFG builds also define.

Verified `obelix@pvt --variant=prf -DCONFIG_RELEASE=y` builds with `CONFIG_RECOVERY_FW=y` / `CONFIG_MFG` unset.


gmarull:
merging as trivial fix, needed for MF release

PR #2077, opened 2026-09-15
shell: configure Standard backlight intensity per board

jplexer:
## Summary

Getafix's Standard backlight preset uses 50% brightness. Add `CONFIG_BACKLIGHT_STANDARD_INTENSITY` (range 1–100, default 50) and set it to 25 in Getafix's shared board defconfig. The Standard preset reads this configuration directly, so both Getafix revisions use 25% for fresh settings and when Standard is selected.

Standard remains the default mode with its Standard dynamic curve. Battery Saver retains its Dim curve, and other boards retain their existing defaults. Saved preferences still take precedence; existing watches storing 50% may report Advanced until Standard is selected again.

Fixes FIRM-4392

The host test build defines the new setting as 50%, matching the Kconfig default, since it does not load firmware Kconfig.

## Validation

- Generated Kconfig for Getafix DVT and DVT2: both resolve Standard intensity to 25%.
- Generated Kconfig for Obelix PVT, Asterix, and QEMU Gabbro: all resolve to 50%.
- Compiled the updated `prefs.c` using the newly generated Getafix DVT2 configuration header and the existing build's compiler settings.
- Recompiled both host-test variants of `prefs.c` with the new setting; preferences tests passed using the rebuilt object. The local activity test link was blocked by incompatible cached objects from newer source.
- Fresh CI test run started for the correction.
- `git diff --check` and commit `gitlint` passed.
- No full firmware build or hardware test. The earlier build dependency step was blocked by missing `rsvg-convert`.

AI-assisted implementation using GPT-6 Astra.


gmarull (commented):
shouldn't we modify standard instead?

src/fw/shell/normal/prefs.c
```diff
@@ -142,7 +142,11 @@ static const BacklightPresetSettings s_backlight_preset_settings[] = {
 #ifdef CONFIG_DYNAMIC_BACKLIGHT
     .dynamic_mode = BacklightDynamicMode_Standard,
 #endif
+#ifdef CONFIG_BOARD_GETAFIX
+    .intensity = BACKLIGHT_INTENSITY_MEDIUM,
+#else
     .intensity = BACKLIGHT_INTENSITY_HIGH,
+#endif
```
gmarull: follow-up: make this into Kconfig, board sets it?

gmarull (approved):

src/fw/shell/normal/prefs.c
jplexer: lets do it

gmarull (approved):

PR #2085, opened 2026-09-16
treewide: run clang-format

gmarull:
Supersedes #2035 (thanks @nschonni for doing the groundwork).

Same intent as #2035, rebased on current `main` and with the config reduced to `BasedOnStyle: Google` plus the actual deviations instead of a full `--dump-config` dump. The dumped config differed from the Google preset in only 12 options, 3 of which are C++/Verilog-only, so the resulting formatting is identical for C (verified by running `--dry-run --Werror` with the #2035 config over the reformatted tree).

Deviations from Google, per the review in #2035:
- `ColumnLimit: 100`, `PointerAlignment: Right`, `SortIncludes: Never` (as before)
- all `AllowShort*OnASingleLine` off: functions, enums, ifs and loops always break
- `BreakAfterOpenBracketBracedList` / `BreakBeforeCloseBracketBracedList`: braced initializers always break after `{` (no `{.name = ...` on one line)
- `AllowAllArgumentsOnNextLine: false`
- `AlignConsecutiveMacros: Consecutive`: consecutive `#define`s stay column-aligned (register maps, colour tables)
- `BreakStringLiterals: false`: long strings are left as the author wrote them, so log / i18n messages stay greppable
- `BracedInitializerIndentWidth: 2` and `SpacesBeforeTrailingComments: 1`: match the prevailing style of the existing sources (measured over the tree), which cuts the reformat diff by ~16%

Other changes vs #2035:
- protos follow the same single config (100 cols) instead of a separate Google-defaults `Language: Proto` section
- generated bitmap data headers (`demo_images/*`, `fps_test_bitmaps.h`, `test_bitmaps.h`) are excluded via `.clang-format-ignore`, since the trailing `/* bytes N - M */` comments made clang-format split every 16-byte row in two
- `tests/` is formatted too; only `tests/fixtures/**` is excluded. That is what broke `test_kraepelin_algorithm` in #2035: it parses the step-sample fixtures as text and needs the `{ x, y, z}` rows to keep their leading space
- the CI job rebases onto the target branch like the other compliance jobs

Verification:
- `clang-format --dry-run --Werror` (v22.1.8, same major as CI) is clean over all tracked `.c/.h/.proto` files
- comment- and whitespace-stripped token streams are identical before/after for every file


.github/workflows/compliance.yml
```diff
@@ -101,3 +101,17 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v7
+        with:
+          fetch-depth: 0
+
+      - name: Rebase onto target branch
+        uses: ./.github/actions/rebase
+
+      - name: Run clang-format
+        uses: jidicula/clang-format-action@v4.18.0
```
nschonni, line 117: Done, pinned to `clang-format-version: "22"`.

.clang-format-ignore
```diff
@@ -0,0 +1,9 @@
+**/*.template.c
+**/*.template.h
+apps/samples/fps_test/src/test_bitmaps.h
+sdk/defaults/lib/lib.c
+sdk/defaults/lib/lib.h
+src/fw/apps/demo/demo_images/*
+src/fw/apps/demo/fps_test_bitmaps.h
+tools/clar/*
+tests/**
```
nschonni: Traced it: `test_kraepelin_algorithm` parses the `tests/fixtures/activity/step_samples/*.c` files as *text* at runtime and only recognises a sample row when the first token is exactly `{` (`{ -120, -376, -912},`). Google's braced-list style drops that space, so the token became `{-120,`, zero samples were parsed and `num_tests > 0` failed.

The fixtures are captured data whose layout is effectively the test's input format (and ~650k lines of pointless churn), so I now exclude only `tests/fixtures/**` and format the rest of `tests/`.

jplexer (approved):

.clang-format-ignore
nschonni: lol, changing my original message to a response doesn't send a notification, but glad you figured it out

gmarull: sorry, we're not AGI yet then lol

PR #1596, opened 2026-06-23
Day/night backlight color

nick1udwig:
## Problem

Would be nice if watches with RGB backlight could automatically switch backlight color, with one color for day and one for night. 

Resolves #1139.

## Solution

For watches that have RGB backlight, add `default_color_night` state along with time tracking for sunset/sunrise.

## Notes

1. Sister PR manages setting the nighttime color: https://github.com/coredevices/mobileapp/pull/265
2. I considered allowing user to set sunrise/set. Instead I decided to have it default to 6am/6pm and have the mobileapp supply more accurate times (based on location) if available. The location logic lives in https://github.com/coredevices/mobileapp/pull/265
3. ~~Tested with QEMU, but not on a live watch (see 9e9a67e)~~ EDIT: also tested on a live watch over the course of multiple days, it worked great

PR #2092, opened 2026-09-17
Day/night backlight color

### LINKED ISSUE #1139: Time 2 - "night mode" using RGB backlight?

### Proposed feature request

Given that the PT2 has an RGB backlight, it would be neat to have a "night mode" backlight toggle that used warmer hues in the evening (either through setting a time window or syncing to sunrise/sunset). Combined with the dynamic backlight setting, it feels like this would make for a very pleasant night viewing experience!

### Alternatives

_No response_

### Anything else?

_No response_

gmarull:
Revives #1596 by @nick1udwig (closed by the stale bot) on top of current `main`. Original authorship is preserved on all commits.

## Problem

Watches with an RGB backlight should be able to switch automatically between a day and a night backlight color.

Resolves #1139.

## Solution

Four new syncable prefs: `lightColorDayNightEnabled`, `lightColorNight`, `lightColorSunriseMinute`, `lightColorSunsetMinute` (defaults 06:00/18:00; the phone can supply location-based times). The light service picks the system color from the schedule wherever it would otherwise use the user's default color, and arms a timer for the next sunrise/sunset boundary so a backlight that is already on re-tints at the transition.

A `backlight daynight <start|status|off>` console command arms a green→red switch one minute out for quick manual testing.

## Changes vs. #1596

- Rebased on the CMake tree; `light_commands.c` folded into `services/light/console.c`, no longer QEMU-only.
- Scheduling follows the Do Not Disturb pattern: `time_util_get_seconds_until_daily_time()`, `light_handle_clock_change()` from the event loop on `PEBBLE_SET_TIME_EVENT`, and `light_handle_color_prefs_changed()` from the pref handlers (covers both local sets and phone sync), instead of subscribing to events inside the service.
- **No capability bit.** Older firmware already rejects unknown pref keys (`settings_blob_db.c`, "Rejecting non-whitelisted setting"), and every other recently added pref, including `lightColor` itself, ships without one. coredevices/mobileapp#265 should drop `SupportsBacklightColorSchedule` and gate the UI on the watch model (RGB backlight hardware), or not gate at all like `lightColor`.
- The original tests were guarded by `CONFIG_BACKLIGHT_HAS_COLOR`, which the test build never defined, so they never ran. `test_light` now builds with it and the fake RTC, covering color selection around sunrise/sunset, midnight-wrapping schedules, the equal-times case, timer arming, and re-tinting a lit backlight when the timer fires.

## Testing

- `pbl test -R test_light` passes (19 cases, 6 new).
- Builds: qemu_emery, obelix@pvt (normal + PRF).
- QEMU (qemu_emery): `backlight daynight start` → green backlight; at the sunset minute the timer switched it to red while lit; `off` restores.


jplexer (approved):

gmarull:
@sjp4 please check the mobile side, otherwise this feature can't be used in practice

PR #2093, opened 2026-09-17
fw/comm/ble/ancs: warn the user when iOS rejects the Control Point

gmarull:
## Summary

Follow-up to the FIRM-4398 / FIRM-4277 analysis: on affected iPhones every ANCS alive check fails with ATT `0x03 Write Not Permitted` on a freshly discovered Control Point, the CCCD writes succeed, and no notification ever arrives. A healthy phone on the same iOS version answers the same write on the same handle with the expected `0xA2`. iOS is refusing this watch (most likely "Share System Notifications" is off, denied, or pending a prompt), and nothing on the watch side can recover it, so the existing recovery loop spins silently forever.

The ANCS client counts consecutive rejected alive checks and, after two (~30 min), hands off to the ANCS notifications service, which pushes a notification telling the user how to re-enable sharing on the iPhone. The warning is shown once per block episode: the flag persists across reconnects and is cleared as soon as iOS answers an alive check normally. The GATT client only detects the condition; building the user-facing notification stays in `services/notifications/ancs`.

## Testing

- `pbl test -R ancs`: new `alive_check_rejected_warns_user_once` covers the threshold, the once-per-episode behaviour across a reconnect, and re-arming after iOS recovers. All six ANCS test binaries pass.
- Builds clean for `obelix@pvt`.
- Not tested on hardware against a phone in the blocked state (needs a user with the condition, see FIRM-4398).


jplexer (approved):

gmarull:
verified on real phone+watch

PR #2101, opened 2026-09-17
docs/architecture: propose a settings dictionary

gmarull:
Design proposal, opened for discussion rather than merge-as-is. Rendered page: `docs/architecture/settings_dictionary.md`.

**Problem.** Adding a syncable setting today touches five places (prefs.c, prefs_values.h.inc, the settings_blob_db whitelist, libpebble3 `WatchPrefEntity.kt`, the CoreApp/iOS screens), the wire encoding is convention only, and the phone cannot learn what a given watch supports.

**Proposal.** A new, independent feature (legacy prefs and BlobDB 0x0F untouched):

- **Dictionary** (`settings.yaml`): firmware-owned data contract. Stable `u16` ids, types with explicit wire encoding, hard bounds, defaults, access, Kconfig gating. No presentation.
- **UI schemas** (`ui/mobile.yaml`, `ui/watch.yaml`): grouping, labels, widget choice, display units, visibility. Reference entries by name, validated against the dictionary at build time. Same `u8` can be four presets on the phone and a raw number on a debug page.
- **Generator**: C table + typed accessors + generic validation for the firmware; CBOR blobs of dictionary and UI schemas embedded as resources and served to the phone (hash-cached, fetched once per firmware update).
- **Protocol**: watch is authoritative. `DICT_INFO`, `READ_ALL`, `READ`, `WRITE`, `CHANGED`. No timestamps, dirty lists or conflict resolution.
- **Mobile**: one CBOR decoder and one generic renderer in libpebble3, no per-setting code. `custom:<name>` widgets as an escape hatch.

Rollout in three steps (subsystem only, then transport, then optional migration of legacy prefs). Open questions are listed at the end of the page.

@jplexer @sjp4 would appreciate your thoughts, in particular on the two-schema split and the watch-authoritative protocol replacing the BlobDB sync model.


docs/architecture/settings_dictionary.md
```diff
@@ -0,0 +1,325 @@
...
+   owned by app or design people. Reference dictionary entries by name and
+   never restate types.
+3. **Generator**: one build step that validates both against each other and
+   emits a C table plus typed accessors for the firmware, and CBOR blobs of the
+   dictionary and each UI schema that the watch serves to the phone.
+
+The watch is authoritative for values. The phone reads all values on connect,
+subscribes to change notifications, and sends writes as commands the watch
+validates and either applies or rejects. There are no timestamps, no dirty
+lists and no conflict resolution.
```
sjp4, line 56: I think that's a pretty big downgrade:
- Syncing every value on every reconnect is a lot of traffic, especially if we're e.g. in a disconnect/reconnect loop
- Conflict resolution is important: the phone is managing a set of preferences which apply to every watch (i.e. if they change a preference it applies to every watch that they connect). This relies on conflict resolution via the timestamp to work.

docs/architecture/settings_dictionary.md
```diff
@@ -0,0 +1,325 @@
...
+The dictionary and UI blobs are fetched through the existing get_bytes
+service with a new object type, which already handles chunking; the hashes
+from `DICT_INFO` let the phone cache them and fetch once per firmware update.
+The same blobs are included in the firmware bundle so the app can prefetch
+them before an OTA and render the new screen immediately after reboot.
+
+On connect the phone checks the hashes, fetches blobs if needed, issues
+`READ_ALL` and then applies `CHANGED` messages to a display cache. An edit
+made while disconnected is a queued `WRITE` that is applied or rejected on
+reconnect. The phone never holds authoritative state.
```
sjp4, line 281: The phone has to hold some authoritative state if it's trying to keep the same preferences consistent across multiple watches (unless I'm misunderstanding what that means)

docs/architecture/settings_dictionary.md
```diff
@@ -0,0 +1,325 @@
...
+## Open questions
+
+- Blob transport: get_bytes object type (reuses chunking) versus chunked reads
+  on the new endpoint (self-contained). The draft assumes get_bytes.
+- Label ownership: English in the YAML with translation via gettext on the
+  watch and app-side tables on the phone, or a separate strings file per
+  locale in the repo.
+- Whether phone writes to entries shown only on debug pages should require a
+  developer-mode flag on the watch.
+- Struct fields versus flattening every struct into scalar entries. Scalars
```
sjp4, line 323: I'm assuming cbor structs are forwards/backwards compatible so that isn't an issue? (structs break the existing system because of versioning)

docs/architecture/settings_dictionary.md
```diff
@@ -0,0 +1,325 @@
...
+## On-watch settings
+
+Because the watch UI schema carries labels, widgets and ranges for the same
+table, the on-watch Settings app can render generic entries from it. Adding a
+plain toggle then becomes a one-line change in `settings.yaml` plus one item
+in each UI schema, with no C and no mobile work.
+
+## Rollout
+
+1. **Subsystem only.** `subsys/settings`, the YAML files, generator, typed
```
sjp4, line 304: Another option: roll out mechanism for syncing catalogue/UI to phone, but still using existing blobdb sync for preference values (from my point of view, would solve the biggest problems we have today - namely keeping everything in sync)

PR #2102, opened 2026-09-18
fw/sdk/notifications: notification unread count tracking and watchface api

bentemple:
# Description

These 3 commits add a notification unread count API to allow watch-faces to show unread notifications, as well as hooks to mark a notification as read when the user actions the notification. Most of the work for this was actually already done, so this is just connecting the remaining dots and exposing it to watchfaces as a system api.

I took a (somewhat) aggressive stance on what constitutes reading a notification. Interacting with a notification in any way will consider a notification as read:
- closing it with the back button when it first pops-up on-screen
- scrolling through each notification
- dismissing the notification on your phone

I have found this approach to result in no false-positives keeping the unread notification genuinely helpful and not something that is shown for a notification I have already seen.

# Test-Plan
Built custom version of PebbleOS
Exported SDK 
Built a custom watch-face depending on the new APIs this exposes
Installed both PebbleOS and the new watchface to my watch
Validated for >7 days that the notification dot was shown when expected, and cleared when it should be based on my own actions (dismissing, interacting, opening notification app, etc...)

I ran all of the new unit tests and verified functionality

I also clanked a custom integration test suite to validate various test scenarios from within an emulator, driven by python scripts.


gmarull (changes requested):
We won't add new APIs that allow accessing user information without adding a permission system first (there's a draft PR with initial ideas).

PR #2104, opened 2026-09-18
fw/services/timeline: add option to skip upcoming repeating alarms on watch

bentemple:
# Description

This adds the option to "skip" an upcoming repeating alarm instance.

This feature is introduced with the following 3 commits:
8f611a2 - Ensures that the from_watch flag is correctly saved on pins sent from the phone. This is a (minor) security fix such that phone supplied watch timeline pins, can never impersonate a watch created pin or reach the alarm_skip_occurance() 
bfbdb29 - Update alarms to be skippable, which is the key feature underpinning the UI work
4cae232 - Add the skip action to repeating alarm pins in the timeline

# Test-Plan
Built and ran new unit tests
Built a custom emulator python test-suite for validating and building the feature
Compiled and installed a custom version of PebbleOS
Validated that alarms still fire as expected
Validated that skipping alarms only skips one instance, as expected

I have been using this code on my watch for 1-2 weeks and it is working as intended


bentemple:
I guess kind of a different approach to how it's done in #1901, though I frankly hadn't seen that one before I opened this. Happy to just close as a duplicate.

I at least like how they're saving the data more explicitly.

